PR #110, opened 2023-01-13
term: do not enable mouse reporting

### LINKED ISSUE #109: term: all mouse events captured in xterm
The escape sequence we use is supposed to ask for only mouse click reporting. Curiously, this hasn't worked in any terminal we tried.

Worse, turns out xterm takes note in a confused way and stops feeding mouse events to the terminal emulator even, preventing one from scrolling the buffer. We should just not issue a mouse reporting escape sequence at all for the time being.

Fang-:
The style of mouse reporting we want, "mouse clicks only", appears to not be supported in any common terminal emulator. Worse, in xterm, that mode ends up preventing the user from scrolling the buffer.

Until such a time where we can handle _all_ mouse inputs gracefully, we will refrain from taking mouse clicks in real terminals.

Closes #109. h/t @ashelkovnykov for spotting the xterm problem.

### DIFF
--- a/pkg/vere/io/term.c
+++ b/pkg/vere/io/term.c
@@ -117,9 +117,6 @@ u3_term_log_init(void)
     //    and simply use constant sequences.
     //
     {
-      uty_u->ufo_u.mon_u = TERM_LIT_BUF("\033[?9h");
-      uty_u->ufo_u.mof_u = TERM_LIT_BUF("\033[?9l");
-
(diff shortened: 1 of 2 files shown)


mcevoypeter (approved):

zalberico:
Once merged, we should deploy this to edge, and then move it over to release to deploy to soon

PR #59, opened 2023-01-09
Automatically create GitHub releases

matthew-levan:
Resolves #57.

PR #136, opened 2023-01-19
Automatically create GH releases

### LINKED ISSUE #57: Automatically create GitHub releases
Use the `gh` CLI to create a new GitHub release as part of the existing release flow in our CI code. GitHub releases should include:
* a compressed archive for each of our supported platforms, 
* release notes, 
* and a list of contributions/commits since last release.

matthew-levan:
Resolves #57. Had to create a new PR, so this deprecates #59 (however, see #59 for further context and history).


zalberico:
Thanks @matthew-levan this is something we need! and thanks for following the feature branch naming scheme ❤️

@philipcmonk @mcevoypeter - based on conversation with philip doing this manually was a confusing extra step, we should get this to be part of the build automation.

mcevoypeter:
Automating GitHub releases is definitely something we want to do, but I'm leery of doing so without thorough testing on a separate repo. In releasing 1.16 and 1.17, I've noticed that creating a GitHub release unexpectedly triggers CI jobs--tracked in #150--which it shouldn't do, so automating the release creation runs the risk of causing even more problems.

matthew-levan:
> Automating GitHub releases is definitely something we want to do, but I'm leery of doing so without thorough testing on a separate repo. In releasing 1.16 and 1.17, I've noticed that creating a GitHub release unexpectedly triggers CI jobs--tracked in #150--which it shouldn't do, so automating the release creation runs the risk of causing even more problems.

Fair. However, [according to the official docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release):
> Workflows are not triggered for the `created`, `edited`, or `deleted` activity types for draft releases.

mcevoypeter:
That's more reassuring, but I still don't think it makes sense to merge this until we resolve #150.

mcevoypeter:
@matthew-levan I'd like to see support for full automation via a release description file mentioned in https://github.com/urbit/vere/issues/57#issuecomment-1398918033 before considering merging this. Let me know if anything in the linked comment is unclear, and I'd be happy to clarify.

matthew-levan:
> @matthew-levan I'd like to see support for full automation via a release description file mentioned in [#57 (comment)](https://github.com/urbit/vere/issues/57#issuecomment-1398918033) before considering merging this. Let me know if anything in the linked comment is unclear, and I'd be happy to clarify.

Thanks for the suggestion @mcevoypeter. During today's release meeting in the gather town, we agreed that it seems best to automate everything in the process of GH release creation *except* the addition of the human-written release notes. It's preferable to add the "author's flavor" to the release notes as a team during the weekly release meeting through the GitHub website rather than having to be encumbered by another file (`.RELEASE_DESCRIPTION.md`) to be committed/pushed and parsed by the workflow. Additionally, I think the auto-generated notes from the initial creation of the release draft would be overwritten by a `gh release edit --notes-file .RELEASE_DESCRIPTION.md` command. Of course, I *could* pull the auto-generated notes into a variable or temporary file and then concatenate the two together in the workflow code, but I don't see the value because of the above point regarding preference for "live" editing of the human-written release notes during the release call.

matthew-levan:
I'm going to test this on my own fork and publish results here.

matthew-levan:
Note: I was going to test this workflow code on my own forked repository, but simply scaffolding the test would require quite a bit of work (in addition to ripping out a number of steps in the workflow, like `gcloud` authentication for example).

With a careful enough review, I'm confident to merge this. At worst, the additional step just fails, which won't effect the rest of the pipeline.

PR #139, opened 2023-01-19
term: make printfs render with 50% grayscale

### LINKED ISSUE #138: Legibility of printfs is poor in common dark color schemes
A number of people have reported that, in their dark mode terminal color schemes, the printfs (rendered as the "bright black" color code) have poor contrast relative to the background they're rendered on. In some cases, they're the exact same color.

Potential solutions here include:
- Making the printfs 50% gray.
- Making the printfs some other color.
- Giving the printfs a background color explicitly.

The first of those options seems most appealing. It keeps in line with the goal of the printf color change at making them jump out _less_ than formal outputs, and should be reasonably legible for both light and dark color themes.

Fang-:
As opposed to the color code for "bright black", which may cause legibility issues in dark color schemes.

Fixes #138.


mcevoypeter (approved):
LGTM. Makes a small but noticeable improvement in the legibility in my dark-colored terminal.

Issue #40, opened 2022-12-16
Fakeship tests fail on macOS builds


matthew-levan:
https://github.com/urbit/vere/actions/runs/3715632920/jobs/6301027661

Same errors on `macos-x86_64` and `macos-arm64`.

joemfb:
Obsolete

PR #141, opened 2023-01-19
Run fake ship tests with optimized binary under memory and CPU debug

### LINKED ISSUE #140: Run fake ship tests with optimized binary under memory and CPU debug
The fake ship tests--in `//pkg/vere:test-fake-ship`--do not use an optimized Urbit binary because the Urbit binary is listed as a dependency of `//pkg/vere:test-fake-ship` via the [tools attribute](https://bazel.build/reference/be/general#genrule.tools), which is built using the exec configuration instead of the target configuration. C compiler options can be set in the exec configuration using [`--host_copt`](https://bazel.build/docs/user-manual#host-copt).

The fake ship tests also fail to turn on CPU and memory debugging via the `U3_CPU_DEBUG` and `U3_MEMORY_DEBUG` macros. This can also be rectified using `--host_copt`.

mcevoypeter:
## Description
Resolves #140. Note that debug symbols are still not included in the exec configuration build despite `--host_copt='-g'` being set in [.bazelrc](https://github.com/urbit/vere/tree/0d7f0c636fa247ebc63dba63fff4c8c8bd5405cb/.bazelrc). It's unclear why this remains the case and may warrant further investigation if we deem it necessary.

## Testing
Ran `bazel build --subcommands //pkg/vere:test-fake-ship` on both `linux-x86_64` and `macos-aarch64` machines and confirmed compiler invocations on both platforms included `-g -O3 -DU3_MEMORY_DEBUG -DU3_CPU_DEBUG`. The `linux-x86_64` build ran successfully, and the `macos-aarch64` build failed due to an existing issue (#40).

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -26,5 +26,10 @@ build --strip=never
 build --copt='-O3'
 build --host_copt='-O3'
 
+# Turn on CPU and memory debug for exec config, which we only use to run the
+# fake ship tests.
+build --host_copt='-DU3_CPU_DEBUG'
+build --host_copt='-DU3_MEMORY_DEBUG'
+
 # Any personal configuration should go in .user.bazelrc.
 try-import %workspace%/.user.bazelrc
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,10 +19,12 @@ build --flag_alias=gcc_version=//:gcc_version
 
 # Always include source-level debug info.
 build --copt='-g'
+build --host_copt='-g'
 build --strip=never
 
 # Use -O3 as the default optimization level.
 build --copt='-O3'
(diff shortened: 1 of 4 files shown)


barter-simsum (approved):

mcevoypeter:
If you're reading this and are wondering why this PR includes changes to C source code, here's some context: prior to #2, we were defining--not just declaring--global variables in header files, which causes multiple definitions in a lot of cases. As part of a move toward better header file hygiene, #2 addressed that, but it missed the global variables gated by `U3_MEMORY_DEBUG`/`U3_CPU_DEBUG`. This PR uncovered that error and corrected it.

PR #144, opened 2023-01-19
Remove lingering artifacts from old build system

### LINKED ISSUE #125: Remove lingering artifacts from old build system

mcevoypeter:
## Description

Resolves #125.


barter-simsum:
LGTM. 

flw.txt still included in https://github.com/urbit/urbit/blob/develop/doc/spec/flw.txt

Mostly consists of removal of files necessary for building on Windows and openbsd, neither of which we support.

A few misc. scripts that bazel obviously no longer depends on, etc.

barter-simsum (approved):

PR #146, opened 2023-01-20
INSTALL.md: clarify macOS prerequisites

belisarius222:
I didn't realize I could just go ahead and run `bazel build :urbit` without first adapting one of the `bazel run //toolchain...` prerequisite commands first.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,9 +24,7 @@ $ brew install automake libtool
 
 ### Linux
 
-We use [musl libc][musl libc] instead of [glibc][glibc] on Linux, which enables us
-to generate statically linked binaries. As a prerequisite, you need to install
-the [musl libc][musl libc] toolchain appropriate for your target platform.
+After installing `automake` and `libtool`, you need to install the [musl libc] toolchain.  We use [musl libc][musl libc] instead of [glibc][glibc] on Linux, which enables us to generate statically linked binaries. As a prerequisite, you need to install the [musl libc][musl libc] toolchain appropriate for your target platform.
 
 #### x86_64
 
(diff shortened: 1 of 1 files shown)


matthew-levan:
Nit: '%s/MacOS/macOS/g'

INSTALL.md
```diff
@@ -15,16 +15,23 @@ run:
...
 ```console
 $ brew install automake libtool
 ```
 
+Now MacOS should be ready to use `bazel` to build Vere.
+
 ### Linux
 
-We use [musl libc][musl libc] instead of [glibc][glibc] on Linux, which enables us
+Make sure `automake` and `libtool` are installed on your system.  TODO: describe how to install these on common distros
```
mcevoypeter, line 32: I think moving instructions to install `automake` and `libtool` under OS-specific headers adds volume without adding value. Instead, I'd suggest the following:

```
## Prerequisites
<no change>

### Linux

After installing `automake` and `libtool`, you need to install the [musl libc] toolchain. We use [musl libc][musl libc] instead of [glibc][glibc] on Linux, which enables us to generate statically linked binaries.

#### x86_64

<snip>

### macOS
After installing `automake` and `libtool`, you're ready to build Vere.
```

mcevoypeter (changes requested):

INSTALL.md
mcevoypeter: Also, as @matthew-levan pointed out, `MacOS` should be `macOS`.

belisarius222:
@mcevoypeter PTAL

mcevoypeter (approved):

Issue #131, opened 2023-01-18
Problem building on macOs Monterey


pkova:
OS version: 12.6.2 (21G320)

Tried the steps in the guide at INSTALL.md, also tried the solutions in #119.

```
Pyrys-MacBook-Pro:vere pkova$ bazel clean --expunge
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
```
```
Pyrys-MacBook-Pro:vere pkova$ bazel --version
bazel 6.0.0
```
```
Pyrys-MacBook-Pro:vere pkova$ clang --version
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
Pyrys-MacBook-Pro:vere pkova$ brew info automake
==> automake: stable 1.16.5 (bottled)
Tool for generating GNU Standards-compliant Makefiles
https://www.gnu.org/software/automake/
/usr/local/Cellar/automake/1.16.5 (131 files, 3.5MB) *
  Poured from bottle on 2023-01-15 at 15:35:23
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/automake.rb
License: GPL-2.0-or-later
==> Dependencies
Required: autoconf ✔
==> Analytics
install: 58,627 (30 days), 178,144 (90 days), 736,041 (365 days)
install-on-request: 39,115 (30 days), 122,671 (90 days), 488,706 (365 days)
build-error: 8 (30 days)
```
```
Pyrys-MacBook-Pro:vere pkova$ brew info libtool
==> libtool: stable 2.4.7 (bottled)
Generic library support script
https://www.gnu.org/software/libtool/
/usr/local/Cellar/libtool/2.4.7 (75 files, 3.8MB) *
  Poured from bottle on 2022-09-27 at 17:21:45
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/libtool.rb
License: GPL-2.0-or-later
==> Dependencies
Required: m4 ✔
==> Caveats
All commands have been installed with the prefix "g".
If you need to use these commands with their normal names, you
can add a "gnubin" directory to your PATH from your bashrc like:
  PATH="/usr/local/opt/libtool/libexec/gnubin:$PATH"
==> Analytics
install: 196,137 (30 days), 646,533 (90 days), 3,404,165 (365 days)
[...]

PR #147, opened 2023-01-20
build: fix broken import for macos-x86_64

pkova:
These were the most minimal set of changes that allowed me to build vere on macOS x86-64. See #131 for context. To build, I ran `bazel build --clang_version="12.0.0" :urbit`.

### DIFF
--- a/pkg/ent/ent.c
+++ b/pkg/ent/ent.c
@@ -12,6 +12,8 @@ int ent_getentropy(void* buf, size_t len) {
 
 #elif defined(ENT_GETENTROPY_SYSRANDOM)
 #include <stddef.h>
+// See https://www.mail-archive.com/bug-gnulib@gnu.org/msg38583.html.
+#include <stdlib.h>
 #include <sys/random.h>
 int ent_getentropy(void* buf, size_t len) {
   return getentropy(buf, len);


bazel/toolchain/BUILD.bazel
```diff
@@ -235,6 +235,7 @@ cc_toolchain_config(
         "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/",
         "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include",
         "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/{compiler_version}/include",
+        "/usr/local/include",
```
pkova, line 238: This is where my box has `blake2.h`, required by `pkg/urcrypt/argon.c`.

zalberico:
It'd be a good idea to update your macOS version - Monterey is a major version behind and I don't think we should put effort into making sure things work on old macOS versions. I'm not sure if that's the case here, but that would have been the first thing I would have done, CC: @mcevoypeter

pkg/ent/ent.c
```diff
@@ -12,6 +12,8 @@ int ent_getentropy(void* buf, size_t len) {
 
 #elif defined(ENT_GETENTROPY_SYSRANDOM)
 #include <stddef.h>
+// See https://www.mail-archive.com/bug-gnulib@gnu.org/msg38583.html.
+#include <stdlib.h>
```
mcevoypeter, line 16: This is a good find.

mcevoypeter (commented):
Introducing a dependency on `/usr/local/include` hurts the reproducibility of the build, and I'm hesitant to incorporate such a change. The issue you ran into is part of a larger issue with the `macos-x86_64` toolchains, which I'd prefer to tackle head-on (see #151) instead of indirectly via an installation-dependent fix like this. The `#include <stdlib.h>` fix should definitely be added, as should disabling `-g` on `macos`, but I don't want to merge the addition of `/usr/local/include` to the `macos-x86_64` toolchain's `sys_includes`. Could you update the PR description to reflect this and remove the `/usr/local/include`-related change?

pkova:
Done!

pkg/vere/BUILD.bazel
```diff
@@ -140,6 +140,10 @@ cc_library(
 
 cc_binary(
     name = "urbit",
+    copts = select({
+        "@platforms//os:macos": [],
+        "//conditions:default": ["-g"],
+    }),
```
mcevoypeter, line 146: After thinking further about this yesterday, I don't think this is the best way to disable inclusion of debug info on macOS. `-g` won't be set on macOS only when building via this target (`//pkg/vere:urbit`), whereas what we really want is to globally turn off `-g` so that all targets on macOS are built off of `-g` whether or not they were triggered by this target or not. To that end, the fix in [`b16ada6baaccdce2a3dda78caf70b55d8dbbc28e`](https://github.com/urbit/vere/tree/b16ada6baaccdce2a3dda78caf70b55d8dbbc28e) is a better one.

pkova: Makes sense to me. Fixed.

jalehman:
@mcevoypeter mind giving this another review and signoff if it looks good now? I believe all previous comments were addressed.

PR #154, opened 2023-01-23
Remove Windows-specific code

### LINKED ISSUE #153: Remove Windows-specific code
As of v1.16, we no longer support Windows. As a result, we should remove Windows-specific code from the runtime, which is gated by the `U3_OS_mingw` macro. In the future, if we add Windows support back, the code removed here can serve as a reference point. Additionally, re-adding Windows support in the future will force us to rewrite Windows-specific logic with the aid of the code removed by this issue, which will presumably lead to a cleaner Windows implementation than attempting to maintain the now-unused Windows code until then and "turn it back on."

mcevoypeter:
## Description

Resolves #153. For the rationale behind this PR, see #153.

## Testing

Ran the following on `linux-x86_64` and `macos-aarch64`:
```console
$ bazel build :urbit
$ ./urbit -F zod -B path/to/solid.pill
$ zod/.run
```
Note `path/to/solid.pill` pointed to a local checkout of commit `ef2a0c913b66a04adda7bf1213ca61ef5c0a68e6` in https://github.com/urbit/urbit.

### DIFF
--- a/pkg/vere/lord.c
+++ b/pkg/vere/lord.c
@@ -1158,7 +1158,7 @@ u3_lord_init(c3_c* pax_c, c3_w wag_w, c3_d key_d[4], u3_lord_cb cb_u)
   //  spawn new process and connect to it
   //
   {
-    c3_c* arg_c[10];
+    c3_c* arg_c[9];
     c3_c  key_c[256];
     c3_c  wag_c[11];
     c3_c  hap_c[11];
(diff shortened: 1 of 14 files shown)


pkg/vere/lord.c
```diff
@@ -1158,7 +1158,7 @@ u3_lord_init(c3_c* pax_c, c3_w wag_w, c3_d key_d[4], u3_lord_cb cb_u)
   //  spawn new process and connect to it
   //
   {
-    c3_c* arg_c[10];
+    c3_c* arg_c[9];
```
zalberico, line 1161: Everything else looks self explanatory except the change here (just because I don't know what it's for)

zalberico (approved):

pkg/vere/lord.c
mcevoypeter: The Windows build used 9 arguments and therefore needed a 10-element array since the array needs to be NULL-terminated. Linux and macOS builds only need 8 arguments and therefore only need a 9-element array.

PR #155, opened 2023-01-24
Eval improvements

mopfel-winrux:
I added the ability for the user pick two flags to pass to the eval subroutine. 

The first is the `--jam` flag. This takes the result from the inputted hoon code , `+jam`'s it and outputs it as a hexadecimal string. 

The second is the `--jamkhan` flag. This takes the result from the inputted hoon code, `+jam`'s it and outputs the noun wrapped in the proper `%khan` header (leading zero, noun size, noun) as binary data.

The proper use of the eval is as follows:
`echo "(add 2 2)" | vere eval --jamkhan > test.jam`

You can directly inject a `|meld` command to a running pier as follows:
`echo "[1 %urth %meld]" | ./vere  eval --jamkhan | nc -U ~/urbit/zod/.urb/conn.sock`

This should be the easiest way to `+jam` a noun for `%khan` use

This is a clone commit of [#6048](https://github.com/urbit/urbit/pull/6048) on urbit/urbit

### DIFF
--- a/pkg/noun/vortex.c
+++ b/pkg/noun/vortex.c
@@ -120,6 +120,31 @@ _cv_nock_wish(u3_noun txt)
   return pro;
 }
 
+/* u3v_wish_n(): text expression with cache.
+*/
+u3_noun
+u3v_wish_n(u3_noun txt)
+{
+  u3t_event_trace("u3v_wish", 'b');
+  u3_weak exp = u3kdb_get(u3k(u3A->yot), u3k(txt));
+
+  if ( u3_none == exp ) {
+    exp = _cv_nock_wish(u3k(txt));
+
+    //  It's probably not a good idea to use u3v_wish()
+    //  outside the top level... (as the result is uncached)
+    //
+    if ( u3R == &u3H->rod_u ) {
+      u3A->yot = u3kdb_put(u3A->yot, u3k(txt), u3k(exp));
+    }
+  }
+
+  u3t_event_trace("u3v_wish", 'e');
+
+  u3z(txt);
+  return exp;
+}
+
 /* u3v_wish(): text expression with cache.
 */
 u3_noun
--- a/pkg/noun/vortex.c
+++ b/pkg/noun/vortex.c
@@ -120,7 +120,7 @@ _cv_nock_wish(u3_noun txt)
   return pro;
 }
 
-/* u3v_wish_n(): text expression with cache.
+/* u3v_wish_n(): text expression with cache. with the input as a u3_noun.
 */
 u3_noun
 u3v_wish_n(u3_noun txt)
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -1246,39 +1246,42 @@ _cw_eval(c3_i argc, c3_c* argv[])
     u3z(gat);
   } else {
     u3_noun sam = u3i_string(evl_c);
-    u3_noun res = u3v_wish_n(sam);
+    u3_noun res = u3m_soft(0, u3v_wish_n, sam);
     c3_d bits = 0;
     c3_d len_d = 0;
     c3_y* byt_y;
-    bits = u3s_jam_xeno(res, &len_d, &byt_y);
-    if ( c3n == kan_o ) {
-      fprintf(stderr,"jammed noun: ");
-      for ( size_t p=0; p < len_d; p++ ){
-        fprintf(stdout,"\\x%2x", byt_y[p++]);
+    if ( 0 == u3h(res) ) {  //  successful execution, print output
+      bits = u3s_jam_xeno(u3t(res), &len_d, &byt_y);
+      if ( c3n == kan_o ) {
+        fprintf(stderr,"jammed noun: ");
+        for ( size_t p=0; p < len_d; p++ ){
+          fprintf(stdout,"\\x%2x", byt_y[p++]);
+        }
+        fprintf(stderr,"\n");
+      } else {
+         fprintf(stderr,"khan jammed noun: ");
+         c3_y out_y[5];
+         out_y[0] = 0x0;
+         out_y[1] = ( len_d        & 0xff);
+         out_y[2] = ((len_d >>  8) & 0xff);
+         out_y[3] = ((len_d >> 16) & 0xff);
+         out_y[4] = ((len_d >> 24) & 0xff);
+         fwrite(out_y, 1, 5, stdout);
+         if( ferror(stdout) ) {
+           fprintf(stderr, "Write Failed : %s\n",strerror(errno) );
+           exit(1);
+         }
+         fwrite(byt_y, 1, len_d, stdout);
+         if( ferror(stdout) ) {
+           fprintf(stderr, "Write Failed : %s\n",strerror(errno) );
+           exit(1);
+         }
+         fprintf(stderr, "\n");
       }
-      fprintf(stderr,"\n");
-    } else {
-       fprintf(stderr,"khan jammed noun: ");
-       c3_y out_y[5];
-       out_y[0] = 0x0;
-       out_y[1] = ( len_d        & 0xff);
-       out_y[2] = ((len_d >>  8) & 0xff);
-       out_y[3] = ((len_d >> 16) & 0xff);
-       out_y[4] = ((len_d >> 24) & 0xff);
-       fwrite(out_y, 1, 5, stdout);
-       if( ferror(stdout) ) {
-         fprintf(stderr, "Write Failed : %s\n",strerror(errno) );
-         exit(1);
-       }
-       fwrite(byt_y, 1, len_d, stdout);
-       if( ferror(stdout) ) {
-         fprintf(stderr, "Write Failed : %s\n",strerror(errno) );
-         exit(1);
-       }
-       fprintf(stderr, "\n");
+    } else {                  //  error, print stack trace
+       u3_pier_punt_goof("eval", u3k(res));
     }
     u3z(res);
-    u3z(sam);
   }
   free(evl_c);
 }
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -1274,7 +1274,7 @@ _cw_eval(c3_i argc, c3_c* argv[])
         {
            fprintf(stderr, "Write Failed : %s\n",strerror(errno) );
            exit(1);
-        }  
+        }
         fprintf(stderr, "\n");
       }
     u3z(res);
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -628,6 +628,7 @@ _cw_usage(c3_c* bin_c)
     "utilities:\n",
     "  %s cram %.*s              jam state:\n",
     "  %s dock %.*s              copy binary:\n",
+    "  %s eval                      evaluate hoon from stdin:\n",
     "  %s grab %.*s              measure memory usage:\n",
     "  %s info %.*s              print pier info:\n",
     "  %s meld %.*s              deduplicate snapshot:\n",
@@ -1159,7 +1160,8 @@ _cw_eval(c3_i argc, c3_c* argv[])
 {
   c3_i ch_i, lid_i;
   c3_w arg_w;
-  c3_o jam_l, khan_l;
+  c3_o jam_l = c3n;
+  c3_o khan_l = c3n;
 
   static struct option lop_u[] = {
     { "loom", required_argument, NULL, c3__loom },
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -1181,14 +1181,17 @@ _cw_eval(c3_i argc, c3_c* argv[])
         }
         u3_Host.ops_u.lom_y = lom_w;
       } break;
+
       case 'j': {
         jam_o = c3y;
         kan_o = c3n;
       } break;
+
       case 'k': {
         jam_o = c3y;
         kan_o = c3y;
       } break;
+
       case '?': {
         fprintf(stderr, "invalid argument\r\n");
         exit(1);
@@ -1236,8 +1239,7 @@ _cw_eval(c3_i argc, c3_c* argv[])
     }
     if ( 0 == u3h(res) ) {  //  successful execution, print output
       u3_pier_tank(0, 0, u3k(u3t(res)));
-    }
-    else {                  //  error, print stack trace
+    } else {                  //  error, print stack trace
        u3_pier_punt_goof("eval", u3k(res));
     }
     u3z(res);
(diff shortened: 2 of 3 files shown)


pkg/vere/main.c
```diff
@@ -628,6 +628,7 @@ _cw_usage(c3_c* bin_c)
     "utilities:\n",
     "  %s cram %.*s              jam state:\n",
     "  %s dock %.*s              copy binary:\n",
+    "  %s eval                      evaluate hoon from stdin:\n",
```
barter-simsum, line 631: Nit: align right hand side with line above

mopfel-winrux: It doesn't line up when printing to the main terminal. It was easier to add the spaces than mess with the for loop a dozen or so lines below.

barter-simsum: Oh, apologies. I see the %.*s where `<pier>` is normally printed and eval is without that.

jalehman:
@barter-simsum would you mind providing an approval if this looks good? 

Are there any potential risks with this that you forsee?

pkg/vere/main.c
```diff
@@ -1170,13 +1175,23 @@ _cw_eval(c3_i argc, c3_c* argv[])
       case c3__loom: {
         c3_w lom_w;
         c3_o res_o = _main_readw(optarg, u3a_bits + 3, &lom_w);
-        if ( (c3n == res_o) || (lom_w < 20) ) {
-          fprintf(stderr, "error: --loom must be >= 20 and <= %u\r\n", u3a_bits + 2);
+        if ( (c3n == res_o) || (lom_w < 24) ) {
+          fprintf(stderr, "error: --loom must be >= 24 and <= %u\r\n", u3a_bits + 2);
```
barter-simsum, line 1179: re: usage of `u3a_bits`. I actually prefer a literal `32` here. The maximum bex of a loom really doesn't have anything to do with `u3a_bits` which just denotes the number of bits in a noun that address the loom.

-- Though it is the pattern followed elsewhere. So probably don't worry about it and I'll just fix it like 51edb14c582817e73069c24c14fa3567e0e2aad0 whenever #164 gets merged into develop

pkg/vere/main.c
```diff
@@ -1159,9 +1160,13 @@ _cw_eval(c3_i argc, c3_c* argv[])
 {
   c3_i ch_i, lid_i;
   c3_w arg_w;
+  c3_o jam_l = c3n;
+  c3_o khan_l = c3n;
 
   static struct option lop_u[] = {
     { "loom", required_argument, NULL, c3__loom },
+    { "j", no_argument, NULL, 'j' },
```
barter-simsum, line 1168: Wondering if we should use a subcommand for this, i.e. `./urbit eval <pier>` akin to `cram`, `dock`, etc. We don't currently use long args with single letters, e.g. `--j`. Perhaps `--gc` is an exception to this rule. `--j` would be easy to mix up with `-j`

barter-simsum: Oh looks like this _is_ and `eval` subcommand. I don't think the "j" and "k" subargs to the command are documented in the help output though. Should they be?

pkg/vere/main.c
```diff
@@ -1216,28 +1231,75 @@ _cw_eval(c3_i argc, c3_c* argv[])
...
-  else {                  //  error, print stack trace
-     u3_pier_punt_goof("eval", u3k(res));
-  }
+    bits = u3s_jam_xeno(res, &len_d, &byt_y);
 
-  u3z(res);
-  u3z(gat);
+    if ( c3n == khan_l ) {
+      fprintf(stderr,"jammed noun: ");
+
```
barter-simsum, line 1270: nit: whitespace. here and throughout changes

pkg/vere/main.c
barter-simsum, line 631: Although you've maintained the alphabetic order by placing this here, I think it should probably be at the top. The missing `<pier>` looks like a mistake otherwise.

pkg/vere/main.c
```diff
@@ -1216,28 +1231,75 @@ _cw_eval(c3_i argc, c3_c* argv[])
...
-     u3_pier_punt_goof("eval", u3k(res));
-  }
+    bits = u3s_jam_xeno(res, &len_d, &byt_y);
 
-  u3z(res);
-  u3z(gat);
+    if ( c3n == khan_l ) {
+      fprintf(stderr,"jammed noun: ");
+       
+      int p=0;
```
barter-simsum, line 1271: declare p as `size_t` since we're iterating. Refactor the while into a for-loop

pkg/vere/main.c
barter-simsum: ... Looking at other subcommands which accept arguments other that `--loom`, e.g. `_cw_next`, they also don't document their arguments in the help output, so don't worry about that.

I _do_ think `j` and `k` should be given in long from though, i.e. `--jam` `--jamkahn`

pkg/noun/vortex.c
```diff
@@ -120,6 +120,31 @@ _cv_nock_wish(u3_noun txt)
   return pro;
 }
 
+/* u3v_wish_n(): text expression with cache.
```
barter-simsum, line 123: update header comment to explain how the new fun differs from `u3v_wish`

barter-simsum:
Everything else lgtm, mostly nits

pkg/vere/main.c
mopfel-winrux: Can you elaborate?

pkg/vere/main.c
mopfel-winrux: I changed it to `--jam` and `--jamkhan`.

pkg/vere/main.c
mopfel-winrux: Moved it to the top

pkg/vere/main.c
mopfel-winrux: Done

pkg/noun/vortex.c
mopfel-winrux: done

mopfel-winrux:
Made changes and pushed them up

pkg/vere/main.c
barter-simsum: There are excess space characters scattered throughout main.c that need to be removed.


Screenshot from 2023-02-06 17-50-44

mopfel-winrux: Removed most of the whitespace I could find.

pkg/vere/main.c
```diff
@@ -1215,29 +1223,63 @@ _cw_eval(c3_i argc, c3_c* argv[])
...
+           exit(1);
+        }
+        fwrite(byt_y, 1, len_d, stdout);
+        if( ferror(stdout))
+        {
+           fprintf(stderr, "Write Failed : %s\n",strerror(errno) );
+           exit(1);
+        }
+        fprintf(stderr, "\n");
+      }
```
ashelkovnykov, line 1279: The indentation in this part is all kinds of messed up:
- body indented by 4 instead of 2
- bodies of `if` statements indented by 3

Also, it's full of inconsistent style, with regards to space before/after conditionals, arguments, and parentheses.

pkg/vere/main.c
```diff
@@ -1159,9 +1160,13 @@ _cw_eval(c3_i argc, c3_c* argv[])
 {
   c3_i ch_i, lid_i;
   c3_w arg_w;
+  c3_o jam_l = c3n;
+  c3_o khan_l = c3n;
```
ashelkovnykov, line 1164: Doesn't conform to naming scheme: `loobean` variables should end with `_o`. Also, it's another naming convention that variable names are three-letter strings (usually consonant-vowel-consonant). I recommend `kan_o`.

ashelkovnykov (changes requested):
More style nits

ashelkovnykov:
Also, I just realized that `--jamkhan` is a misnomer, because it has nothing to do with the Khan vane. What it's doing is creating a `newt` encoded jammed noun; beyond that, the noun could be used for multiple purposes, even by `conn.c`.

For example, the `[1 %urth %meld]` from the description is a command to Vere.

@mopfel-winrux What's the use of the `--jam` option (i.e. who/where needs jammed nouns encoded as hex-strings)? It seems to be consensus that the thin waist of Urbit should be `newt`-encoded jammed nouns.

pkg/vere/main.c
mopfel-winrux: Changed

jalehman:
> @ashelkovnykov `--jamkhan` was the name Joe and I came up with while working on this. I can change it to whatever you'd like.

@joemfb want to weigh in here?

pkg/vere/main.c
mopfel-winrux: I looked through the commit and tried to fix up everything. Is there a formatting guide somewhere I'm not aware of?

ashelkovnykov:
I would remove `--jam` and rename `--jamkhan` to `--jam`. I'd also try to replace the logic in `_cw_eval` that replicates `newt` encoding with `u3_newt_send`. Both of these changes can be done in later commits, if merging this is a priority.

pkg/vere/main.c
ashelkovnykov: There are some conventions in [here](https://github.com/urbit/urbit/blob/master/doc/spec/u3.md). There has been murmur about introducing auto-formatting using some third-party tool, though this was months ago when Tlon was managing the repo. New community ownership might be more interested if it's reintroduced.

ashelkovnykov (approved):

joemfb:
https://github.com/urbit/vere/issues/197#issuecomment-1420789870 has some good context from @ashelkovnykov.

The *noun* is the exclusive, universal data type of urbit. *Jam* is universal noun serialization. And the "newt" wire-framing is how we pass around those serialized nouns in a streaming context. Everything in urbit is a noun, `+jam` is used anytime we want to convert an arbitrary noun to an atom or a byte buffer (nouns are often imported/exported/distributed as "jamfiles"), "newt" is used for all our IPC. The first two will most likely never change, the third is much more contingent (shared-memory IPC is probably the future for mars/urth).

The `--jam` option is useful if you need a lightweight way to generate a jammed noun, or if you're implementing `+jam`, `+cue`, or a new interpreter, or if you're moving nouns between nock interpreters, or if you just want to poke around with implementation details. I don't have an immediate operational use-case for it -- it's just obvious and low-cost.

We'll want to have a `+cue` command as well, maybe taking a hoon mold as text to control pretty-printing. And a version that can also cope with "newt" wire-framing, for handling responses from conn.

The naming of `%khan`/conn remains a sore spot and source of confusion -- the relationship between them is quite minimal.  The most technically correct names for these arguments would be `--jam`/-j` and `--jam-newt` (or `--jam --newt`/`-jn`). Which isn't to say that they're *good*. I'm not sure what would be ...

pkg/vere/main.c
```diff
@@ -1215,29 +1226,60 @@ _cw_eval(c3_i argc, c3_c* argv[])
...
-    res = u3m_soft(0, u3n_kick_on, cor);
-  }
-
-
-  if ( 0 == u3h(res) ) {  //  successful execution, print output
-     u3_pier_tank(0, 0, u3k(u3t(res)));
-  }
-  else {                  //  error, print stack trace
-     u3_pier_punt_goof("eval", u3k(res));
+    u3_noun res = u3v_wish_n(sam);
```
joemfb, line 1249: This evaluation should be virtualized with `u3m_soft()`. This is the reason to add `u3v_wish_n()` instead of just calling `u3v_wish()`.

The implementations of those two should also be combined at some point (and arguably the names should be reversed: `u3v_wish()` and `u3v_wish_c()` would be more consistent with u3 conventions). But that can be done some other time -- it would complicate this PR.

joemfb (changes requested):

pkg/vere/main.c
mopfel-winrux: When I try to replicate the `u3m_soft()` I do for the non jam case the result is incorrect. I think its because I'm jamming a vase but I'm not sure.

joemfb: `u3m_soft(0, u3v_wish_n, sam)`. Then, just as in the previous case, you'll need to branch on success/failure (`0 == u3h(res)`) and print the trace or proceed appropriately.

mopfel-winrux: Whenever I do that I get
```
$ echo "(add 2 2)" | bazel-bin/pkg/vere/urbit eval --jam
loom: mapped 2048MB
lite: arvo formula 2a2274c9
lite: core 4bb376f0
lite: final state 4bb376f0
eval:
jammed noun: 
home: bailing out
Aborted (core dumped)
```

The code change is below:
```
else {
     u3_noun sam = u3i_string(evl_c);
     u3_noun res = u3m_soft(0, u3v_wish_n, sam);
     c3_d bits = 0;
     c3_d len_d = 0;
     c3_y* byt_y;
     if ( 0 == u3h(res) ) {  //  successful execution, print output
        bits = u3s_jam_xeno(res, &len_d, &byt_y);
        if ( c3n == kan_o ) {
          fprintf(stderr,"jammed noun: ");
          for ( size_t p=0; p < len_d; p++ ){
            fprintf(stdout,"\\x%2x", byt_y[p++]);
          } 
          fprintf(stderr,"\n");
 ```
  ...
``` 
     } else {                  //  error, print stack trace
        u3_pier_punt_goof("eval", u3k(res));
     }  
```

Comment (joemfb) [reply pkg/vere/main.c:1249]: You'll want to `u3s_jam_xeno(u3t(res) ...)`. But I don't see why anything should crash here. Can you reproduce that in a debugger with a breakpoint on `u3m_bail()`?

Comment (mopfel-winrux) [reply pkg/vere/main.c:1249]:
```
Breakpoint 1 at 0x43bdf0: file pkg/noun/manage.c, line 688.
(gdb) run eval --jam < ~/add.hoon
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /home/amadeo/.cache/bazel/_bazel_amadeo/bfa2b923decdeb8d329ce362a61e8ce7/execroot/__main__/bazel-out/k8-fastbuild/bin/pkg/vere/urbit eval --jam < ~/add.hoon
loom: mapped 2048MB
lite: arvo formula 2a2274c9
lite: core 4bb376f0
lite: final state 4bb376f0
eval:
jammed noun: 

Breakpoint 1, u3m_bail (how=how@entry=1819635558) at pkg/noun/manage.c:688
688	  if ( &(u3H->rod_u) == u3R ) {
(gdb) bt
#0  u3m_bail (how=how@entry=1819635558) at pkg/noun/manage.c:688
#1  0x0000000000428414 in _me_lose_north (dog=2150035627) at pkg/noun/allocate.c:1375
#2  0x0000000000402d83 in _cw_eval (argv=<optimized out>, argc=<optimized out>)
    at pkg/vere/main.c:1286
#3  _cw_utils (argv=<optimized out>, argc=<optimized out>) at pkg/vere/main.c:2029
#4  main (argc=<optimized out>, argv=<optimized out>) at pkg/vere/main.c:2064
(gdb) 
```

The relevent part of the eval loop is below:
```
1281     } else {                  //  error, print stack trace
1282        u3_pier_punt_goof("eval", u3k(res));
1283     }
1284 
1285     u3z(res);
1286     u3z(sam);
1287   }
1288   free(evl_c);
1289 }
```

I think I'm freeing `sam` when I dont need to because I'm using it in `u3m_soft()`. I removed line 1286 and it seems to stop crashing

mopfel-winrux: Pushed up the changes

joemfb: Good catch!

joemfb (approved):
this will do

further refactoring, argument tweaks/renames, or additional behaviors (jam into a file vs print the bytes) can come in as needed.

PR #142, opened 2023-01-19
Add debug configuration to build system

mcevoypeter:
## Description

Resolves #83.

## Testing

```console
$ bazel build :urbitdbg
```
When running the resulting binary under `gdb`, no arguments are optimized out, as expected with an optimization level of `-O0` (set by the debug configuration).

```console
$ gdb ./urbit
GNU gdb (GDB) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./urbit...
(gdb) list pkg/noun/allocate.h:310
305                                  ? (c3_w)(r->mat_p - r->cap_p) \
306                                  : (c3_w)(r->cap_p - r->mat_p) )
307
308     #     define  u3a_north_is_senior(r, dog) \
309                     __((u3a_to_off(dog) < r->rut_p) ||  \
310                            (u3a_to_off(dog) >= r->mat_p))
311
312     #     define  u3a_north_is_junior(r, dog) \
313                     __((u3a_to_off(dog) >= r->cap_p) && \
314                            (u3a_to_off(dog) < r->mat_p))
(gdb) list pkg/noun/urth.c:371
366     /* _cu_realloc(): hash-cons roots off-loom, reallocate on loom.
367     */
368     static ur_nref
369     _cu_realloc(FILE* fil_u, ur_root_t** tor_u, ur_nvec_t* doc_u)
370     {
371     #ifdef U3_MEMORY_DEBUG
372       c3_assert(0);
373     #endif
374
375       //  bypassing page tracking as an optimization
(gdb) set args zod
(gdb) r
Starting program: /home/tlon/code/gh/urbit/vere/urbit zod

This GDB supports auto-downloading debuginfo from the following URLs:
https://debuginfod.archlinux.org
Enable debuginfod for this session? (y or [n]) n
Debuginfod has been disabled.
[...]

PR #164, opened 2023-01-25
Pointer Compression to enable 8G Loom

### LINKED ISSUE #163: Pointer compression to enable 8G loom
DWORD aligned pointers compressed to give us an extra virtual bit with which to address memory and a migration routine to move prior piers to this new state.

barter-simsum:
### Description

Resolves #163

In fulfillment of https://urbit.org/grants/loom-pointer-compression

#### Basic brass pill fakezod boot benchmark - x86_64 linux

Pay primary attention to `Elapsed (wall clock) time`, `Maximum resident set size
(kbytes)` and `Major (requiring I/O) page faults`

##### Takeaway

We expected increased memory usage because this is naturally a tradeoff of
alignment. Do note, that runs (2) and (3) included changes to align the _stack_
as well as the _heap_. In the run without stack alignment (4) you can see that
stack alignment has no effect on max RSS -- at least when booting from a
pill. From some basic evaluation in gdb I've done in the past, I expect stack
usage when DWORD-aligned to increase by ~50% (rather than a theoretical
100%). Stack usage is quite small compared to heap usage however, so you
shouldn't expect to see this reflected in maximum RSS. Overall maximum resident
memory increased by about ~33%.

The number of major pagefaults encountered during a brass boot is roughly equal
to prior.

The elapsed (wall clock) time difference between (2) and (3) is essentially
zero. There is essentially no performance gained by the virtual bit size being a
compile-time constant.

There is a small latency cost in the current DWORD-aligned heap allocation
implementation as compared to a runtime that doesn't require allocations to be
aligned. Compare the elapsed times of (1) -- 2:21.09 or 141.09 s -- and (2) --
2:23.50 or 143.50 s -- result: ~1.7% increased latency. If you look at run (4)
however, which excluded stack alignment changes -- 2:22.55 or 142.55 s --, we
split the difference at ~1.0% increased latency. Note, this _is_ repeatable and
the 1% difference isn't random. Running the same program over again on the same
system exhibits tiny variance.

##### 1) -O3 no pointer compression vere/develop

A run from the HEAD of vere/develop

commit 7c890c3350e3b10e918ea04b92a34dc25af1552d

```
Command being timed: "./urbit -t -q -F zod -B brass.pill -c zod"
User time (seconds): 1.25
System time (seconds): 0.03
Percent of CPU this job got: 0%
Elapsed (wall clock) time (h:mm:ss or m:ss): 2:21.09
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 148036
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 3492
Minor (reclaiming a frame) page faults: 6188
Voluntary context switches: 68
Involuntary context switches: 5
Swaps: 0
File system inputs: 14866
File system outputs: 21544
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

##### 2) -O3 compiletime determined virtual bit size i/163/pointer-compression

State of pointer compression work prior to migration (where concessions to
runtime determined virtual bit size were made)

commit 4083f1c660c1db682516b57e32780fd59bff7eaf

```
Command being timed: "./urbit -t -q -F zod -B brass.pill -c zod"
User time (seconds): 1.39
System time (seconds): 0.04
Percent of CPU this job got: 1%
Elapsed (wall clock) time (h:mm:ss or m:ss): 2:23.50
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 197176
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 3487
Minor (reclaiming a frame) page faults: 6219
Voluntary context switches: 68
Involuntary context switches: 2
Swaps: 0
File system inputs: 14866
File system outputs: 21544
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

##### 3) -O3 runtime determined virtual bit size i/163/pointer-compression

Current state of pointer compression work -- after implementation of migration
and runtime determined virtual bit size concession

commit 8dffe067e133ac895a5f8befb67762b58116ae75:

```
Command being timed: "./urbit -t -q -F zod -B brass.pill -c zod"
User time (seconds): 1.40
System time (seconds): 0.06
Percent of CPU this job got: 1%
Elapsed (wall clock) time (h:mm:ss or m:ss): 2:23.52
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 197200
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 3489
Minor (reclaiming a frame) page faults: 7242
Voluntary context switches: 69
Involuntary context switches: 4
Swaps: 0
File system inputs: 14866
File system outputs: 21544
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

##### 4) -O3 runtime determined virtual bit size _WITHOUT_ stack alignment barter-simsum/pointer-compression-no-align-stack

commit 8b0438ab3bb00cf662c7cef6128f441210728f85

```
Command being timed: "./urbit -t -q -F zod -B brass.pill -c zod"
User time (seconds): 1.42
System time (seconds): 0.06
Percent of CPU this job got: 1%
Elapsed (wall clock) time (h:mm:ss or m:ss): 2:22.55
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 197204
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 3492
Minor (reclaiming a frame) page faults: 6221
Voluntary context switches: 68
Involuntary context switches: 3
Swaps: 0
File system inputs: 14866
File system outputs: 21544
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

##### FINAL BENCHMARK BEFORE MERGE:

This was run after some fairly significant changes to minimize malloc padding,
fix memory corruption when run with `U3_MEMORY_DEBUG`, and more.

It was agreed to keep stack alignment out of this PR as it currently isn't used
and costs us a bit of latency.

Runtime vs compiletime determined pointer compression still shows no latency
difference on the x86 linux machine tested (ddr4 memory). On an m2 mac air,
there _was_ a 5% latency increase from compiletime to runtime pointer
compression. This may be fixed later and would not necessitate another
migration.

The additional free list sanity checking done in `u3a_loom_sane` introduces
negligible latency in `u3e_save`. On a relatively fragmented heap, it only takes
60ms to complete. This will be kept in order to detect _some_ memory corruption
if it occurs and prevent that corruption from propagating to disk.

A brass pill boot was performed off of
13e0b43d8da4bdd318fcd4e3d3610caa3af4608a. Observe there is no regression in the
Elapsed (wall clock) time statistic. Further, the maximum resident set size has
been reduced by 25% back to its pre pointer compression size (150M). This is
likely due to a decrease in the average allocation's padding.

Lastly, total sweep size was compared between a freshly booted pier without
pointer compression and with pointer compression post migration. There is no
noticeable increase in the overall size of allocations.

```
Command being timed: "./burbit -t -q -F zod -B brass.pill -c brasspillbench"
User time (seconds): 1.27
System time (seconds): 0.04
Percent of CPU this job got: 0%
Elapsed (wall clock) time (h:mm:ss or m:ss): 2:21.49
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 150088
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 3490
Minor (reclaiming a frame) page faults: 6188
Voluntary context switches: 64
Involuntary context switches: 4
Swaps: 0
File system inputs: 14866
File system outputs: 21544
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```


mcevoypeter:
1. Worst case scenario is we use a directory prefix for include paths, so `#include "noun.h"` becomes `#include "noun/noun.h"` and similar, allowing us to have `"vere/version.h"` and `"noun/version.h"`.
2. #142 is relevant here. You'd only need to add `build --copt='-DC3DBG'` to `.bazelrc`.

pkg/vere/main.c
```diff
@@ -367,9 +367,9 @@ _main_getopt(c3_i argc, c3_c** argv)
       }
       case c3__loom: {
         c3_w lom_w;
-        c3_o res_o = _main_readw(optarg, u3a_bits + 3, &lom_w);
+        c3_o res_o = _main_readw(optarg, u3a_bits_max+1, &lom_w);
         if ( (c3n == res_o) || (lom_w < 20) ) {
-          fprintf(stderr, "error: --loom must be >= 20 and <= %u\r\n", u3a_bits + 2);
+          fprintf(stderr, "error: --loom must be >= 20 and <= %u\r\n", u3a_bits_max);
```
joemfb, line 372: This change needs to also be applied to all the other subcommands that use `--loom` as an argument (`_cw_meld()`, `_cw_pack()`, &c).

joemfb (changes requested):

pkg/vere/main.c
barter-simsum: done.

joemfb (approved):
this will do

barter-simsum:
Don't merge yet doing a few more tests on live ships.

barter-simsum:
Ran into some weird behavior on a comet when running the migration. Indicates free list corruption. Can't repro now and unfortunately didn't have a debugger attached. Assuming the free list corruption is a false positive, though don't know what could put the ship into that state and also what between 5 and 6 could magically resolve it.

1) Create comet with urbit-v1.13

2) echo edge > testcomet/.bin/pace

3) prep and next -> version incremented to 1.16. Run on that version

4) next again and now on version 1.22-f2df391. Run on that version

5) Run with pointer compression enabled binary

Get the following error repeatedly:
```
live: loaded: MB/264.437.760
loom: migration running. This may take several minutes to perform.
loom: have version: 1 migrating to version: 2
Assertion '!"loom: corrupt"' failed in pkg/noun/allocate.c:192
Assertion failed: !"loom: corrupt" (pkg/noun/allocate.c: _box_detach: 192)
                                                                          pier: serf unexpectedly shut down
```

6) Run pointer compression binary with -L, suddenly the migration works.

Between 5 and 6, I also ran one more time following the free list corruption messages on 1.22-f2df391. Unsure if that could have had any effect and can't repro at all now. Nothing special about this comet. It was freshly booted.

philipcmonk:
Seeing corruption like that certainly gives me pause.  I feel like it would be worth writing a function that sanity checks the snapshot (free list at minimum, could also traverse the nouns) and keeping it running so that we'll detect more corruption (the worst is when the corruption lasts so long that you write another snapshot before it triggers anything).  If it's fast enough, it should run every time before we write a snapshot, but at the least it will be useful for debugging.

barter-simsum:
@philipcmonk Sorry should have clarified. This is actually happening in `|pack` which is called prior to starting the migration. This would be the equivalent of detecting corruption when issuing pack as a subcommand.

Here's the gist of what happened. Only reproed once on a fresh comet. Hasn't since. Unsure of root cause

1) Briefly ran pier on 1.22-f2df391. Exited.
2) Ran on binary containing compression/migration code. Pack fails with freelist corruption.
3) Repeat (2) multiple times. Same result.
4) Rerun briefly on 1.22-f2df391.
5) Repeat (2). Works fine this time.

So apparently the corrupted image changed between (2) and (5) -- I suppose with the rerun of (4).

philipcmonk:
`|pack` will obliterate all free lists, so I'm not sure how it was noticing free list corruption.  Maybe in the initial "clear some caches" step?  I would want to know more specifically when it encountered corruption and ideally how long it had been there.

philipcmonk:
Snapshots can be corrupted for a long time without us noticing, which is why a checkpoint-is-sane function would be informative.

barter-simsum:
I agree adding a loom sanity check prior to writing the snapshot would be useful in tracking down the cause of this, even if it's somewhere else in vere and not in this PR.

And yes, the free list corruption must have been detected as part of the `u3m_reclaim` call in `|pack`. The migration has enough logging to tell it definitely occurred in pack though.

barter-simsum:
@philipcmonk @belisarius222
I found the issue. It was introduced in an unrelated refactor of `_box_slot` 1c43861d5a9ce6cd8667d10be755c13f5c217bf5. It ultimately led to an OBO for certain `siz_w`s. The `siz_w` is repeatedly divided by 2 (because our freelists size by powers of two). We add one though before right shifting so odd `siz_w`s round up once. Since all boxes on the migrated loom are DWORD multiples (even), it wouldn't matter. But when freeing boxes as part of the pack/migration on the old loom, if we attempt to detach a head box from a freelist and it happens to have a `siz_w` equal to `(2^n)-1 for n=5..u3a_fbox_no-1`, we'll get an incorrect box slot which is caught by the last assert in `_box_detach`.

So not exactly heap corruption. If the assert weren't there, we would have simply had the chance of moving a freelist matching the criteria above one slot lower than it should be -- which is fine because the pack immediately following would have cleared all that freespace anyway.

barter-simsum:
Added `u3a_loom_sane`. Currently gets called after the image is loaded, after the migration, and every time `u3e_save` is called before we write to disk. Don't have a an easy way to profile the bazel-built vere though. Maybe a brass pill boot time would do -- or will that not invoke `u3e_save`?

philipcmonk:
Nice!  We basically just want `u3a_loom_sane` to not add a ton of time to `u3e_save`, so I'd be inclined to just print out the time before/after `u3e_save` overall and `u3a_loom_sane` specifically, and the run that on (-L copies of) some large, infrequently packed ships and see if anything bad happens.  My guess is it takes a lot less time than writing to disk.

barter-simsum:
Benchmarked a fresh comet. Increased `u3e_save` latency by about 1%. Can't just run on any old freelist fragmented pier though, because the migration will of course clear all the fragmentation. Going to just use the comet and let it run for a while to try to increase fragmentation.

barter-simsum:
On a pretty fragmented ship like mine, the freelist traversal increased `u3e_save` latency by 20%! Maybe doing this is still worth it, but I'm surprised by the latency difference.

belisarius222:
What are the absolute numbers for how long it took u3e_save() to run on your fragmented ship?  My guess is the 20% doesn't matter, but if it's starting to get really slow, we should be careful.

barter-simsum:
@belisarius222 
Apologies, was calculating the time diff wrong. It looks like on the fairly fragmented pier of ~barter-simsum, `u3a_loom_sane` takes only about 60ms. The amount of time that `u3e_save` takes varies of course by how much data it has to write. If you don't do any allocation following the last snapshot, it's roughly 3 seconds on my x86 linux machine. If you do a bit of allocation, possibly up to 10 seconds.

The longest runtime for `u3e_save` is going to be following a migration where we have to write the entire filled amount to disc. On a 700MB pier, this took about 30 seconds on my machine. Piers approaching 4G might take multiple minutes.

Anyway, given the average of 60 milliseconds to validate the freelist, I say we keep this in.

barter-simsum:
rebased `_box_slot` fix into bcac1d6d2475550d734cc47c2dca900a32b49112 where the regression was originally introduced.

Please read "FINAL BENCHMARK BEFORE MERGE" section in PR description.

This is ready to merge into develop. Whoever makes the merge commit, please keep the default merge commit description which should be prepopulated with the PR description.

barter-simsum:
Just fixing `-Wformat` complaint when printing `u3a_bits_max`

belisarius222 (approved):

PR #158, opened 2023-01-24
I/157/event log replay restoration

sigilante:
### Description

Resolves #157.

Copies back in replay-related changes from 1.14, without guard page code.

PR #165, opened 2023-01-25
chop: offline event log truncation

### LINKED ISSUE #122: Implement `chop` (event log truncation tool)
Write a minimally viable event log truncation program. Intended for power users with particularly large piers, it should:
- [x] Safely reduce a ship's pier size
- [x] Operates only on a ship which has been gracefully shutdown (i.e., has no unprocessed events in its queue)
  - [x] Performs a graceful shutdown if the ship is currently running (maybe?)
- [x] Exports the event log to a backup file for optional archiving/compressing/deleting

Upon completion, a more permanent, tunable, and feature-rich event log truncation system could be built.

References:
- [event log truncation on the roadmap](https://roadmap.urbit.org/project/event-log-truncation)
- https://github.com/urbit/urbit/pull/5701
- [`chop` design](https://gist.github.com/matthew-levan/1db63ed7806d6fdcd1dc46759e84d06b)
- https://github.com/urbit/vere/pull/165

Tagging: @mcevoypeter @joemfb @zalberico

matthew-levan:
## `chop`

`urbit chop <pier>` implements a simple, offline **event log truncation**[^1] tool.

`chop` gracefully stops the given pier (if running), backs up the current snapshot to `<pier>/.urb/bhk`, makes sure a current snapshot exists (i.e., is fully written to disk in `chk/*.bin` with no existing patch files), reads the metadata and the last event from the pier's event log, initializes a fresh event log in the `<pier>/.urb/log/chop` directory,  writes the metadata and last event from the original log into the fresh one, renames the original event log to `<pier>/.urb/log/chop/data_<first>_<last>.mdb.bak` where `first` and `last` are the first and last event numbers from the event log, and exits. 

Pilots are then free to move, archive, or delete their `.bak` event log file, resume normal operation of their ship, and enjoy the many benefits of lowered disk pressure and any reductions in associated hosting costs.

I've tested `chop` successfully on my own planet `~mastyr-bottec` (multiple times), three different comets (all fresh), and multitudes of fake galaxies.

Resolves #122.

Note: `knit`, which is the "undo" button for `chop`, is being implemented in its own PR #184.

[^1]: https://roadmap.urbit.org/project/event-log-truncation

### DIFF
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -1103,7 +1103,7 @@ u3e_save(void)
   _ce_patch_free(pat_u);
   _ce_patch_delete();
 
-  _ce_backup();
+  u3e_backup();
 }
 
 /* u3e_live(): start the checkpointing system.
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -1041,6 +1041,12 @@ u3e_backup(c3_o ovw_o)
   fprintf(stderr, "loom: image backup complete\r\n");
 }
 
+c3_o
+u3e_curr(void)
+{
+  return __(0 == _ce_patch_open());
+}
+
 /*
   u3e_save(): save current changes.
 
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -1044,7 +1044,12 @@ u3e_backup(c3_o ovw_o)
 c3_o
 u3e_curr(void)
 {
-  return __(0 == _ce_patch_open());
+  u3_ce_patch* pat_u = _ce_patch_open();
+  if ( 0 != pat_u ) {
+    _ce_patch_free(pat_u);
+    return c3n;
+  }
+  return c3y;
 }
 
 /*
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -992,10 +992,10 @@ _ce_image_copy(u3e_image* fom_u, u3e_image* tou_u)
   return c3y;
 }
 
-/* u3e_backup();
+/* _ce_backup();
 */
 void
-u3e_backup(void)
+u3e_backup(c3_o ovw_o)
 {
   u3e_image nop_u = { .nam_c = "north", .pgs_w = 0 };
   u3e_image sop_u = { .nam_c = "south", .pgs_w = 0 };
@@ -1004,7 +1004,7 @@ u3e_backup(void)
 
   snprintf(ful_c, 8192, "%s/.urb/bhk", u3P.dir_c);
 
-  if ( c3_mkdir(ful_c, 0700) ) {
+  if ( (c3n == ovw_o) && c3_mkdir(ful_c, 0700) ) {
     if ( EEXIST != errno ) {
       fprintf(stderr, "loom: image backup: %s\r\n", strerror(errno));
     }
@@ -1038,6 +1038,7 @@ u3e_backup(void)
 
   close(nop_u.fid_i);
   close(sop_u.fid_i);
+  fprintf(stderr, "loom: image backup complete\r\n");
 }
 
 /*
@@ -1103,7 +1104,7 @@ u3e_save(void)
   _ce_patch_free(pat_u);
   _ce_patch_delete();
 
-  u3e_backup();
+  u3e_backup(c3n);
 }
 
 /* u3e_live(): start the checkpointing system.
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -994,7 +994,7 @@ _ce_image_copy(u3e_image* fom_u, u3e_image* tou_u)
 
 /* _ce_backup();
 */
-void
+c3_o
 u3e_backup(c3_o ovw_o)
 {
   u3e_image nop_u = { .nam_c = "north", .pgs_w = 0 };
@@ -1007,22 +1007,25 @@ u3e_backup(c3_o ovw_o)
   if ( (c3n == ovw_o) && c3_mkdir(ful_c, 0700) ) {
     if ( EEXIST != errno ) {
       fprintf(stderr, "loom: image backup: %s\r\n", strerror(errno));
+      fprintf(stderr, "loom: image backup failed\r\n");
+      return c3n;
     }
-    return;
   }
 
   snprintf(ful_c, 8192, "%s/.urb/bhk/%s.bin", u3P.dir_c, nop_u.nam_c);
 
   if ( -1 == (nop_u.fid_i = c3_open(ful_c, mod_i, 0666)) ) {
     fprintf(stderr, "loom: c3_open %s: %s\r\n", ful_c, strerror(errno));
-    return;
+    fprintf(stderr, "loom: image backup failed\r\n");
+    return c3n;
   }
 
   snprintf(ful_c, 8192, "%s/.urb/bhk/%s.bin", u3P.dir_c, sop_u.nam_c);
 
   if ( -1 == (sop_u.fid_i = c3_open(ful_c, mod_i, 0666)) ) {
     fprintf(stderr, "loom: c3_open %s: %s\r\n", ful_c, strerror(errno));
-    return;
+    fprintf(stderr, "loom: image backup failed\r\n");
+    return c3n;
   }
 
   if (  (c3n == _ce_image_copy(&u3P.nor_u, &nop_u))
@@ -1034,11 +1037,14 @@ u3e_backup(c3_o ovw_o)
     c3_unlink(ful_c);
     snprintf(ful_c, 8192, "%s/.urb/bhk", u3P.dir_c);
     c3_rmdir(ful_c);
+    fprintf(stderr, "loom: image backup failed\r\n");
+    return c3n;
   }
 
   close(nop_u.fid_i);
   close(sop_u.fid_i);
   fprintf(stderr, "loom: image backup complete\r\n");
+  return c3y;
 }
 
 /*
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -992,10 +992,10 @@ _ce_image_copy(u3e_image* fom_u, u3e_image* tou_u)
   return c3y;
 }
 
-/* _ce_backup();
+/* u3e_backup();
 */
-static void
-_ce_backup(void)
+void
+u3e_backup(void)
 {
   u3e_image nop_u = { .nam_c = "north", .pgs_w = 0 };
   u3e_image sop_u = { .nam_c = "south", .pgs_w = 0 };
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -1041,17 +1041,6 @@ u3e_backup(c3_o ovw_o)
   fprintf(stderr, "loom: image backup complete\r\n");
 }
 
-c3_o
-u3e_curr(void)
-{
-  u3_ce_patch* pat_u = _ce_patch_open();
-  if ( 0 != pat_u ) {
-    _ce_patch_free(pat_u);
-    return c3n;
-  }
-  return c3y;
-}
-
 /*
   u3e_save(): save current changes.
 
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -992,7 +992,7 @@ _ce_image_copy(u3e_image* fom_u, u3e_image* tou_u)
   return c3y;
 }
 
-/* _ce_backup();
+/* u3e_backup();
 */
 c3_o
 u3e_backup(c3_o ovw_o)
@@ -1007,16 +1007,14 @@ u3e_backup(c3_o ovw_o)
   if ( (c3n == ovw_o) && c3_mkdir(ful_c, 0700) ) {
     if ( EEXIST != errno ) {
       fprintf(stderr, "loom: image backup: %s\r\n", strerror(errno));
-      fprintf(stderr, "loom: image backup failed\r\n");
-      return c3n;
     }
+    return c3n;
   }
 
   snprintf(ful_c, 8192, "%s/.urb/bhk/%s.bin", u3P.dir_c, nop_u.nam_c);
 
   if ( -1 == (nop_u.fid_i = c3_open(ful_c, mod_i, 0666)) ) {
     fprintf(stderr, "loom: c3_open %s: %s\r\n", ful_c, strerror(errno));
-    fprintf(stderr, "loom: image backup failed\r\n");
     return c3n;
   }
 
@@ -1024,7 +1022,6 @@ u3e_backup(c3_o ovw_o)
 
   if ( -1 == (sop_u.fid_i = c3_open(ful_c, mod_i, 0666)) ) {
     fprintf(stderr, "loom: c3_open %s: %s\r\n", ful_c, strerror(errno));
-    fprintf(stderr, "loom: image backup failed\r\n");
     return c3n;
   }
 
--- a/pkg/vere/db/lmdb.c
+++ b/pkg/vere/db/lmdb.c
@@ -386,50 +386,6 @@ u3_lmdb_read_one(MDB_env* env_u,
   }
 }
 
-/* u3_lmdb_read_meta_one(): load [key_c] into [val_u].
-*/
-c3_o
-u3_lmdb_read_meta_one(MDB_env*    env_u,
-                      MDB_val*    val_u,
-                      const c3_c* key_c)
-{
-  MDB_txn* txn_u;
-  MDB_dbi  mdb_u;
-  c3_w     ret_w;
-
-  //  create a read transaction
-  //
-  if ( (ret_w = mdb_txn_begin(env_u, 0, MDB_RDONLY, &txn_u)) ) {
-    mdb_logerror(stderr, ret_w, "lmdb: meta read one: txn_begin fail");
-    return c3n;
-  }
-
-  //  open the database in the transaction
-  //
-  if ( (ret_w =  mdb_dbi_open(txn_u, "META", 0, &mdb_u)) ) {
-    mdb_logerror(stderr, ret_w, "lmdb: meta read one: dbi_open fail");
-    mdb_txn_abort(txn_u);
-    return c3n;
-  }
-
-  //  read by string key
-  {
-    MDB_val key_u = { .mv_size = strlen(key_c), .mv_data = (void*)key_c };
-
-    if ( (ret_w = mdb_get(txn_u, mdb_u, &key_u, val_u)) ) {
-      mdb_logerror(stderr, ret_w, "lmdb: meta read one failed");
-      mdb_txn_abort(txn_u);
-      return c3n;
-    }
-    else {
-      //  read-only transactions are aborted when complete
-      //
-      mdb_txn_abort(txn_u);
-      return c3y;
-    }
-  }
-}
-
 /* u3_lmdb_save(): save [len_d] events starting at [eve_d].
 */
 c3_o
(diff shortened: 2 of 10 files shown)


pkg/vere/db/lmdb.c
```diff
@@ -29,11 +29,24 @@ intmax_t mdb_get_filesize(mdb_filehandle_t han_u);
...
+//      - backup an environment
+//      - delete an environment
 //      - read/save metadata
 //      - read the first and last event numbers
 //      - read/save ranges of events
 //
 
+/* u3_lmdb_backup(): backup the lmdb environment.
+*/
+c3_o u3_lmdb_backup(MDB_env* env_u, const c3_c* pax_c)
```
belisarius222, line 41: style: return type needs to be on the previous line

pkg/vere/db/lmdb.c
```diff
@@ -29,11 +29,24 @@ intmax_t mdb_get_filesize(mdb_filehandle_t han_u);
...
+{
+  if ( 0 != mdb_env_copy(env_u, pax_c) ) {
```
matthew-levan, line 43: I took a look at the [implementation](https://github.com/LMDB/lmdb/blob/3947014aed7ffe39a79991fa7fb5b234da47ad1a/libraries/liblmdb/mdb.c#L10517) of this and it uses `write()`, which will be a lot slower than just using `rename()` from `stdio.h`.

matthew-levan:
This code has now been tested successfully on fake ships and a comet. Thanks to @mopfel-winrux for helping me clean the code up a little, but it still needs more polish. 

Additionally, since chopping one's pier leaves a backup event log file behind (and pilots may indeed chop their piers multiple times throughout its life), we also had the idea to include an `urbit knit` subcommand in this PR that would accept a list of event log backup files and insert their events back into a single `.mdb` file of the user's choice (whether that's a new one or the existing `data.mdb`). This command would allow previously chopped event logs to be restored and replayed, if desired/necessary.

pkg/vere/main.c
```diff
@@ -1820,19 +1828,19 @@ _cw_chop(c3_i argc, c3_c* argv[])
...
   }
-  ver_n = u3i_bytes(ver.mv_size, ver.mv_data);
-  ver.mv_data = &ver_n;
-  who_n = u3i_bytes(who.mv_size, who.mv_data);
-  who.mv_data = &who_n;
-  fak_n = u3i_bytes(fak.mv_size, fak.mv_data);
-  fak.mv_data = &fak_n;
-  lif_n = u3i_bytes(lif.mv_size, lif.mv_data);
-  lif.mv_data = &lif_n;
+  c3_d* ver_d = c3_malloc(ver.mv_size);
```
belisarius222: Nitpick: when you have operations 1 and 2 to perform on items a,b,c,d, it's usually better to do:
```
1 a; 2 a
1 b; 2 b
1 c; 2 c
1 d; 2 d
```
rather than
```
1 a; 1 b; 1 c; 1 d
2 a; 2 b; 2 c; 2 d
```
because it's easier to pull out `1 then 2` into a function later.

belisarius222 (commented):
This looks good to me.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,173 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+    u3_king_bail();
+  }
+  c3_d ver_d = *(c3_d*)ver.mv_data;
+
+  MDB_val who;
+  if ( c3y != u3_lmdb_read_meta_one(env_u, &who, "who") ) {
+    fprintf(stderr, "chop: failed to read who\r\n");
+    u3_king_bail();
+  }
+  c3_d who_d = *(c3_d*)who.mv_data;
```
mcevoypeter, line 1825: This doesn't look quite right. The value stored under the key `"who"` in the database is an atom that needs to be transformed into `c3_d[2]`. https://github.com/urbit/vere/tree/develop/pkg/vere/disk.c#L573 is a useful reference.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,173 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+    u3_king_bail();
+  }
+
+  // get the metadata
+  MDB_val ver;
+  if ( c3y != u3_lmdb_read_meta_one(env_u, &ver, "version") ) {
+    fprintf(stderr, "chop: failed to read version\r\n");
+    u3_king_bail();
+  }
+  c3_d ver_d = *(c3_d*)ver.mv_data;
```
mcevoypeter, line 1818: The value stored under the `"version"` key doesn't appear to be a double word. Instead, it looks either an unsigned 8-bit or unsigned 32-bit value.

mcevoypeter (dismissed):
Thanks for churning this out so quickly. The overall work looks great, but I have a number of comments and questions on small details.

pkg/vere/main.c
belisarius222: Good catch.  @matthew-levan I thought in a previous version of this code, you just left the metadata values as buffers, without converting them.  What's the reason to convert them to some more specific datatype?  I thought we just needed to write the same values into the new db.  Or am I remembering wrong?

Even simpler: can we not just use `u3_disk_read_meta()` and `u3_disk_save_meta()` directly?  Something like this:

```C
{
  u3_disk* log_u;
  c3_d     who_d[2];
  c3_o     fak_o;
  c3_w     lif_w;
  
  if ( c3n == u3_disk_read_meta(log_u, who_d, &fake_o, &lif_w) ) {
    //  error handling
  }
  
  if ( c3n == u3_disk_save_meta(log_u, who_d, fake_o, lif_w) ) {
    //  error handling
  }
}
```

We'd need to do the read and save in the same function to avoid sticking these values in some heap-allocated struct, but I think that's doable here.

pkg/vere/db/lmdb.c
```diff
@@ -342,6 +342,50 @@ u3_lmdb_read(MDB_env* env_u,
...
+    }
+    else {
+      //  read-only transactions are aborted when complete
+      //
+      mdb_txn_abort(txn_u);
+      return c3y;
+    }
+  }
+}
+
```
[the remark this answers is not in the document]
belisarius222: Well, I had originally advised @matthew-levan to just use a callback, but it did strike me as a bit odd that our lmdb wrapper has a synchronous write interface and an asynchronous read interface.  For reading a whole sequence of values, it makes sense to supply a callback that gets fired on each value, but it seems useful to me to have a synchronous API in both reading and writing -- or only async APIs, but I was surprised to find both.

Having a synchronous API here allows client code to be much more straightforward -- it doesn't need a heap-allocated struct to track intermediate products.  Maybe it's more lines of code total, but I'd argue it's less mental overhead.

A counterargument is that future database APIs, such as anything that involves a distributed database, are likely to be async-only.  Then any code built on synchronous reads would have to be significantly reorganized later -- of course, that's also true for any code we have now that uses the synchronous write API, but that doesn't mean we have to make the problem worse.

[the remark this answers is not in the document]
matthew-levan: The reason I did it was because it allowed me to read the data I needed for `chop` much more quickly in terms of mental effort. 

I started working with the existing asynchronous functions and wasn't able to code a solution as quickly as I would've liked, so I trashed the approach and wrote these sync functions. 

LMDB's API is synchronous and quite straightforward, and (as a new contributor to `vere`, of course) I humbly submit that we should have similarly synchronous and straightforward wrappers to work with it.

I'm more than happy to be wrong on this point and remove the additional functions in `lmdb.c`, but it will also take some more work to implement `chop` using the original async functions.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,173 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+  // see disk.c L885
+  const size_t siz_i =
+  // 500 GiB is as large as musl on aarch64 wants to allow
+  #if (defined(U3_CPU_aarch64) && defined(U3_OS_linux))
+    0x7d00000000;
+  #else
+    0x10000000000;
+  #endif
+  c3_c log_c[8193];
+  snprintf(log_c, 8192, "%s/.urb/log", u3_Host.dir_c);
```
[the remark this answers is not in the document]
matthew-levan: Ok, no problem. I used the magic numbers because I found the same pattern extensively throughout other files in the codebase, like in `events.c` for example.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,173 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+  MDB_env* env_u = u3_lmdb_init(log_c, siz_i);
+
+  // get the first/last event numbers
+  c3_d fir_d, las_d;
+  if ( c3n == u3_lmdb_gulf(env_u, &fir_d, &las_d) ) {
+    fprintf(stderr, "disk: failed to load latest event from database\r\n");
+    u3_king_bail();
```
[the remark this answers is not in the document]
matthew-levan: Sorry, there wasn't much rationale. I mistakenly thought I saw it being used in lieu of `exit()` but upon second review I don't know why. I will replace these with regular exits.

matthew-levan:
@mcevoypeter and @belisarius222 thank you very much for the thorough review/suggestions. I've pushed a commit that implements the suggested changes, please take a look at your earliest convenience. I've successfully tested the latest on a fakezod and a comet so far.

pkg/vere/db/lmdb.c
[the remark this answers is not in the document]
joemfb: `u3_lmdb_read()` takes a callback, but invokes it synchronously. It's setup this way so to support reading multiple key/value pairs without committing to a particular representation or allocation strategy. The `disk.c` read API on top of this *is* async.

NB: there's an iterator-based read API that @belisarius222 and I wrote some time ago in #158.

pkg/vere/main.c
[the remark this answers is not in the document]
mcevoypeter: I appreciate the adherence to existing norms, but this particular norm is one we want to move away from.

pkg/vere/db/lmdb.c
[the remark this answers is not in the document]
mcevoypeter: @matthew-levan I agree the callback-based interface to `u3_lmdb_read()` requires a little more thought to use when compared to `u3_lmdb_read_one()`, but I still think `u3_lmdb_read()` will yield a cleaner solution. To that end, can you replace `u3_lmdb_read_one()` with `u3_lmdb_read()` and see how it turns out?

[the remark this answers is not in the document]
belisarius222: We switched to using the iterators @joemfb mentioned.  This has the advantage of not needing a callback, while also not requiring adding code to the lmdb wrapper for reading a single event, which is likely to only have this one call-site.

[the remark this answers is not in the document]
matthew-levan: `u3_lmdb_walk_*` functions are awesome.

pkg/noun/events.c
```diff
@@ -1038,6 +1038,13 @@ _ce_backup(void)
 
   close(nop_u.fid_i);
   close(sop_u.fid_i);
+  fprintf(stderr, "loom: image backup complete\r\n");
+}
+
+c3_o
+u3e_curr(void)
+{
+  return __(0 == _ce_patch_open());
```
joemfb, line 1047: This will leak memory in the case where a patch does exist -- needs to be coupled with `_ce_patch_free(pat_u)`

matthew-levan: Thanks, fixed.

pkg/vere/main.c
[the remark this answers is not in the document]
joemfb: It's important to use `u3_king_bail()` if the process is acting as a "king". If not, it'll just cause problems.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,165 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+
+  // check if there's a *current* snapshot
+  if ( old_u->dun_d != u3A->eve_d ) {
+    fprintf(stderr, "chop: error: snapshot is out of date, please "
+                    "start/shutdown your pier gracefully first\r\n");
+    fprintf(stderr, "chop: eve_d: %" PRIu64 ", dun_d: %" PRIu64 "\r\n", u3A->eve_d, old_u->dun_d);
+    exit(1);
+  }
+
+  u3e_backup(c3y);  //  backup current snapshot
```
joemfb, line 1798: We can only continue here if `u3e_backup()` succeeds. It's currently a `void` function, with no indication of error (other than printfs). It'll need to return a status flag (at least success/failure), and we'll have to fail here if the backup was unsuccessful.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,165 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+    fprintf(stderr, "chop: failed to read event\r\n");
+    exit(1);
+  }
+  u3_lmdb_walk_done(&itr_u);
+
+  // initialize a fresh lmdb environment in the "chop" subdir
+  c3_c cho_c[8193];
+  snprintf(cho_c, sizeof(cho_c), "%s/chop", log_c);
+  c3_mkdir(cho_c, 0700);
+  MDB_env* new_u = u3_lmdb_init(cho_c, siz_i);
```
joemfb, line 1846: We need to check for a failed initialization here (`!new_u`).

pkg/vere/main.c
```diff
@@ -1726,6 +1727,165 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+  c3_c dat_c[8193], bak_c[8193];
+  snprintf(dat_c, sizeof(dat_c), "%s/data.mdb", log_c);
+  // "data_<first>-<last>.mdb.bak"
+  snprintf(bak_c, sizeof(bak_c), "%s/data_%" PRIu64 "-%" PRIu64 ".mdb.bak", cho_c, fir_d, las_d);
+  c3_rename(dat_c, bak_c);
+
+  // rename new database file to be official
+  c3_c new_c[8193];
+  snprintf(new_c, sizeof(new_c), "%s/data.mdb", cho_c);
+  c3_rename(new_c, dat_c);
```
joemfb, line 1871: Both of these `rename()` calls need error handling/printfs.

joemfb (changes requested):
This code is clear and makes sense. I've noted a few spots where we'll need more/better error handling:

pkg/vere/main.c
matthew-levan: Thank you, I should've caught this. 

Related: any more thoughts on whether or not we're going to be happy overwriting the existing backup snapshot?

pkg/noun/events.c
```diff
@@ -1004,25 +1004,28 @@ _ce_backup(void)
...
   snprintf(ful_c, 8192, "%s/.urb/bhk", u3P.dir_c);
 
-  if ( c3_mkdir(ful_c, 0700) ) {
+  if ( (c3n == ovw_o) && c3_mkdir(ful_c, 0700) ) {
     if ( EEXIST != errno ) {
       fprintf(stderr, "loom: image backup: %s\r\n", strerror(errno));
+      fprintf(stderr, "loom: image backup failed\r\n");
+      return c3n;
     }
-    return;
```
joemfb, line 1011: This change is not right -- we need to return in the outer conditional. Also, no additional printf is needed.

pkg/noun/events.c
```diff
@@ -1004,25 +1004,28 @@ _ce_backup(void)
...
   }
 
   snprintf(ful_c, 8192, "%s/.urb/bhk/%s.bin", u3P.dir_c, nop_u.nam_c);
 
   if ( -1 == (nop_u.fid_i = c3_open(ful_c, mod_i, 0666)) ) {
     fprintf(stderr, "loom: c3_open %s: %s\r\n", ful_c, strerror(errno));
-    return;
+    fprintf(stderr, "loom: image backup failed\r\n");
```
joemfb, line 1019: This additional printf is not needed.

pkg/noun/events.c
```diff
@@ -1004,25 +1004,28 @@ _ce_backup(void)
...
+    return c3n;
   }
 
   snprintf(ful_c, 8192, "%s/.urb/bhk/%s.bin", u3P.dir_c, sop_u.nam_c);
 
   if ( -1 == (sop_u.fid_i = c3_open(ful_c, mod_i, 0666)) ) {
     fprintf(stderr, "loom: c3_open %s: %s\r\n", ful_c, strerror(errno));
-    return;
+    fprintf(stderr, "loom: image backup failed\r\n");
```
joemfb, line 1027: This additional printf is not needed.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,182 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+  u3_Host.dir_c = _main_pier_run(argv[0]);
+
+  while ( -1 != (ch_i=getopt_long(argc, argv, "", lop_u, &lid_i)) ) {
+    switch ( ch_i ) {
+      case c3__loom: {
+        c3_w lom_w;
+        c3_o res_o = _main_readw(optarg, u3a_bits + 3, &lom_w);
+        if ( (c3n == res_o) || (lom_w < 20) ) {
+          fprintf(stderr, "error: --loom must be >= 20 and <= %u\r\n", u3a_bits + 2);
+          u3_king_bail();
```
joemfb, line 1752: This (and all other such calls added in this PR) should be returned to `exit(1)`. This is a utility command, not a "king" process (ie, `u3_king_commence()` has not been called). Sorry for creating confusion.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,182 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+
+  // success
+  fprintf(stderr, "chop: event log truncation complete\r\n");
+  fprintf(stderr, "chop: event log backup written to %s\r\n", bak_c);
+  fprintf(stderr, "chop: WARNING: ENSURE YOU CAN RESTART YOUR SHIP "
+                  "BEFORE DELETING YOUR EVENT LOG BACKUP FILE!\r\n");
+  fprintf(stderr, "chop: if you can't restart your ship, restore your "
+                  "event log backup file with the following command:\r\n"); 
+  fprintf(stderr, "chop: mv %s %s\r\n", bak_c, dat_c);
+  fprintf(stderr, "chop: then restart your ship\r\n");
```
joemfb, line 1903: These printfs can be combined, and should not all be prefixed like this. Subsequent output lines should be indented for vertical alignment, and wrapped to stay within 80 columns.

pkg/vere/main.c
joemfb: You can't just compare against `errno` here, you have to check the return code from `rename()`.

pkg/vere/main.c
```diff
@@ -1726,6 +1727,182 @@ _cw_prep(c3_i argc, c3_c* argv[])
...
+  if ( c3n == u3_lmdb_walk_next(&itr_u, &len_i, buf_v) ) {
+    fprintf(stderr, "chop: failed to read event\r\n");
+    u3_king_bail();
+  }
+  u3_lmdb_walk_done(&itr_u);
+
+  // initialize a fresh lmdb environment in the "chop" subdir
+  c3_c cho_c[8193];
+  snprintf(cho_c, sizeof(cho_c), "%s/chop", log_c);
+  c3_mkdir(cho_c, 0700);
```
joemfb, line 1848: The return code of `mkdir()` needs to be checked for errors as well.

joemfb (changes requested):

pkg/vere/main.c
joemfb: We don't really have better options right now. The way to resolve this is the epoch system.

matthew-levan: Update: Instead of having to decide whether or not to overwrite the backup snapshot (in `.urb/bhk`) if it exists, I'm just going to make a backup of the backup snapshot if it exists. It'll go in the `.urb/log/chop/bhk` folder.

joemfb: That won't work, you absolutely need the latest (current!) snapshot.

pkg/vere/main.c
matthew-levan: Thanks, WIP...

pkg/vere/main.c
matthew-levan: Ok so we will go ahead with the original implementation: backup snapshot in (in `.urb/bhk`) will simply be overwritten. The real solution to this is the epoch system.

matthew-levan:
Here's an example output:
```
matt@mbp14 vere % ./urbit chop zod         
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/139.001.856
boot: installed 652 jets
loom: image backup complete
chop: event log truncation complete
      event log backup written to zod/.urb/log/chop/data_152-152.mdb.bak
      WARNING: ENSURE YOU CAN RESTART YOUR SHIP BEFORE DELETING YOUR EVENT LOG BACKUP FILE!
      if you can't, restore your log by running:
      `mv zod/.urb/log/chop/data_152-152.mdb.bak zod/.urb/log/data.mdb` then try again
```

belisarius222 (approved):

PR #171, opened 2023-01-26
Create pull_request_template.md

### LINKED ISSUE #170: create pull request template
When a PR is created, this should automatically populate as the PR description
comment. If we're trying to adhere to this form, it's probably good to be
reminded of it.

barter-simsum:
### Description

Resolves #170 

When a PR is created, this should automatically populate as the PR description
comment. If we're trying to adhere to this form, it's probably good to be
reminded of it.

Note, this doesn't take effect unless added to /master.


.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+Review https://github.com/urbit/vere/blob/develop/CONTRIBUTING.md
+
+-->
+
+### Description
+
+Resolves # <!-- All PRs must link an issue. Do it here. -->
+
+A thorough description of the changes made.
```
zalberico, line 11: We probably want this to be a comment too.

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Other sections
+
+...
+
+### Related
+
+References to any related issues, links, papers, etc.
```
zalberico, line 19: Same with this one (comment)

zalberico (changes requested):
We probably want the two lines from the docs to also be comments, otherwise seems fine to me. Thanks for thinking of this.

barter-simsum:
@Fang- , opposed?

Fang-:
Yes, somewhat opposed. I've voiced my reservations about gh templates in the past. Especially for PRs (as opposed to issues) I think it's fair to expect a higher standard of description without hand-holding the author into it. To me, pre-filled templates always feel oppressive rather than helpful.

Just a gentle nudge to "please review contributing.md" as is at the top here is probably the least invasive thing we could do. As long as that's there, not sure what purpose it serves to also pre-write the section headings. The "Other sections" heading feels doubly out of place, since it's not even in the contributing.md.

Also still not a fan of squash-merging, but perhaps that ship has sailed. Somewhat out of scope for discussion here anyway.

zalberico:
The main bit I care about is `Fixes: ####` so people remember to put the issue link in there.

barter-simsum:
@Fang-  I agree. I find pull request templates in general pretty annoying. Often pull request descriptions don't even really line up with the template given. Maybe just including a `Description` header -- does that even add anything? -- and I think the "Resolves#" string is fine to include and serves as a good reminder to link an issue. Will delete all comments etc.

How about that?

barter-simsum:
The template is now a single line "Resovled #" as a basic reminder to link an issue. Doesn't require any annoying deletion of text from contributor.

@Fang-

zalberico:
Cool - if we find people leaving the details blank we can ask them to add details (and if necessary introduce some templating if it's an issue).

belisarius222:
Should this PR be targeting master, as it is now, or develop?

mcevoypeter:
It should target `develop`. Good catch, @belisarius222.

mcevoypeter (approved):

mcevoypeter:
@Fang- if you feel good about this, we can merge it.

Fang-:
Sure, this is minimal enough. Looks better!

PR #175, opened 2023-01-27
term: stop setting scroll regions for printfs

### LINKED ISSUE #168: term: content pushed out of scroll region disappears in some terminals
A report from the field, thanks to ~nomryg-nilref:

> After running `-test` and the output exceeds the screen size I can no longer scroll back to earlier output. Only the current screen's output can be viewed, (...) in Gnome's terminal.

> Seems to have something to do with test output. I can say for sure when all tests pass I cannot scroll past one screen's worth of output in Gnome terminal. (...)
> Last time (an hour or 2 ago) I had failing tests I could scroll completely back through all the test output in Gnome terminal. Nothing was wrong with that terminal.

As of vere v1.17, we have changed the way printfs get rendered. Instead of overwriting the line at the bottom of the screen, we use [control sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html) to 1) set the scroll region to exclude the bottom line (the prompt), 2) scroll up by one line, 3) draw the printf in the empty space we have created. Step (1) ensures we never clobber the prompt, wrapping text will simply push other content up. (Afterwards, we reset the scroll region, back to the whole screen.)

On all terminals I have personally tested, (1) does not interfere with the backlog scrolling afforded natively by those terminals. Content simply gets pushed up into scrollback as normal. Perhaps, for Gnome's default terminal, this is not the case?

Unclear what can be done about this without changing the whole printf printing approach. The previous implementation stored the bottom line of the screen so that it could be overwritten and reconstructed.

Notably this problem will not exist once we separate out printfs from formal output entirely, by making vere start up in a non-interactive mode by default, and attaching to its terminal sessions using a separate process. See also logging reform.

Fang-:
We were leveraging scrolling regions (the [`r` CSI command](https://terminalguide.namepad.de/seq/csi_sr/)) and scroll commands (the [`S` CSI command](https://terminalguide.namepad.de/seq/csi_cs/), for "scroll up") to create space in which printfs could exist. (This, besides the spinner, is where we depend on the assumption that there is a prompt on the bottom-most line of the screen.)

In the context of tmux, a scroll region scrolling up would simply push its contents up into scrollback/history. However, in the context of other terminals, 'S' is interpreted as "delete the top line(s), then shift all remaining content up". This obliterates the content of those top lines from existence irretrievably.

An alternative implementation prints a newline character at the bottom of the configured scroll region. This preserves the lines in scrollback in xterm, but this doesn't seem to be consistent across terminal emulators. (This seems like a desirable solution for webterm though, because we know we're always running xterm(.js) there.)

The solution we land on here, which behaves consistently in different environments, is reminiscent of the pre-terminal update implementation. We do not set a scroll region, instead we clear the bottom-most line, write the printf, bump it up by printing a newline character, and then restore the contents of the bottom-most line. Now that no scroll region is explicitly set, all tested terminals properly accept the bumped-up lines into their scrollback buffers.

We can do this only because we already track the contents that get drawn to the bottom-most line of the screen, for purposes of restoring it after we've clobbered it by drawing the spinner on top. The change herein is, in that sense, a deepening of the coupling to drum's semantics. Fortunately, both of these problems can be resolved by the much-desired, long-awaited project of Logging Reform. We must, as always, announce: Soon™.

Fixes #168.

### DIFF
--- a/pkg/vere/io/term.c
+++ b/pkg/vere/io/term.c
@@ -484,7 +484,8 @@ _term_it_save_stub(u3_utty* uty_u, u3_noun tub)
   u3_utat* tat_u = &uty_u->tat_u;
   u3_noun  lin   = tat_u->mir.lin;
 
-  //  keep track of changes to bottom-most line, to aid spinner drawing logic.
+  //  keep track of changes to bottom-most line, to aid screen in restoration
+  //  after printfs or spinners.
   //  -t mode doesn't need this logic, because it doesn't render the spinner.
   //
   if ( ( tat_u->siz.row_l - 1 == tat_u->mir.rus_w ) &&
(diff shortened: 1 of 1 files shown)


istathar (approved):
"or just not use scroll regions" :)

zalberico:
Nice - if we get this merged by Monday we can get it in the branch for release on Tuesday (with a ship target then of the following Tuesday: Feb 7th)

mcevoypeter (approved):

zalberico:
@Fang- go ahead and merge to get this in the branch for release tomorrow :)

PR #183, opened 2023-01-31
Merge develop into release

jalehman:
Per @mcevoypeter , this wasn't taken care of after the last release, so we're doing it now.

### DIFF
--- a/.github/workflows/feature.yml
+++ b/.github/workflows/feature.yml
@@ -1,7 +1,6 @@
 name: Feature pull request
 
 on:
-  release: null
   pull_request:
     paths:
       - '.bazelrc'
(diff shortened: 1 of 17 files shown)


zalberico:
I'm confused - soon is on 1.18 and has not been released? We want to do this after we ship, but not before. I'll join the sync in a minute

jalehman:
We had thought that this step hadn't been done after the last release but have since realized that that is not the case. Disregard.

mcevoypeter:
This stemmed from a misunderstanding on my part. For the sake of clarity, the 1.18 release candidate was cut on Friday, Jan 20 when @philipcmonk merged `develop` into `release`. Everything that is on `develop` now (but not on `release`) is slated for 1.19, which will be cut later today (by merging `develop` into `release`) after 1.18 is released to the live network.

PR #199, opened 2023-02-06
http: check requests for auth using eyre's tokens

Fang-:
Cherry-picked over from urbit/urbit#5973. Depends on the eyre-side change there, so this will be part of a kelvin update.

Accept the newly added `%session` effect in the http io driver. Store the tokens received in it, and refer back to them when checking requests for authentication.

Adds logic for checking requests for authentication based on the presence of a valid urbauth cookie for the host ship, and updates the slogstream endpoint to use that instead of the scry-based authentication check.

### DIFF
--- a/pkg/vere/io/http.c
+++ b/pkg/vere/io/http.c
@@ -83,6 +83,8 @@ typedef struct _u3_h2o_serv {
 */
   typedef struct _u3_hfig {
     u3_form*         for_u;             //  config from %eyre
+    c3_c*            key_c;             //  auth token key
+    u3_noun          ses;               //  valid session tokens
     struct _u3_hreq* seq_u;             //  open slog requests
     uv_timer_t*      sit_u;             //  slog stream heartbeat
   } u3_hfig;
@@ -300,6 +302,71 @@ _http_heds_from_noun(u3_noun hed)
 
   u3z(deh);
   return hed_u;
+}
+
+/* _http_req_is_auth(): returns c3y if rec_u contains a valid auth cookie
+*/
+static c3_o
+_http_req_is_auth(u3_hfig* fig_u, h2o_req_t* rec_u)
+{
+  //  try to find a cookie header
+  //
+  h2o_iovec_t coo_u = {NULL, 0};
+  {
+    //TODO  http2 allows the client to put multiple 'cookie' headers,
+    //      runtime should support that once eyre does too.
+    ssize_t hin_i = h2o_find_header_by_str(&rec_u->headers, "cookie", 6, -1);
+    if ( hin_i != -1 ) {
+      coo_u = rec_u->headers.entries[hin_i].value;
+    }
+  }
+
+  //  if there is no cookie header, it can't possibly be authenticated
+  //
+  if ( NULL == coo_u.base ) {
+    return c3n;
+  }
+  //  if there is a cookie, see if it contains a valid auth token
+  //
+  else {
+    c3_c* key_c = fig_u->key_c;
+    c3_c  val_c[128];
+    c3_y  val_y = 0;
+    c3_y    i_y = 0;
+    c3_y    j_y = 0;
+
+    //  step through the cookie string
+    //
+    while (i_y < coo_u.len) {
+      //  if we found our key, read the value
+      //
+      if (key_c[j_y] == '\0' && coo_u.base[i_y] == '=') {
+        i_y++;
+        while (i_y < coo_u.len && coo_u.base[i_y] != ';') {
+          val_c[val_y] = coo_u.base[i_y];
+          val_y++;
+          i_y++;
+        }
+        break;
+      }
+      //  keep reading the key as long as it matches
+      //
+      else if (coo_u.base[i_y] == key_c[j_y]) {
+        j_y++;
+      }
+      else {
+        j_y = 0;
+      }
+      i_y++;
+    }
+
+    u3_noun aut = u3kdi_has(
+      u3k(fig_u->ses),
+      u3i_bytes(val_y, (c3_y*)val_c));
+    c3_assert(c3y == aut || c3n == aut);
+
+    return aut;
+  }
 }
 
 /* _http_req_find(): find http request in connection by sequence.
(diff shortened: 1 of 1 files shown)


pkg/vere/io/http.c
```diff
@@ -302,6 +304,71 @@ _http_heds_from_noun(u3_noun hed)
...
+    return c3n;
+  }
+  //  if there is a cookie, see if it contains a valid auth token
+  //
+  else {
+    c3_c* key_c = fig_u->key_c;
+    c3_c  val_c[128];
+    c3_y  val_y = 0;
+    c3_y    i_y = 0;
+    c3_y    j_y = 0;
```
mcevoypeter, line 336: Is the max length of `coo_u.base` less than or equal to `255`? The reason I ask is because it's unclear to me whether this code avoids overflowing its index variables i.e. `i_y`, `j_y`. On a similar note, is the max auth token length `128`?

pkg/vere/io/http.c
```diff
@@ -302,6 +304,71 @@ _http_heds_from_noun(u3_noun hed)
...
+      }
+      else {
+        j_y = 0;
+      }
+      i_y++;
+    }
+
+    u3_noun aut = u3kdi_has(
+      u3k(fig_u->ses),
+      u3i_bytes(val_y, (c3_y*)val_c));
```
mcevoypeter, line 365: Nit: this is a little hard to read. I think one of the following would be easier to read:
```c
u3_noun aut = u3kdi_has(u3k(fig_u->ses),
                        u3i_bytes(val_y, (c3_y*)val_c));
// or
u3_noun aut = u3kdi_has(
    u3k(fig_u->ses),
    u3i_bytes(val_y, (c3_y*)val_c)
);

File: pkg/vere/io/http.c [hunk 3]
```diff
@@ -1892,6 +1906,16 @@ _http_form_free(u3_httd* htd_u)
...
+/* _http_auth_free(): free stored auth token state
+*/
+static void
+_http_auth_free(u3_httd* htd_u)
+{
+  u3z(htd_u->fig_u.ses);
+  htd_u->fig_u.ses = u3_nul;
+  c3_free(htd_u->fig_u.key_c);
+}
+
```
mcevoypeter, line 1918: This is very similar to `u3_http_ef_auth()`, which has me thinking it may be cleaner to combine them into a single function, `_http_set_auth_tokens()` (or similar):
```c
void
_http_set_auth_tokens(u3_httd* htd_u, u3_noun fig)
{
  u3z(htd_u->fig_u.ses);
  htd_u->fig_u.ses = fig;
}
```

The call site for `u3_http_ef_auth()` below would then turn into:
```c
_http_set_auth_tokens(htd_u, u3k(dat));
```
and the call site for `_http_auth_free()` would turn into:
```c
_http_set_auth_tokens(htd_u, u3_nul);
c3_free(htd_u->fig_u.key_c);
```

pkg/vere/io/http.c
Fang-: Yes, auth tokens are 128 bits of entropy, encoded in `@uv` representation. In practice this'll be 33 characters or thereabouts, so 128 characters is plenty.

Good catch on the overflows though. I bumped them up to `c3_w` which is much more reasonable here. What's the appropriate way to say "if `coo_u.len` is larger than `c3_w`'s max value, consider it unauthed"?

pkg/vere/io/http.c
Fang-: Not a fan of the proposed change here, but kind of difficult to put my finger on why. I like the `c3_free` living explicitly alongside the rest of the bookkeeping for auth state, rather than implicitly as just the next step in `_http_io_exit`. So if anything, I would just have `_http_auth_free` call `u3_http_ef_auth`, but that feels a little fake.

belisarius222:
@mcevoypeter I think this is ready for another review.  I just read through this -- it looks like all review comments have been addressed, unless you consider https://github.com/urbit/vere/pull/199#discussion_r1097836917 a blocker for merging this.

pkg/vere/io/http.c
mcevoypeter: I defer to your judgement here. I agree about `_http_auth_free` calling `u3_http_ef_auth` feeling fake. Sounds like we should leave it as it is.

pkg/vere/io/http.c
mcevoypeter: > Yes, auth tokens are 128 bits of entropy, encoded in @uv representation. In practice this'll be 33 characters or thereabouts, so 128 characters is plenty.

Great. Thanks for clarifying.

> What's the appropriate way to say "if coo_u.len is larger than c3_w's max value, consider it unauthed"?

Do you think it makes sense to bump the types up to `size_t`? That way, they can't possibly overflow. Another approach would be to check `coo_u.len` against the hard-coded max of `c3_w` and include a comment indicating the hard-coded max must be updated if the types of `i_w` and `j_w` change.

pkg/vere/io/http.c
```diff
@@ -302,6 +304,69 @@ _http_heds_from_noun(u3_noun hed)
...
+    c3_w    j_w = 0;
+
+    //  step through the cookie string
+    //
+    while (i_w < coo_u.len) {
+      //  if we found our key, read the value
+      //
+      if (key_c[j_w] == '\0' && coo_u.base[i_w] == '=') {
+        i_w++;
+        while (i_w < coo_u.len && coo_u.base[i_w] != ';' && val_y < 128) {
```
mcevoypeter, line 345: Let's use `sizeof(val_c)` here instead of the hard-coded `128`.

mcevoypeter (changes requested):
Last round of comments. Should be good to go after this.

mcevoypeter (approved):

Fang-:
Wait, I'm dumb, maybe I should not have merged this. 

> Depends on the eyre-side change [in urbit/urbit#5973], so this will be part of a kelvin update.

Those changes _have_ been merged, but the code here blocks even authenticated requests to the slogstream endpoint, if the ship we're running doesn't have those eyre changes yet.

"It's just the slogstream", but also we'll want to gate this new code behind a kelvin bump regardless, probably.

So maybe this should just be reverted for now?

PR #214, opened 2023-02-09
disambiguate header includes

### LINKED ISSUE #213: disambiguate header includes
1) Add `pkg` to all package include paths
2) For local dependencies, header includes should use an unambiguous path, e.g.

```
#include "noun/noun.h"
#include "vere/ivory.h"
#include "ur/ur.h"
...
```

Obviously this avoids name collisions, but I think it also makes the dependency tree of any C file more legible.

barter-simsum:
[x] Add `pkg` to all include paths
[ ] Disambiguate header includes

Resolves #213

### DIFF
--- a/pkg/c3/BUILD.bazel
+++ b/pkg/c3/BUILD.bazel
@@ -30,6 +30,9 @@ cc_library(
     includes = ["."],
     linkstatic = True,
     visibility = ["//pkg:__subpackages__"],
+    copts = [
+        "-Ipkg",
+    ],
 )
 
 #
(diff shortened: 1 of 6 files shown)


matthew-levan:
If I remember correctly, @mcevoypeter shared a better way of doing this (instead of using `-Ipkg` everywhere), but I can't find it now. @barter-simsum do you recall?

mcevoypeter:
> If I remember correctly, @mcevoypeter shared a better way of doing this (instead of using -Ipkg everywhere), but I can't find it now. 

I think this is what you're referring to: https://github.com/urbit/vere/issues/213#issuecomment-1425968673.

barter-simsum:
no longer care about this. closing

PR #229, opened 2023-02-14
CI: add `-v` and `*-musl` to the `chown` path

### LINKED ISSUE #227: `chown /usr/local` takes too long in our CI workflow
See: https://github.com/urbit/vere/actions/runs/4175998956/jobs/7231701539

matthew-levan:
Resolves #227.

### DIFF
--- a/.github/workflows/shared.yml
+++ b/.github/workflows/shared.yml
@@ -66,7 +66,7 @@ jobs:
           # this on linux-aarch64 because we run on a self-hosted GitHub Actions
           # runner for which `/usr/local` has already been `chown`d to the user
           # running the GitHub Actions runner.
-          sudo chown $(whoami) /usr/local && (chown -R $USER /usr/local*-musl || true)
+          sudo chown $(whoami) /usr/local && (chown -R $USER /usr/local/*-musl || true)
 
       - name: Set up build cache
         uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


belisarius222:
```
chown: cannot access '/usr/local/*-musl': No such file or directory
```
@matthew-levan said this will happen whenever there's a cache miss, since the directory won't be present.  I'm starting to wonder whether the ideal way to fix this might be to figure out how to avoid using sudo when building musl libc.  That way we should't have to chown anything.

Alternatively, we could investigate in more detail why it is that `sudo chown /usr/local` is taking ten minutes -- maybe there's something we could do to make sure that folder doesn't have so much stuff in it.

mcevoypeter:
If I remember correctly, the problem is GitHub Actions can't populate `/usr/local/*-musl` from the cache unless `/usr/local` is owned by `runner`, not `root`. The musl-libc toolchain needs to be installed in a configuration-independent directory so that Bazel can specify absolute paths to it that work on all platforms and user configurations. All directories that meet that criteria that I can think of are owned by `root`, so avoiding `sudo` may be unavoidable. A work-around might be to `chown $USER /usr/local && (chown -R $USER /usr/local*-musl || true)`.

matthew-levan:
@mcevoypeter's workaround works great, reducing the time for the `chown /usr/local/` step from ~10 minutes to 0 (!) seconds[^1]. The command used is `sudo chown $(whoami) /usr/local && (chown -R $USER /usr/local/*-musl || true)`.

[^1]: https://github.com/urbit/vere/actions/runs/4185211831

belisarius222 (approved):

PR #236, opened 2023-02-16
change copt usage in build configs to per_file_copt

### LINKED ISSUE #224: Decrease build time when switching copts
As has been explained by @mcevoypeter, switching a bazel build's command line `copt` switch rebuilds all dependencies which is a pain when you're testing vere's behavior with/without a given copt, e.g. `--copt="-DU3_MEMORY_DEBUG"`

I think an option for this would be defining aliases that do a per-file copt instead of a global copt. See https://bazel.build/docs/user-manual#per-file-copt

The per file copt can be used to scope the copt. Doing this manually would be tedious though. Can we define some sort of alias? We could always wrap the bazel commands in a Makefile.

Including @joemfb

@matthew-levan for his thoughts as well

barter-simsum:
Resolves #224

When rotating between various combinations of build configs defined in /.bazelrc
time to rebuild is significantly reduced by scoping their effects to only the
vere src via `--per_file_copt`.

whatever permutations of the following args when building vere:

```
bazel build ARGS :urbit

ARGS:
--config=dbg                    # -O0 opt and -g3 dbg symbols
--config=mem_dbg                # eq to -DU3_MEMORY_DEBUG
--config=cpu_dbg                # eq to -DU3_CPU_DEBUG
```

The `--config=[mem|cpu]_dbg` is a necessary evil indirection for faster builds
when iteratively switching between mem and or cpu debugging since passing the
macro define flag to bazel like: `bazel build --copt="-DU3_CPU_DEBUG" :urbit`
sends that macro define down to all dependencies -- thus rebuilding the entire
project rather than just the vere source.

Still feels a bit fragile, but this is the smallest easiest fix.


.bazelrc
```diff
@@ -23,24 +23,31 @@ build --flag_alias=gcc_version=//:gcc_version
...
+build:mem_dbg --per_file_copt='//:.*@-DU3_MEMORY_DEBUG'
+build:cpu_dbg --per_file_copt='//:.*@-DU3_CPU_DEBUG'
+
 # Enable maximum debug info and disable optimizations for debug config. It's
 # important that these lines come after setting the default debug and
 # optimization level flags above.
-build:dbg --copt='-O0'
-build:dbg --copt='-g3'
+build:dbg --copt="-O0"
+# build:dbg --per_file_copt='//:.*@-O0' # ;;: this breaks the build of softfloat for some reason I haven't investigated yet.
```
mcevoypeter, line 49: If you remove line 48 and uncomment line 49, `:urbit` builds without issue. Line 48 passes `-O0` to all targets, not just the ones in `//pkg/...`, and I suspect softfloat has an issue with `-O0`, it being an optimized arithmetic library.

.bazelrc
```diff
@@ -23,24 +23,31 @@ build --flag_alias=gcc_version=//:gcc_version
 # Don't include source level debug info on macOS. See
 # https://github.com/urbit/urbit/issues/5561 and
 # https://github.com/urbit/vere/issues/131.
-build:linux --copt='-g'
+build:linux --per_file_copt='//:.*@-g'
```
mcevoypeter, line 26: Are we sure `//:.*` is the correct regex? I'm concerned it'll only match files in the top level of the workspace. `//.*\.[ch]` seems like it may be what we want.

matthew-levan:
@barter-simsum what's the status on this?

.bazelrc
barter-simsum: Did you pass the `--config=dbg` argument when building? Uncommenting line 49 is what breaks debug builds for me.

```
ERROR: /Users/d/src/urbit/vere/pkg/vere/BUILD.bazel:141:10: Linking pkg/vere/urbit failed: (Exit 1): clang failed: error executing command (from target //pkg/vere:urbit) /usr/bin/clang -o bazel-out/darwin_arm64-fastbuild/bin/pkg/vere/urbit bazel-out/darwin_arm64-fastbuild/bin/pkg/vere/_objs/urbit/main.o ... (remaining 36 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Undefined symbols for architecture arm64:
  "_softfloat_countLeadingZeros16", referenced from:
      _softfloat_normSubnormalF16Sig in libsoftfloat_aarch64.a(s_normSubnormalF16Sig.o)
      _softfloat_subMagsF16 in libsoftfloat_aarch64.a(s_subMagsF16.o)
  "_softfloat_countLeadingZeros32", referenced from:
      _softfloat_mulAddF16 in libsoftfloat_aarch64.a(s_mulAddF16.o)
      _softfloat_normSubnormalF32Sig in libsoftfloat_aarch64.a(s_normSubnormalF32Sig.o)
      _softfloat_shiftNormSigF128M in libsoftfloat_aarch64.a(s_shiftNormSigF128M.o)
      _softfloat_subMagsF16 in libsoftfloat_aarch64.a(s_subMagsF16.o)
      _softfloat_subMagsF32 in libsoftfloat_aarch64.a(s_subMagsF32.o)
      _softfloat_normRoundPackMToF128M in libsoftfloat_aarch64.a(s_normRoundPackMToF128M.o)
      _softfloat_normRoundPackToF32 in libsoftfloat_aarch64.a(s_normRoundPackToF32.o)
      ...
  "_softfloat_countLeadingZeros64", referenced from:
      _softfloat_mulAddF32 in libsoftfloat_aarch64.a(s_mulAddF32.o)
      _softfloat_mulAddF64 in libsoftfloat_aarch64.a(s_mulAddF64.o)
      _softfloat_normSubnormalF64Sig in libsoftfloat_aarch64.a(s_normSubnormalF64Sig.o)
      _softfloat_subMagsF64 in libsoftfloat_aarch64.a(s_subMagsF64.o)
      _softfloat_normRoundPackToF64 in libsoftfloat_aarch64.a(s_normRoundPackToF64.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Target //pkg/vere:urbit failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 58.815s, Critical Path: 54.50s
INFO: 654 processes: 9 internal, 645 darwin-sandbox.
FAILED: Build did NOT complete successfully

```

.bazelrc
barter-simsum: `//:.*` seems fine. The build has approximately the same elapsed time, so I don't think we're building anything unnecessarily with the simpler regex.

mcevoypeter: My concern isn't that we're rebuilding anything unnecessarily, it's the opposite: I'm concerned we're *not* rebuilding everything we should. The [`--per_file_copt` documentation](https://bazel.build/docs/user-manual#per-file-copt) doesn't make it particularly clear whether `//:.*` will match files in subdirectories of the workspace root or not. I'd think we can answer this definitively using `bazel query` somehow.

barter-simsum: fyi, tested `//.*\.[ch]` and it doesn't fix the issue.

.bazelrc
barter-simsum: @mcevoypeter response to above?

mcevoypeter: I see what you're saying. I just ran the following:
```
$ bazel build --config=dbg --subcommands --per_file_copt='//:.*@-O0'  :urbit
/usr/local/x86_64-linux-musl/bin/x86_64-linux-musl-gcc -MD -MF bazel-out/k8-fastbuild/bin/external/softfloat/_objs/softfloat_x86_64/s_propagateNaNF128UI.d <snip> -O2 -g -O3 -g3 -O0 <snip>
```

```console
$ bazel build --config=dbg --subcommands --copt='-O0' :urbit
/usr/local/x86_64-linux-musl/bin/x86_64-linux-musl-gcc -MD -MF bazel-out/k8-fastbuild/bin/external/softfloat/_objs/softfloat_x86_64/extF80M_to_i32_r_minMag.d <snip> -O2 -g -O3 -g3 <snip>
```
Counterintuitively, `--copt='-O0'` doesn't append `-O0` to softfloat's compilation commands, but `--per_file_copt='//:.*@-O0'` does.

mcevoypeter: The only explanation I can think of is that `--copt` doesn't override `copts` attribute of `cc_library`/`cc_binary` targets when it recognizes conflicting options, but `--per_file_copt` does. The documentation doesn't make this explicit, but this is what I'm going off of:

[`--copt` documentation](https://bazel.build/docs/user-manual#copt):
> Also note that copts values listed in specific cc_library or cc_binary build rules will be placed on the compiler command line after these options.

[`--per_file_copt` documentation](https://bazel.build/docs/user-manual#per-file-copt):
> When present, any C++ file with a label or an execution path matching one of the inclusion regex expressions and not matching any of the exclusion expressions will be built with the given options.

barter-simsum: Looked closer at the compile commands you sent. Yeah I see what you're saying. Any clue why -O3 is getting appended there in both cases? That shouldn't be happening at all. The optimization level we define in softfloat.BUILD is -O2. FYI when you give multiple values of the same options like this, gcc takes the last given, so that binary got compiled at -O3

@mcevoypeter

barter-simsum: If that's the behavior of --per_file_copt, maybe our regex can be more restrictive? I.e. whitelist the directories under /pkg??

mcevoypeter: > Any clue why -O3 is getting appended there in both cases?

We specify `-O3` as the default a few lines above in `.bazelrc` with `build --per_file_copt='//:.*@-O3'`. So `--per_file_copt` seems to be more aggressive than `--copt`.

> whitelist the directories under /pkg??

Sounds promising. So the regex would be `//pkg:.*` instead of `//:.*`.

INSTALL.md
```diff
@@ -68,18 +68,18 @@ Note that you cannot change the optimization level for third party
...
+bazel build --config=cpu_dbg --config=mem_dbg :urbit
 ```
 Note that defining these two debug symbols will produce ships that are
 incompatible with binaries without these two debug symbols defined.
 
 If you need to specify arbitrary C compiler or linker options, use
 [`--copt`][copt] or [`--linkopt`][linkopt], respectively:
 ```console
-bazel build --copt='-O0' :urbit
+bazel build --copt='-Wall' :urbit
```
barter-simsum, line 82: fyi, changed this line because using `-O0` may be confusing

.bazelrc
barter-simsum: Just verifying the more restrictive pattern works. Pushed those changes. This looks ready to merge to me. Thanks @mcevoypeter for the help.

.bazelrc
```diff
@@ -23,24 +23,31 @@ build --flag_alias=gcc_version=//:gcc_version
...
+# including/excluding a symbol define
+build:mem_dbg --per_file_copt='pkg/.*@-DU3_MEMORY_DEBUG'
+build:cpu_dbg --per_file_copt='pkg/.*@-DU3_CPU_DEBUG'
+
 # Enable maximum debug info and disable optimizations for debug config. It's
 # important that these lines come after setting the default debug and
 # optimization level flags above.
-build:dbg --copt='-O0'
-build:dbg --copt='-g3'
+# build:dbg --copt="-O0"
```
mcevoypeter, line 48: Can you remove this line?

.bazelrc
```diff
@@ -23,24 +23,31 @@ build --flag_alias=gcc_version=//:gcc_version
...
+build:dbg --per_file_copt='pkg/.*@-O0' # ;;: this breaks the build of softfloat for some reason I haven't investigated yet.
```
mcevoypeter, line 49: Can you remove the trailing comment now that the softfloat build works?

mcevoypeter (changes requested):
Looks good to me. I'll hold off approving until after the stale lines I commented on are removed since my review would just be dismissed anyway.

.bazelrc
barter-simsum: sorry about that had these changes locally, but didn't upstream them

barter-simsum:
@mcevoypeter addressed the dangling comments. Additionally, modified install.md to mention using `per_file_copt` since it could be useful for devs.

mcevoypeter (dismissed):

mcevoypeter (approved):

PR #255, opened 2023-02-23
eval: add --cue option

### LINKED ISSUE #197: Khan/Conn: Working client for sending commands and returning output to an urbit from outside of it.
**Khan**: The existing urbit vane part of the code
**Conn**: The existing runtime part of the code
**Client**: The external tool used to interact with **conn** in order to execute commands from outside of urbit.

Khan and Conn exist to allow a user to execute commands on an urbit from the outside and get the results. Currently though, there is no client for interacting with this service.

@ashelkovnykov is working on a client that will make it possible execute commands and get the relevant output from the already existing Khan/Conn service in urbit. The primary consumer of this client will be Tlon Hosting and its design is specifically targeted for our needs.

This issue will be the source of truth about the state of things and the place to discuss problems or questions. When a PR exists, this is the issue it will reference.

ashelkovnykov:
Partially Resolves #197

Adds the `--cue` option to the `eval` utility. This allows jammed nouns to be passed over `stdin` and pretty-printed to `stdout`.

Also includes a sub-option which will expect the incoming jammed noun to be Khan output. If the thread resulted in a `goof`, it's pretty-printed to `stderr` as a stack trace and the output is simply the `conn.c` header info (unique ID, `%avow` mark signalling Khan output, `%.n` loobean signalling thread failure).

Also includes the "todo"s from the comments for #155.

**Testing:**
- Built Vere in debug mode; no new warnings introduced
- Manually ran the following test cases in the terminal:
```
echo '42' | ~/Urbit/urbit eval -jn | ~/Urbit/urbit eval -cn
echo '[[1 2] [3 4]]' | ~/Urbit/urbit eval -jn | ~/Urbit/urbit eval -cn
echo "(cat 3 'abc' 'def')" | ~/Urbit/urbit eval -jn | ~/Urbit/urbit eval -cn
```
- I've done more strenuous testing, but it's hard to explicitly list here since it relies on a thin client which will be committed as a separate PR after this one is merged

Not strictly necessary for this PR, but I would like to add explicit testing for both `--jam` and `--cue`. @mcevoypeter @matthew-levan If either of you can recommend how to modify Bazel to do so, it would be appreciated.

### DIFF
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -829,6 +829,15 @@ _cw_serf_fail(void* ptr_v, ssize_t err_i, const c3_c* err_c)
   exit(1);
 }
 
+/* _cw_king_fail(): local failure stub.
+*/
+static void
+_cw_king_fail(void* ptr_v, ssize_t err_i, const c3_c* err_c)
+{
+  fprintf(stderr, "king: eval error: %s\r\n", err_c);
+  exit(1);
+}
+
 /* _cw_serf_send(): send plea back to daemon.
 */
 static void
@@ -925,7 +934,7 @@ static void
 _cw_serf_stdio(c3_i* inn_i, c3_i* out_i)
 {
   //  the serf is spawned with [FD 0] = events and [FD 1] = effects
-  //  we dup [FD 0 & 1] so we don't accidently use them for something else
+  //  we dup [FD 0 & 1] so we don't accidentally use them for something else
   //  we replace [FD 0] (stdin) with a fd pointing to /dev/null
   //  we replace [FD 1] (stdout) with a dup of [FD 2] (stderr)
   //
@@ -960,7 +969,7 @@ static void
 _cw_init_io(uv_loop_t* lup_u)
 {
   //  mars is spawned with [FD 0] = events and [FD 1] = effects
-  //  we dup [FD 0 & 1] so we don't accidently use them for something else
+  //  we dup [FD 0 & 1] so we don't accidentally use them for something else
   //  we replace [FD 0] (stdin) with a fd pointing to /dev/null
   //  we replace [FD 1] (stdout) with a dup of [FD 2] (stderr)
   //
@@ -1186,19 +1195,21 @@ _cw_eval_get_input(FILE* fil_u, size_t siz_i)
 static void
 _cw_eval(c3_i argc, c3_c* argv[])
 {
-  c3_i ch_i, lid_i;
-  c3_w arg_w;
-  c3_o jam_o = c3n;
-  c3_o kan_o = c3n;
+  u3_mojo std_u;
+  c3_i    ch_i, lid_i;
+  c3_w    arg_w;
+  c3_o    jam_o = c3n;
+  c3_o    new_o = c3n;
 
   static struct option lop_u[] = {
-    { "loom", required_argument, NULL, c3__loom },
-    { "jam", no_argument, NULL, 'j' },
-    { "jamkhan", no_argument, NULL, 'k' },
+    { "loom", required_argument,  NULL, c3__loom },
+    { "jam",  no_argument,        NULL, 'j' },
+    { "newt", no_argument,        NULL, 'n' },
+    //
     { NULL, 0, NULL, 0 }
   };
 
-  while ( -1 != (ch_i=getopt_long(argc, argv, "", lop_u, &lid_i)) ) {
+  while ( -1 != (ch_i=getopt_long(argc, argv, "jn", lop_u, &lid_i)) ) {
     switch ( ch_i ) {
       case c3__loom: {
         c3_w lom_w;
@@ -1212,12 +1223,10 @@ _cw_eval(c3_i argc, c3_c* argv[])
 
       case 'j': {
         jam_o = c3y;
-        kan_o = c3n;
       } break;
 
-      case 'k': {
-        jam_o = c3y;
-        kan_o = c3y;
+      case 'n': {
+        new_o = c3y;
       } break;
 
       case '?': {
@@ -1226,6 +1235,12 @@ _cw_eval(c3_i argc, c3_c* argv[])
       } break;
     }
   }
+
+  //  newt meaningless without jam
+  //
+  if ( ( c3y == new_o ) && ( c3n == jam_o) ) {
+    fprintf(stderr, "--newt meaningless w/o --jam; ignoring\r\n");
+  }
   //  argv[optind] is always "eval"
   //
   if ( optind + 1 != argc ) {
@@ -1233,6 +1248,19 @@ _cw_eval(c3_i argc, c3_c* argv[])
     exit(1);
   }
   c3_c* evl_c = _cw_eval_get_input(stdin, 10);
+
+  //  configure stdout as u3_mojo
+  //
+  {
+    c3_i err_i;
+    err_i = uv_pipe_init(uv_default_loop(), &std_u.pyp_u, 0);
+    c3_assert(!err_i);
+    uv_pipe_open(&std_u.pyp_u, 1);
+
+    std_u.ptr_v = NULL;
+    std_u.bal_f = _cw_king_fail;
+  }
+
   //  initialize the Loom and load the Ivory Pill
   //
   {
@@ -1254,6 +1282,7 @@ _cw_eval(c3_i argc, c3_c* argv[])
       exit(1);
     }
   }
+
   fprintf(stderr, "eval:\n");
   if ( c3n == jam_o ) {
     //  +wish for an eval gate (virtualized twice for pretty-printing)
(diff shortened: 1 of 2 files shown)


pkg/vere/main.c
```diff
@@ -1170,43 +1179,80 @@ _cw_eval_get_input(FILE* fil_u, size_t siz_i)
...
 
+/* _cw_eval_get_newt(): read a newt-encoded jammed noun from file and return a
+**                      malloc'd byte buffer
+*/
+c3_y*
+_cw_eval_get_newt(FILE* fil_u, c3_d* len_d)
+{
+  //  TODO: can't reuse u3_newt_decode; coupled to reading from pipe
+  c3_d  i;
+  c3_y* byt_y = c3_realloc(NULL, 5); //  size is start size
```
mcevoypeter, line 1190: Why not use `c3_malloc()` to allocate for the first time instead of using the edge case behavior of `c3_realloc()`?

mcevoypeter:
> Not strictly necessary for this PR, but I would like to add explicit testing for both --jam and --cue. @mcevoypeter @matthew-levan If either of you can recommend how to modify Bazel to do so, it would be appreciated.

For testing the behavior of the command line application itself, I think the best way to do this is by writing a shell script that invokes `--jam` and `--cue` and compares the generated output to the expected output. We can then wrap that shell script in a [`sh_test` target](https://bazel.build/reference/be/shell#sh_test) in Bazel.

pkg/vere/main.c
```diff
@@ -65,17 +65,31 @@ _main_self_path(void)
...
   c3_w  par_w = strtoul(str_c, &end_c, 0);
 
   if ( *str_c != '\0' && *end_c == '\0' && par_w < max_w ) {
     *out_w = par_w;
-    return c3y;
+    return 0;
   }
-  else return c3n;
+  else return 1;
+}
```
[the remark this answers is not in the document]
ashelkovnykov: I'll first mention that I included this commit in the PR because the use of `c3_o` for non-Urbit control flow bothered me so I wanted to force a discussion on the matter. However, the commit has nothing to do with `--cue` and I'm prepared to pull it out into a separate PR if the discussion runs too long.

> you've created a situation in which the if branch of the following code is taken when _main_readw() fails and the else branch is taken when _main_read() succeeds, which is directly opposite the convention in C

This is incorrect. It's even in the C standard for the main function that a non-zero return value indicates that an error has occurred [7.20.4.3.5](https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf). It's also a POSIX standard that `0` means success (no link; it costs money and is tedious to access).

If you look over the changes in the commit, you will see that all of the cases in/of `_main_readw()` and `_main_getopt()` were already using `1` as an error state: `c3n` (i.e. `1`) meant failure; `c3y` (i.e. `0`) meant success. Depending on how you look at it, they were actually *breaking* the C/POSIX convention, since the "false" loobean was being used to signal error.

Regardless of the above, using `c3_o` as part of the control flow for Vere is a bad idea that should be done away with. Loobeans should only be used in the cases where Vere is directly generating / operating on Urbit nouns. In the cases where loobean values from Urbit determine Vere control flow, we already have macros for converting loobeans to booleans.

pkg/vere/main.c
```diff
@@ -1170,43 +1179,80 @@ _cw_eval_get_input(FILE* fil_u, size_t siz_i)
...
+**                      malloc'd byte buffer
+*/
+c3_y*
+_cw_eval_get_newt(FILE* fil_u, c3_d* len_d)
+{
+  //  TODO: can't reuse u3_newt_decode; coupled to reading from pipe
+  c3_d  i;
+  c3_y* byt_y = c3_realloc(NULL, 5); //  size is start size
+
+  for ( i = 0; i < 5; ++i ) {
```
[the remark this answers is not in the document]
ashelkovnykov: That's exactly what it represents - good tip, thank you.

pkg/vere/main.c
ashelkovnykov: `_cw_eval_get_newt` used `_cw_eval_get_string` as a base, so I wanted to keep their styles identical. I assumed that this is some sort of C trick with which I am unfamiliar that is faster or more defensive. If it's not, I can replace the call with `malloc`. Should I do so in both functions?

pkg/vere/main.c
```diff
@@ -1243,64 +1314,106 @@ _cw_eval(c3_i argc, c3_c* argv[])
...
+    c3_d    len_d;
+    c3_y*   byt_y;
+    u3_weak som;
+    if ( new_t ) {
+      byt_y = _cw_eval_get_newt(stdin, &len_d);
+      som = u3s_cue_xeno(len_d, byt_y);
+    } else {
+      fprintf(stderr, "cue only supports newt encoding (for now)\n");
+      exit(1);
+    }
```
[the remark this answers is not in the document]
ashelkovnykov: Good idea - will do

pkg/vere/main.c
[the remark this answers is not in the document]
mcevoypeter: I appreciate you forcing the issue because I agree that `c3_o` should not be used except when interacting directly with nouns; I'm in favor of replacing all uses of `c3_o` that don't interact with nouns with `c3_t` or similar. If we're going to do so, we should do so methodically, changing from `c3_o` to `c3_t` in a file at a time, with an accompanying PR for each file changed. This conversation quickly turns into a conversation about Vere handles errors, which means we should probably adopt `errno` (or `errno`-like conventions) for returning errors.

With respect to `0` indicating success: yes, but not in the case when that `0` is a boolean (i.e. `bool` is used via `stdbool.h`). From my perspective, a non-zero, non-negative return value only indicates failure when returned from `main()`. In all other cases, I expect a non-zero, non-negative return value to indicate success of some sort with a negative return value reserved for failures. All this to say, I hear where you're coming from, and I don't disagree, but I do think the context here doesn't clearly indicate that `1` is actually failure (unless of course you were to document the functions indicating as much). `-1`, on the other hand, would make it more clear, at least to me, that `return -1` is a failure case.

pkg/vere/main.c
mcevoypeter: `realloc(NULL, size)` is equivalent to `malloc(size)`. From the `realloc` man page:

> If ptr is NULL, then the call [to realloc] is equivalent to malloc(size), for all  values
       of size.

I'd change both `realloc(NULL, /*snip*/)` calls to `malloc(/*snip*/)` calls.

pkg/vere/main.c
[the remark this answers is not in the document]
ashelkovnykov: Introducing `errno` semantics would be great as part of improving the Vere error logging, but too big for what I'm attempting here. However, I think that this proposed commit is still valuable as an incremental improvement.

How would you feel about:
1. Moving this commit to its own PR
2. Replacing `1` with `-1`
3. Possibly replacing `c3_t` with another type like `c3_ws` or `c3_i` (since overloading `c3_t` with multiple `true` values would be out of line with the intent of the type)

?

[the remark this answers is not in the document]
mcevoypeter: Agreed about `errno` semantics on both points. (1)-(3) sound like a good plan. I'd prefer `c3_i` over `c3_ws` since it's in line with POSIX/Linux.

mcevoypeter (approved):

pkg/vere/main.c
```diff
@@ -256,14 +270,7 @@ _main_getopt(c3_i argc, c3_c** argv)
...
       case 5: {  //  urth-loom
-        c3_w lut_w;
-        c3_o res_o = _main_readw(optarg, u3a_bits_max+1, &lut_w);
-        if ( (c3n == res_o) || (lut_w < 20) ) {
-          fprintf(stderr, "error: --urth-loom must be >= 20 and <= %zu\r\n", u3a_bits_max);
-          return c3n;
-        }
-
-        u3_Host.ops_u.lut_y = lut_w;
+        _main_readw_loom(&u3_Host.ops_u.lut_y);
```
joemfb, line 273: The repetition of `--loom` argument parsing logic is irritating. But factoring it out like this changes the error message here (`--loom` vs `--urth-loom`). And it changes the control-flow on errors everywhere (previously returning `c3n` for invalid values, now continuing to parse). By the time you fix both, you're almost back to just duplicating the logic ... Any ideas?

joemfb (changes requested):

mcevoypeter (approved):

pkg/vere/main.c
ashelkovnykov: I noticed that they weren't identical, but hoped that I'd be told it didn't matter during review. How do you feel about have just one duplication: a special case for `urth-loom`?

mcevoypeter: We could add a second parameter to `_main_readw_loom()` that is the argument name, so this call site would become `_main_readw_loom(&u3_Host.ops_u.lut_y, "--urth-loom")`, and the other call sites would become `_main_readw_loom(&u3_Host.ops_u.lom_y, "--loom")`.

joemfb: That works fine for the printf. I think it's important that the process exit with an error given invalid args, so you'll also need to propagate the return.

jalehman:
When @joemfb removes his changes requested we'll reconsider merging this. For now, looks like this is still under deliberation.

pkg/vere/main.c
belisarius222: I think it's better to just duplicate the logic in this case.  If we're worried about future changes needing to be applied in multiple places, we could `#define U3_MIN_LOOM 20`.

mcevoypeter: Looks like @ashelkovnykov already `exit(1)`s on error in `_main_readw_loom()`, so adding a call to `u3_ve_usage()` before exiting should be sufficient to handle the error in case of invalid args.

ashelkovnykov: Quick recap - we have three cases:
1. `"--urth-loom"` and `return c3n`
2. `"--loom"` and `return c3n`
3. `"--loom"` and `exit(1)`

Proposed solution A:
- Change input of `_main_readw_loom()` to `const c3_c*` for custom `fprintf`
- Change return type of `_main_readw_loom()` to `c3_i`
- Return `0` on success
- Return `-1` on failure
- Each usage of `_main_readw_loom()` can determine what failure means

Proposed solution B:
- Leave `_main_readw_loom()` as is (bad input to `--loom` and `--urth-loom` will no longer show usage, just error message)
- Duplicate logic once for `--urth-loom` as special case

Proposed solution C:
- Make error message for `_main_readw_loom()` more generic:
  `fprintf(stderr, "error: loom arguments must be >= 20 and <= %u\r\n", u3a_bits_max);`

All of these are fine by me. I think Solution A is the most correct.

belisarius222: When I first read this, I hadn't realized how many call sites there are to `_main_readw_loom()` -- more than enough to justify having a function.  I think we should go with solution A, since it makes the function useful from all of its call sites, without needing to carve out special cases or make error messages less distinct.

joemfb: Solution A is most correct, C is probably fine as well (so long as the process errors out on invalid args).

ashelkovnykov: Perfect - I think that's exactly what @mcevoypeter was describing in his two comments, so we've got agreement across the board. I'll edit the PR as soon as I'm able.

mcevoypeter: Yep, Solution A sounds best to me too.

mcevoypeter (approved):

pkg/vere/main.c
```diff
@@ -1181,58 +1195,123 @@ _cw_eval_get_input(FILE* fil_u, size_t siz_i)
...
+  //
+  {
+    c3_i err_i;
+    err_i = uv_pipe_init(uv_default_loop(), &std_u.pyp_u, 0);
+    c3_assert(!err_i);
+    uv_pipe_open(&std_u.pyp_u, 1);
+
+    std_u.ptr_v = NULL;
+    std_u.bal_f = _cw_king_fail;
+  }
```
joemfb, line 1313: This usage of newt and libuv is actually a little odd -- there's no subsequent `uv_run()` invocation, so no async callbacks will ever fire for resource cleanup or error handling. The "newt" protocol is so trivial that "reimplementation" (a la `_cw_eval_get_newt()` here) doesn't bother me -- using the async mechanisms here is cumbersome at best.

But this is not blocking, and will be easy to change if it becomes an actual problem.

joemfb (approved):
this will do (with a caveat about the newt usage pattern)

PR #259, opened 2023-02-25
http: add scry and noun caches

lukechampine:
The runtime component of the Eyre cache.

This is a draft as I still have a few open questions (and haven't tested anything yet, except confirming that this at least compiles...). Questions are in-line below.

### DIFF
--- a/pkg/vere/io/http.c
+++ b/pkg/vere/io/http.c
@@ -97,11 +97,18 @@ typedef struct _u3_httd {
   u3_hfig            fig_u;             //  http configuration
   u3_http*           htp_u;             //  http servers
   SSL_CTX*           tls_u;             //  server SSL_CTX*
+  u3p(u3h_root)      sax_p;             //  url->scry cache
+  u3p(u3h_root)      nax_p;             //  scry->noun cache
 } u3_httd;
 
 static void _http_serv_free(u3_http* htp_u);
 static void _http_serv_start_all(u3_httd* htd_u);
 static void _http_form_free(u3_httd* htd_u);
+static void _http_start_respond(u3_hreq* req_u,
+                    u3_noun status,
+                    u3_noun headers,
+                    u3_noun data,
+                    u3_noun complete);
 
 static const c3_i TCP_BACKLOG = 16;
 static const c3_w HEARTBEAT_TIMEOUT = 20 * 1000;
@@ -613,6 +620,68 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
   }
 }
 
+/* _http_cache_respond(): respond with a simple-payload:http
+*/
+static void
+_http_cache_respond(u3_hreq* req_u, u3_noun nun) {
+  // XX check auth
+  u3_noun auth, response_header, data;
+  u3x_qual(u3k(u3t(u3t(nun))), &auth, 0, &response_header, &data);
+  u3_noun status, headers;
+  u3x_cell(response_header, &status, &headers);
+
+  req_u->sat_e = u3_rsat_plan;
+  _http_start_respond(req_u, u3k(status), u3k(headers), u3k(data), c3y);
+  u3z(nun);
+}
+
+typedef struct _cache_scry_cb_t {
+  u3_hreq* req_u;
+  u3_noun  pax;
+} cache_scry_cb_t;
+
+/* _http_cache_scry_cb(): insert scry result into noun cache
+*/
+static void
+_http_cache_scry_cb(void* vod_p, u3_noun nun)
+{
+  cache_scry_cb_t* cbt = vod_p;
+  u3_httd* htd_u = cbt->req_u->hon_u->htp_u->htd_u;
+  u3h_put(htd_u->nax_p, cbt->pax, nun);
+  u3z(cbt->pax);
+  _http_cache_respond(cbt->req_u, nun);
+  c3_free(cbt);
+}
+
+/* _http_req_cache(): attempt to serve http request from cache
+*/
+static c3_o
+_http_req_cache(u3_hreq* req_u)
+{
+  c3_assert(u3_rsat_init == req_u->sat_e);
+
+  u3_httd* htd_u = req_u->hon_u->htp_u->htd_u;
+
+  u3_noun url = u3dc("scot", 't', _http_vec_to_atom(req_u->rec_u->path));
+  u3_weak sac = u3h_get(htd_u->sax_p, url);
+  if ( u3_none == sac ) {
+    return c3n;
+  }
+
+  u3_weak nac = u3h_get(htd_u->nax_p, sac);
+  if ( u3_none == nac ) {
+    // noun not in cache; scry it
+    cache_scry_cb_t* cbt = c3_malloc(sizeof(cache_scry_cb_t));
+    cbt->req_u = req_u;
+    cbt->pax = sac;
+    u3_pier_peek_last(htd_u->car_u.pir_u, u3_nul, c3__ex,
+                      u3_nul, sac, cbt, _http_cache_scry_cb);
+    return c3y;
+  }
+  _http_cache_respond(req_u, nac);
+  return c3y;
+}
+
 /* _http_hgen_dispose(): dispose response generator and buffers
 */
 static void
@@ -985,7 +1054,9 @@ _http_rec_accept(h2o_handler_t* han_u, h2o_req_t* rec_u)
   }
   else {
     u3_hreq* req_u = _http_req_prepare(rec_u, _http_req_new);
-    _http_req_dispatch(req_u, req);
+    if ( c3n == _http_req_cache(req_u) ) {
+      _http_req_dispatch(req_u, req);
+    }
   }
 
   return 0;
@@ -2013,6 +2084,14 @@ _http_ef_http_server(u3_httd* htd_u,
   else if ( c3y == u3r_sing_c("sessions", tag) ) {
     u3_http_ef_auth(htd_u, u3k(dat));
   }
+  //  handles a cache notification
+  //
+  else if ( c3y == u3r_sing_c("grow", tag) ) {
+    // cache paths are /cache/(scot %ud aeon)/(scot %t url)
+    u3_noun pax = u3k(dat);
+    u3_noun url = u3h(u3t(u3t(pax)));
+    u3h_put(htd_u->sax_p, url, pax);
+  }
   //  responds to an open request
   //
   else if ( 0 != (req_u = _http_search_req(htd_u, sev_l, coq_l, seq_l)) ) {
@@ -2216,6 +2295,9 @@ _http_io_exit(u3_auto* car_u)
 {
   u3_httd* htd_u = (u3_httd*)car_u;
 
+  u3h_free(htd_u->sax_p);
+  u3h_free(htd_u->nax_p);
+
   //  dispose of configuration to avoid restarts
   //
   _http_form_free(htd_u);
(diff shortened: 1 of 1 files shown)


pkg/vere/io/http.c
```diff
@@ -546,6 +553,57 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
+
+/* _http_cache_scry_cb(): insert scry result into noun cache
+*/
+static void
+_http_cache_scry_cb(void* vod_p, u3_noun nun)
+{
+  u3_httd* htd_u = vod_p;
+
+  // XX need both the scry path and the noun
+  u3_noun pax = XX;
```
lukechampine, line 565: What's the proper way to pass both a `u3_httd*` and a scry path to this callback? Should I define a struct for this, or can I get away with something ad-hoc declared inline?

pkg/vere/io/http.c
```diff
@@ -546,6 +553,57 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
+  if ( u3_none == sac ) {
+    return c3n;
+  }
+
+  u3_weak nac = u3h_get(htd_u->nax_p, sac);
+  if ( u3_none == nac ) {
+    // noun not in cache; scry it
+    u3_pier_peek_last(htd_u->car_u.pir_u, u3_nul, c3__ex,
+                      u3_nul, sac, htd_u, _http_cache_scry_cb);
+    return c3n; // XX return c3y, and respond async in _http_cache_scry_cb?
```
lukechampine, line 593: currently this returns c3n, so `_http_rec_accept` will forward the request to Eyre as usual while the scry occurs async (and hopefully, by the time the URL is hit again, the scry will have completed). Would it be better to delay the response until the scry completes?

Relatedly, do we need to worry about multiple concurrent requests to the same URL? Currently this will result in redundant scries, which is wasteful, but idk if it's wasteful enough to justify additional complexity.

pkg/vere/io/http.c
```diff
@@ -546,6 +553,57 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
+  }
+
+  // XX check auth
```
lukechampine, line 596: IIUC, assuming https://github.com/urbit/vere/pull/199 is merged with its current API, the check should look like:
```c
if ( (c3y == auth) && (c3y == _http_req_is_auth(htd_u->fig_u, req_u->rec_u) ) {
   return c3n;
}
```

pkg/vere/io/http.c
```diff
@@ -1977,6 +2038,12 @@ _http_ef_http_server(u3_httd* htd_u,
   if ( c3y == u3r_sing_c("set-config", tag) ) {
     u3_http_ef_form(htd_u, u3k(dat));
   }
+  //  handles a cache notification
+  //
+  else if ( c3y == u3r_sing_c("grow", tag) ) {
+    // cache paths are /cache/(scot %u aeon)/(scot %t url)
+    u3h_put(htd_u->sax_p, u3t(u3t(dat)), dat);
```
lukechampine, line 2045: could be more explicit here:
```c
u3_noun pax = dat;
u3_noun url = u3t(u3t(pax));
```

pkg/vere/io/http.c
```diff
@@ -2266,6 +2336,8 @@ u3_auto*
 u3_http_io_init(u3_pier* pir_u)
 {
   u3_httd* htd_u = c3_calloc(sizeof(*htd_u));
+  htd_u->sax_p = u3h_new();
+  htd_u->nax_p = u3h_new_cache(1000); // XX what size?
```
lukechampine, line 2340: should this be configurable? Regardless, we'll want a reasonable default; not sure how to determine that.

pkg/vere/io/http.c
belisarius222: yeah sounds like you need a struct

pkg/vere/io/http.c
belisarius222: Vere should not inject requests for URLs that have mappings to scry paths as Arvo events.  It should just scry.  Otherwise, every incoming request will trigger an event, which would put unneeded load on the ship.

belisarius222: As for multiple concurrent requests, I would say that's not a huge priority to handle well, since the likelihood of many requests all coming in before the scry result gets cached is fairly low.

An optimization beyond that would be to keep a list of open HTTP requests for each pending scry request, then whenever a scry finishes, Vere sends the result to all the open requests and closes them.  That would make sure a sudden flurry of requests for the same path doesn't end up causing several repeated Arvo scry calls.

I don't think that optimization is worth building for this first version.  Better to stick with the simpler version, which I think will work fine.

pkg/vere/io/http.c
belisarius222: Be careful with refcounting when inserting into a hashtable.  The hashtable has different refcounting semantics for key vs. value.

pkg/vere/io/http.c
belisarius222: I'd pick a power of 2 for this number, but 1000 strikes me as vaguely reasonable, if a little on the high side.  512 would be fine given our current use case, I think.  Limiting total memory use would probably be better, but that would require a lot more work.

belisarius222: Configurability would be good, but not crucial in this first phase.

pkg/vere/io/http.c
lukechampine: Ah, so leave the key alone, and `u3k` the value?

pkg/vere/io/http.c
lukechampine: > Vere should not inject requests for URLs that have mappings to scry paths as Arvo events. It should just scry.

How do I scry without going through Arvo? grepping around, I can see a lot of `*peek` functions, but idk which one to use.

belisarius222: I meant it should call something like `u3v_peek()` to call Arvo's `+peek` arm.  This does use the real Arvo core, but it doesn't inject an Arvo *event*, which would call Arvo's `+poke` arm, producing effects and a mutant Arvo state.  Events must also be written to disk, whereas calls to Arvo's `+peek` arm don't change any persistent state, so their inputs don't need to be written to disk.

pkg/vere/io/http.c
belisarius222:
```C
     /* u3h_put(): insert in hashtable.
      **
      ** `key` is RETAINED; `val` is transferred.
      */
        void
        u3h_put(u3p(u3h_root) har_p, u3_noun key, u3_noun val);
```

So yes, that sounds right.

pkg/vere/io/http.c
belisarius222: To address your original question more directly:
> _http_rec_accept will forward the request to Eyre as usual while the scry occurs async

I think this should *not* forward requests to Eyre at all if we're also scrying for them.  If we run both the new scry-based pathway and the current event-based pathway, then we haven't reduced load on the publishing ship, because each incoming HTTP request will still trigger a disk write by injecting it as an Arvo event.

I would note that this code is running in the Urth process, not the Mars process that runs Arvo, so both a scry request and an Arvo event are async -- they both get sent over IPC to the Mars process.

lukechampine:
Updated and tested; basic functionality seems to be working!

Add a cached response:
```
> |pass [%e %set-response '/foo' `[%.n %payload [200 ~] `11+'foo-payload']]
```
Scry it:
```
> .^(cache-entry:eyre %ex /=//=/cache/1/~~~2f.foo)
[ auth=%.n
    body
  [ %payload
      simple-payload
    [ response-header=[status-code=200 headers=~]
      data=[~ [p=11 q=121.352.706.915.279.217.521.291.110]]
    ]
  ]
]
```
And curl it:
```
$ curl localhost:80/foo
foo-payload
```
Logging confirms that the response is being served from the noun cache, not from Eyre.


However, when the entry is a `%scry`, I get a strange crash:
```
> |pass [%e %set-response 'bail' `[%.n %scry /ex/~zod//~2023.3.18..04.30.04..8101/cors]
> .^(cache-entry:eyre %ex /=//=/cache/1/~~bail)
Assertion '0 != u3R->ski.gul' failed in pkg/noun\manage.c:1256
Assertion failed: (0 != u3a_Road->ski.gul), function u3m_soft_esc, file manage.c, line 1256.
```
Not sure what's going on here, will investigate further.

lukechampine:
Oh yeah, let's try benchmarking it!

Without cache:
```
$ go-wrk -c 2 -d 5 http://localhost:80/foo
Running 5s test @ http://localhost:80/foo
  2 goroutine(s) running concurrently
1008 requests in 5.003845417s, 107.30KB read
Requests/sec:           201.45
Transfer/sec:           21.44KB
Avg Req Time:           9.928264ms
Fastest Request:        6.5115ms
Slowest Request:        75.220166ms
Number of Errors:       0
```

With cache:
```
$ go-wrk -c 2 -d 5 http://localhost:80/foo
Running 5s test @ http://localhost:80/foo
  2 goroutine(s) running concurrently
56713 requests in 4.976299745s, 5.90MB read
Requests/sec:           11396.62
Transfer/sec:           1.18MB
Avg Req Time:           175.49µs
Fastest Request:        123.125µs
Slowest Request:        5.91275ms
Number of Errors:       0
```
A 50x speedup!

belisarius222:
50x!  Nice.  Great to see this is running.

The crash is coming from having a .^ in Eyre.  Kernel code is run without a scry handler, so a .^ will always crash in kernelspace.  I think I led you astray there, apologies.  I should have recommended (rof ...) instead of dotket.

lukechampine:
ok, I fixed the scry handling in https://github.com/urbit/urbit/pull/5927, so this is fully-functional now!

pkg/vere/io/http.c
```diff
@@ -613,6 +620,70 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
 
+
+
+/* _http_cache_respond(): respond with a simple-payload:http
+*/
+static void
+_http_cache_respond(u3_hreq* req_u, u3_noun nun) {
+  // XX check auth
+  u3_noun auth, response_header, data;
+  u3x_qual(u3t(u3t(nun)), &auth, 0, &response_header, &data);
```
belisarius222, line 631: I think this function leaks `u3h(nun)` and `u3h(u3t(nun))` -- they're no longer referred to, but the refcount is still nonzero.  The easy way to fix this is to `u3k(u3t(u3t(nun)))` here when destructuring it, then at the end of the function, `u3z(nun)`.

pkg/vere/io/http.c
```diff
@@ -985,6 +1056,9 @@ _http_rec_accept(h2o_handler_t* han_u, h2o_req_t* rec_u)
   }
   else {
     u3_hreq* req_u = _http_req_prepare(rec_u, _http_req_new);
+    if ( c3y == _http_req_cache(req_u) ) {
+      return 0;
```
belisarius222, line 1060: The early return here seems inconsistent to me.  I think this would be clearer like this:
```C
if ( c3n == _http_req_cache(req_u) ) {
  _http_req_dispatch(req_u, req);
}
```
That way the final `return 0` would run no matter which conditional branches were taken.

belisarius222 (changes requested):
Requested a couple of minor changes.  Looks good generally.

lukechampine:
updated and rebased for `next/kelvin/413`

jalehman:
Needs your final stamp @belisarius222

belisarius222 (approved):

vcavallo:
https://github.com/urbit/vere/pull/259#issuecomment-1474883042

I'm salivating

PR #263, opened 2023-02-27
Restore ability to build Vere while using NixOS

midlyx-hatrys:
(After the move to Bazel took it away)

Can build, can run tests (which all even pass), can create usable compile_commands.json, yay.


matthew-levan:
Cool, thanks for opening this. I'm sure some folks will be happy to use it. 

`libtoolize` is `glibtoolize` on macOS, so running `bazel build :urbit` just fails (as it can't find `libtoolize` when running the patched `secp256k1` script). Is there a way we can omit using `libtoolize`, only apply the patch on Linux, or somehow gate the `libtoolize` command in the script to only run on Linux?

midlyx-hatrys:
> Is there a way we can omit using `libtoolize`, only apply the patch on Linux, or somehow gate the `libtoolize` command in the script to only run on Linux?

Did the proper conservative thing and confined this change specifically to NixOS, thanks!

matthew-levan (approved):
Looks good, thanks again.

Issue #190, opened 2021-06-07
Running the binary against a non-existent pier name creates a new comet instead of complaining


lycurgus:
**Describe the bug**
Running `urbit thispierdoesntexist` starts a new comet, instead of complaining and expecting the `-c` flag.

**To Reproduce**
1. run `urbit thispierdoesntexist` (substitute a different pier name if you happen to have a pier with that name :P )
2. observe that the runtime happily boots a new comet

**Expected behaviour**
If the pier isn't found and `-c` (create a new pier) isn't specified, the runtime should throw an error (possibly suggesting the use of `-c` to the user).

**System:**
 - OS: linux64 but also observed on Mac
 - Vere and Urbit OS versions: occurs on all recent vere (from 1.0 to 1.5)

**Additional context**
I've seen a few users run into confusion when they are in the wrong directory, or have moved their pier, and run an older command in an attempt to start up a ship they think should be there. A particular example (which prompted this report) was from a user who migrated their ship into Port:

- normally, run `./urbit sampel-palnet` to start up
- import ship into Port, encounter errors and give up
  - unbeknownst to the user, Port defaults to moving their pier into its own directory to avoid double-booting
- try to go back to using the regular command - but `./urbit sampel-palnet` now boots a comet in the same location because the folder had been removed from its previous spot

Similar things have happened in the absence of Port when e.g. the user installs the binaries and their pier into `~/urbit` and doesn't realise they need to be in that same directory to start back up again later - suddenly `./urbit sampel-palnet` complains about `urbit` being a directory. They do some digging and find that the binary they want is _inside_ that directory, so they run `./urbit/urbit sampel-palnet`, but because they're a directory _up_ from where the pier is, they get a new comet at `~/sampel-palnet` instead of booting `~/urbit/sampel-palnet` as they were intending. (I don't think it's sensible to go as far as searching for pier-shaped directories and asking "did you mean to use \<this other path to a pier\>?" but throwing an error would give the user a clear indication that something was amiss with what they were asking the binary to do).

As it stands these sorts of cases appear to the user like "my planet suddenly turned into a comet and I don't know why".

barter-simsum:
This seems to have been an intentional change ("make -c optional") introduced by @joemfb at some point. Joe, what do you think of requiring `-c` rather than implicitly creating a comet when the pier directory doesn't exist.

https://github.com/urbit/vere/blob/17c14ace3096d6851899c3b8ec7bbbb5a0c1633e/pkg/vere/main.c#L464

[...]

PR #270, opened 2023-02-28
Adjust new pier creation logic

ashelkovnykov:
More elegant solution to #190, plus some other clean up.

The `-c` arg was documented as an opt with an argument, but it operated like a flag. This behaviour has been changed to be consistent with every other argument. In addition, the implicit relationships between `-c`, `-F`, and `-w` have been made explicit.

Tested the following cases:
- Tests for failure on invalid case
    - no arguments
    - pier doesn't exist but no `-c` opt
    - pier already exists but `-c`, `-F`, or `-w` opt present
    - both valid `-c`, `-F`, or `-w` opt and an existing pier provided
    - multiple pier arguments provided
    - `-F` and -w` opts
- Tests for success on valid case
    - open existing pier
    - open existing docked pier
    - new pier from valid `-c`, `-F`, or `-w` opt
    - new pier with both a valid `-F` or `-w` opt and a valid `-c` opt


joemfb (approved):
this will do

PR #274, opened 2023-03-02
fix strange page errors by ensuring single guard page

### LINKED ISSUE #234: Frequent guard page issues causing crashes during OTAs
### Description

There are three problems with the guard page system as implemented:

1. Hitting `loom: not enough memory to recenter the guard page` does not always fail gracefully (as it should).
  a. If the processing of an event results in the exhaustion of all available `loom` space, the processing *should* be aborted, a "failed to process" event should be injected, and the ship should return to its state just prior to the failed attempt.
2. The "strange page" error, which occurs due to old guard pages not being "cleaned up" (i.e., marked as being readable/writable again).
  a. See [urbit/urbit#6125](https://github.com/urbit/urbit/pull/6125) for more details.
  b. Resolved by #274.
3. The system always bails when centering the guard page in the same location, when *MAYBE* it shouldn't.
  a. Further verification is required here. I don't believe this is actually an issue.

The first two issues *sometimes* result in crashes. 

In various instances, the issues are ameliorated with:
- a `|pack` and `|meld`
- a simple ship restart
- an increase in `--loom`
- a "one-by-one" upgrade (for example, if a ship needs to upgrade from `%zuse 418` straight to `%zuse 415`)

In other instances, a ship can't be recovered without an actual fix implemented in a new `vere` binary.

### Issues

See [urbit/urbit#6316](https://github.com/urbit/urbit/issues/6316), for example. Conversely, [urbit/urbit#6318](https://github.com/urbit/urbit/issues/6318) demonstrates a "correct" (i.e., expected) crash scenario and is not relevant to the issues described above.

Here's [an example](https://github.com/urbit/urbit/issues/6193) of a `strange page` error which was resolved by running a `|pack` and `|meld`.

### Reproductions

Memory-intensive operations are the surest way to reproduce these issues. They often occur during OTA updates. I have found a reliable reproduction for the `loom: not enough memory...` crash, but not for the `strange page` error. 

#### `loom: not enough memory...` + crash

To trigger a `loom: not enough memory to recenter the guard page` followed by a crash:
1. Boot a fresh comet using the `urbit-v1.10` binary
2. `|exit`
5. Restart the comet with the latest binary (`v1.22` as of now)
6. `|ota ~wanzod`
7. `|bump, =force %.y`

```
loom: not enough memory to recenter the guard page

recover: dig: meme
behn: bail 1

home: bailing out

Aborted.
```

### Background

In `vere`'s memory management system, we use a guard page to track the approximate center of our free memory block. After each allocation, we re-center the guard page, so it should always be in the center of our free memory. When the guard page *cannot* be re-centered, this indicates we are out of `loom` space (even though there are usually free pages interspersed throughout and recorded in our free list). See the top of `pkg/noun/events.c` for a complete description of the current implementation.

Use of a guard page in the loom was initially implemented during the summer of 2022 and released with `urbit-v1.10`. The initial implementation had some bugs. Various improvements and fixes to the guard page system were released up through `urbit-v1.14`. 

However, `urbit-v1.14` was reverted entirely because of (possibly related) memory corruption issues.  

A more detailed history (with links) is provided below.

#### Guard Page System History/Timeline

| Merged on     | Version | Description | Link |
| ------------- | ------- | ----------- | ---- |
| `~2022.7.8` | `1.10`  | Guard page is added to the loom| [urbit/urbit#5881](https://github.com/urbit/urbit/pull/5881) 
| `~2022.7.7` | `1.10`  | Guard page system is improved| [urbit/urbit#5882](https://github.com/urbit/urbit/pull/5882)
| `~2022.11.14` | `1.12`  | Guard page invariants are enforced| [urbit/urbit#6079](https://github.com/urbit/urbit/pull/6079)
| `~2022.12.2` | `1.14`  | Multiple guard pg. bug is fixed (`strange page`)| [urbit/urbit#6125](https://github.com/urbit/urbit/pull/6125)
| `~2022.12.2` | `1.14`  | Related: snapshot invariants are enforced/doc'd| [urbit/urbit#6126](https://github.com/urbit/urbit/pull/6126)
| `~2022.12.2` | `1.14`  | Guard page invariants are re-enforced after snapshot work| [urbit/urbit#6127](https://github.com/urbit/urbit/pull/6127)
| `~2022.12.13` | `1.14`  | Entire snapshot system (incl. guard pages) interface refactored (with demand paging)| [urbit/urbit#6152](https://github.com/urbit/urbit/pull/6152)
| `~2022.1.3` | `1.15`  | `1.14` reverted to `1.13` (as `1.15`) | [urbit/urbit#6186](https://github.com/urbit/urbit/pull/6186)

matthew-levan:
Partially fixes #234.

This PR is a port of @joemfb's [urbit/urbit#6125](https://github.com/urbit/urbit/pull/6125).

Note that any stale `PROT_NONE` pages in live ship looms that currently exist will be cleaned out upon restart (i.e., when they upgrade to the new `vere`, at the least).

### DIFF
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -1258,8 +1258,27 @@ void
 u3e_ward(u3_post low_p, u3_post hig_p)
 {
 #ifdef U3_GUARD_PAGE
-  if ( (low_p > gar_pag_p) || (hig_p < gar_pag_p) ) {
+  const u3p(c3_w) gar_p = gar_pag_p;
+
+  if ( (low_p > gar_p) || (hig_p < gar_p) ) {
     _ce_center_guard_page();
+
+    if ( 0 != mprotect(u3a_into(gar_p),
+                       pag_siz_i,
+                       (PROT_READ | PROT_WRITE)) )
+    {
+      fprintf(stderr, "loom: failed to unprotect old guard page: %s\r\n",
+                      strerror(errno));
+      c3_assert(0);
+    }
+
+    {
+      c3_w pag_w = gar_p >> u3a_page;
+      c3_w blk_w = (pag_w >> 5);
+      c3_w bit_w = (pag_w & 31);
+
+      u3P.dit_w[blk_w] |= (1 << bit_w);
+    }
   }
 #endif
 }


pkg/noun/events.c
```diff
@@ -1258,8 +1258,27 @@ void
...
+                      strerror(errno));
+      c3_assert(0);
+    }
+
+    {
+      c3_w pag_w = gar_p >> u3a_page;
+      c3_w blk_w = (pag_w >> 5);
+      c3_w bit_w = (pag_w & 31);
+
+      u3P.dit_w[blk_w] |= (1 << bit_w);
```
barter-simsum, line 1280: why is it necessary to dirty the block containing the guard page? This makes sense in `u3e_fault` ofc, but why here?

matthew-levan: We mark old guard pages with `PROT_READ | PROT_WRITE` *and* mark it dirty because it handles for the (probably rare) case that the old guard page had been sitting on pending-to-be-saved-to-snapshot data.

belisarius222 (approved):

pkg/noun/events.c
joemfb: The only pages that should be marked clean and protected read-only are pages that are present-in/unmodified-from the snapshot (in practice, portions of the home road heap). Everything else should be dirty/read-write, except for the (singular) guard page, which should be dirty/prot_none.

PR #280, opened 2023-03-11
build: fix docker pace deploy

ashelkovnykov:
Resolves [builds failing when attempting to tag release with pace on DockerHub](https://github.com/urbit/vere/actions/runs/4298049619/jobs/7491731178).

Can we please review this more carefully? I don't want this to be a repeat of #272. I'm also surprised that no one has commented on [every build on push to deploy failing for 10 days](https://github.com/urbit/vere/actions).

I genuinely don't know Bazel well. I'm attempting to implement the correct syntax using the existing files and Bazel docs as references, but I don't know that a PR is correct when I submit it, especially since it's impossible to test something like this locally.

### DIFF
--- a/docker/BUILD.bazel
+++ b/docker/BUILD.bazel
@@ -77,7 +77,7 @@ genrule(
 
 genrule(
     name = "pace_tag",
-    srcs = ["PACE"],
+    srcs = ["//:PACE"],
     outs = ["docker_pace_tag"],
     cmd = """
     tr -d '\n' < $< > $@


docker/BUILD.bazel
```diff
@@ -77,7 +77,7 @@ genrule(
```
mcevoypeter, line 77: You need to add `export_files(["PACE"])` to the top of `BUILD.bazel` right below `load("//bazel:common_settings.bzl", "string_flag")` so the `//docker` package has access to `PACE`, which exists outside of the `//docker` package. You can then test by running `bazel build //docker:push_pace_x86_64`.

matthew-levan: I'll fix this thanks Peter.

mcevoypeter:
> Ok yea this should actually work.

@matthew-levan `bazel build //docker:push_pace_x86_64` unfortunately fails for me locally on this branch without adding the `export_files()` command I mentioned above. Can you open another PR that adds the `export_files()` command so @ashelkovnykov doesn't have to open yet another Docker-related PR?

matthew-levan:
Yes sorry for testing this live (so to speak), I will fix it now.

mcevoypeter:
Thanks! Appreciate it.

matthew-levan:
Agreed we should just add this to the contributors doc. cc @ashelkovnykov for his thoughts though too.

ashelkovnykov:
Yes - I was hoping to get **two** people to look at this and the previous PR before merging, namely @matthew-levan and @mcevoypeter. If both of you had said "LGTM", I would have merged.

PR #286, opened 2023-03-14
pier: decrement zuse/lull

matthew-levan:
Support for `zuse %414` and `lull %325`.

### DIFF
--- a/pkg/vere/pier.c
+++ b/pkg/vere/pier.c
@@ -676,8 +676,8 @@ _pier_wyrd_fail(u3_pier* pir_u, u3_ovum* egg_u, u3_noun lud)
 //  XX organizing version constants
 //
 #define VERE_NAME  "vere"
-#define VERE_ZUSE  415
-#define VERE_LULL  326
+#define VERE_ZUSE  414
+#define VERE_LULL  325
 
 /* _pier_wyrd_aver(): check for %wend effect and version downgrade. RETAIN
 */


mcevoypeter:
Will decrementing the Zuse and Lull versions in the runtime before they're decremented in Arvo causes issues? Lull is still on 326 and Zuse is still on 415 on `develop` in https://github.com/urbit/urbit.

matthew-levan:
> Will decrementing the Zuse and Lull versions in the runtime before they're decremented in Arvo causes issues? Lull is still on 326 and Zuse is still on 415 on `develop` in https://github.com/urbit/urbit.

I don't think so, but merging of urbit/urbit#6365 is imminent. I'm trying to test locally with a `vere` build from this branch and the Arvo files from urbit/urbit#6365 but unfortunately I haven't been able to solid a pill without running out of memory yet.

philipcmonk:
In general, it is actually preferable for the runtime to decrement kelvins before arvo.  The runtime must be updated before the arvo update can apply.  The reason we don't decrement even more ahead of time is that once you decrement the kelvin in the runtime, you're declaring that no further runtime changes will need to be made for that kelvin version.  If you later decide you need more runtime changes for that kelvin to run, you must decrement a second time.

mcevoypeter:
Thanks for the clarification, @philipcmonk.

mcevoypeter (approved):

mcevoypeter:
@matthew-levan it looks like the `macos-x86_64` CI failure was due to a spurious network failure. I just reran the job. Please hold off on merging until it passes, just to be sure.

PR #289, opened 2023-03-16
pier: attempt safe `slog`

### LINKED ISSUE #288: `u3`: `slog` isn't as safe as it could be
One symptom of the borked "mega-upgrade" issue (and crash) described in #234 is that a malformed `tank` is produced (either because it's too long or because it's invalid in another way which the Hoon compiler doesn't catch) and sent to `u3_auto_bail_slog` for printing to `stderr`. Unfortunately, `u3_auto_bail_slog` doesn't handle the malformed `tank` gracefully and crashes the entire process instead.

We should make `u3_auto_bail_slog` safer, because the last thing a programmer wants when dealing with an error is for the backtrace printer to crash too.

matthew-levan:
I'm not sure how to go about handling `%error` results returned by `u3m_soft` calls yet, so this is still a WIP.

Resolves #288.

### DIFF
--- a/pkg/vere/pier.c
+++ b/pkg/vere/pier.c
@@ -2397,11 +2397,11 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
       u3_noun cor = u3nc(u3k(u3h(gat)), u3nc(sam, u3k(u3t(u3t(gat)))));
       res = u3m_soft(0, u3n_kick_on, cor);
     }
-    if (0 == u3h(res) ) {
+    if ( c3y == u3h(res) ) {
       u3_noun wol = u3k(u3t(res)); 
       _pier_dump_wall(fil_u, wol);
     } else if ( c3__exit == u3h(res) ) {
-      //  ???
+      fprintf(fil_u, "####\r\n");  //  stack trace printing failed
     }
     u3z(res);
     u3z(gat);
--- a/pkg/vere/pier.c
+++ b/pkg/vere/pier.c
@@ -2400,7 +2400,7 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
     if (0 == u3h(res) ) {
       u3_noun wol = u3k(u3t(res)); 
       _pier_dump_wall(fil_u, wol);
-    } else if ( 0 == strcmp(u3r_string(u3k(u3h(res))), "exit") ) {
+    } else if ( c3__exit == u3h(res) ) {
       //  ???
     }
     u3z(res);
--- a/pkg/vere/pier.c
+++ b/pkg/vere/pier.c
@@ -2401,7 +2401,7 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
       u3_noun wol = u3k(u3t(res)); 
       _pier_dump_wall(fil_u, wol);
     } else if ( c3__exit == u3h(res) ) {
-      fprintf(fil_u, "####\r\n");  //  stack trace printing failed
+      fprintf(fil_u, "%%%%r\n");  //  stack trace printing failed
     } else {
       fprintf(fil_u, "pier: tank rendering failed\r\n");
     }
--- a/pkg/vere/pier.c
+++ b/pkg/vere/pier.c
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
       _pier_dump_tape(fil_u, u3k(u3t(tac)));
     }
   }
-  //  We are calling nock here, but hopefully need no protection.
+  //  We are calling nock here, so we virtualize for safety.
   //
   else {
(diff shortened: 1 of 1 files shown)


pkg/vere/pier.c
```diff
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
...
+    {
+      u3_noun sam = u3nc(tab_l, col_l);
+      u3_noun cor = u3nc(u3k(u3h(gat)), u3nc(sam, u3k(u3t(u3t(gat)))));
+      res = u3m_soft(0, u3n_kick_on, cor);
+    }
+    if (0 == u3h(res) ) {
+      u3_noun wol = u3k(u3t(res)); 
+      _pier_dump_wall(fil_u, wol);
+    } else if ( c3__exit == u3h(res) ) {
+      //  ???
```
belisarius222, line 2404: I think the thing to do in this case is to print something to indicate that printing the stack trace failed.  I seem to recall that in hoon, it will print `####` in that case, so we could replicate that here perhaps.  I would also print the bail mote that came out of the serf.

pkg/vere/pier.c
```diff
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
...
-    _pier_dump_wall(fil_u, wol);
+    u3_noun gat = u3v_wish("wash");
+    u3_noun res;
+    
+    {
+      u3_noun sam = u3nc(tab_l, col_l);
+      u3_noun cor = u3nc(u3k(u3h(gat)), u3nc(sam, u3k(u3t(u3t(gat)))));
+      res = u3m_soft(0, u3n_kick_on, cor);
+    }
+    if (0 == u3h(res) ) {
```
belisarius222, line 2400: style: need space between `(` and `0`

belisarius222, line 2400: The `0` here is really `c3y`, I think -- it's the tag in a Hoon `+each`.

pkg/vere/pier.c
matthew-levan: Great call, thanks!

pkg/vere/pier.c
```diff
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
...
   //
   else {
-    u3_noun wol = u3dc("wash", u3nc(tab_l, col_l), u3k(tac));
-
-    _pier_dump_wall(fil_u, wol);
+    u3_noun gat = u3v_wish("wash");
+    u3_noun res;
+    
+    {
+      u3_noun sam = u3nc(tab_l, col_l);
```
joemfb, line 2396: The sample here should be `[[tab col] tac]` -- `u3nc(u3nc(tab_l, col_l), u3k(tac))`. See `+wash` in `hoon.hoon`.

pkg/vere/pier.c
```diff
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
...
+      u3_noun cor = u3nc(u3k(u3h(gat)), u3nc(sam, u3k(u3t(u3t(gat)))));
+      res = u3m_soft(0, u3n_kick_on, cor);
+    }
+    if ( c3y == u3h(res) ) {
```
joemfb, line 2400: Style nit: the vere idiom for `%0` is actually `u3_blip`. The return of `u3m_soft()` isn't `each`, it's (vere's superset of) `$toon`.

pkg/vere/pier.c
```diff
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
...
+      u3_noun wol = u3k(u3t(res)); 
+      _pier_dump_wall(fil_u, wol);
+    } else if ( c3__exit == u3h(res) ) {
+      fprintf(fil_u, "####\r\n");  //  stack trace printing failed
```
joemfb, line 2404: `+mook` already uses `####` for invalid tanks. So this is either good reuse of an existing convention, or confusing ambiguation of crash conditions. I'm not sure which!

pkg/vere/pier.c
```diff
@@ -2386,12 +2386,25 @@ u3_pier_tank(c3_l tab_l, c3_w pri_w, u3_noun tac)
...
+    }
```
joemfb, line 2405: We probably want a generic `else` branch to capture any crash type that prevents us from rendering the tank. (Separate output here probably makes `####` more defensible above.)

joemfb (changes requested):
IIRC, I did this once and it was shockingly slow. It'll be worth testing that once the sample is fixed. If it's still slow, we can probably cheat by printing from the inner road (copying out would presumably be the bulk of the cost).

pkg/vere/pier.c
matthew-levan: We could do `%%%%` or something, just to keep things tidy.

PR #305, opened 2023-03-23
Remote Scry

### LINKED ISSUE #303: Add Remote Scry
Placeholder issue mirroring: https://github.com/urbit/urbit/issues/5788

wicrum-wicrun:
Resolves #303


pkg/vere/io/ames.c
```diff
@@ -1494,6 +2323,7 @@ _ames_io_info(u3_auto* car_u)
   u3_ames*  sam_u = (u3_ames*)car_u;
 
   return u3i_list(
+    u3_pier_mase("scry-cache",       u3i_word(u3h_mark(sam_u->fin_s.sac_p))),
```
joemfb, line 2326: This leaks memory -- calls to `*_mark()` functions must be followed by a call to `u3a_sweep()`, and that requires that all roots are properly recorded and marked. You should be able to use `u3h_count()` and `u3h_discount()` to do a granular measurement (effectively a partial mark/sweep collection).

joemfb: Also, this should be duplicated as a printf in `_ames_io_info()`.

PR #317, opened 2023-03-27
Command Line Option to set the time between snapshots

mopfel-winrux:
This PR adds a command line option that allows the user to set the time in minutes between snapshots. This helps resolve the issues discussed in https://github.com/urbit/urbit/issues/4150. 

By allowing the user to set the time between shapshots it lets them decide how long they want to possibly wait to replay events. This also may allow urbit to run on an SD Card (which has ~10,000 write cycles.); reducing the cost to home host.

### DIFF
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -309,7 +309,9 @@ _main_getopt(c3_i argc, c3_c** argv)
         break;
       }
       case c3__snap: {
-        u3_Host.ops_u.sap_w = _main_readw(optarg, 65536, &arg_w) *60.0;
+        if ( c3n == _main_readw(optarg, 65536, &arg_w) ) {
+          return c3n;
+        } else u3_Host.ops_u.sap_w = arg_w*60;
         break;
       }
       //  opts with args
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -312,7 +312,7 @@ _main_getopt(c3_i argc, c3_c** argv)
         if ( c3n == _main_readw(optarg, 65536, &arg_w) ) {
           return c3n;
         } else {
-          u3_Host.ops_u.sap_w = arg_w*60;
+          u3_Host.ops_u.sap_w = arg_w * 60;
           if ( 0 == u3_Host.ops_u.sap_w) 
             return c3n;
         }
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -181,6 +181,7 @@ _main_init(void)
   u3_Host.ops_u.hap_w = 50000;
   u3_Host.ops_u.kno_w = DefaultKernel;
 
+  u3_Host.ops_u.sap_w = 120;    /* aka 2 minutes */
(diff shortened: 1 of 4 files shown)


pkg/vere/main.c
```diff
@@ -306,6 +308,16 @@ _main_getopt(c3_i argc, c3_c** argv)
         u3_Host.ops_u.doc = c3n;
         break;
       }
+      case c3__snap: {
+        if ( c3n == _main_readw(optarg, 65536, &arg_w) ) {
+          return c3n;
+        } else {
+          u3_Host.ops_u.sap_w = arg_w*60;
```
matthew-levan, line 315:
```suggestion
          u3_Host.ops_u.sap_w = arg_w * 60;
```

pkg/vere/main.c
```diff
@@ -727,6 +739,7 @@ u3_ve_usage(c3_i argc, c3_c** argv)
     "-p, --ames-port PORT          Set the ames port to bind to\n",
     "    --http-port PORT          Set the http port to bind to\n",
     "    --https-port PORT         Set the https port to bind to\n",
+    "    --snap-time TIME          Set the snapshotting rate in minutes (> 0)\n",
```
matthew-levan, line 742:
```suggestion
    "    --snap-time TIME          Set the minutes between snapshots (> 0)\n",
```

pkg/vere/save.c
```diff
@@ -49,7 +49,7 @@ u3_save_io_init(u3_pier *pir_u)
 
   sav_u->tim_u.data = pir_u;
   uv_timer_init(u3L, &sav_u->tim_u);
-  uv_timer_start(&sav_u->tim_u, _save_time_cb, 120000, 120000);
+  uv_timer_start(&sav_u->tim_u, _save_time_cb, u3_Host.ops_u.sap_w*1000, u3_Host.ops_u.sap_w*1000);
```
matthew-levan, line 52:
```suggestion
  uv_timer_start(&sav_u->tim_u, _save_time_cb, u3_Host.ops_u.sap_w * 1000, u3_Host.ops_u.sap_w * 1000);
```

pkg/vere/main.c
barter-simsum: personally find the original phrasing a bit more clear

mopfel-winrux:
> There should probably be a hard limit defined in the vere arg-parser. Certainly `UINT32_MAX / 1000` is a hard limit ;). Additionally lower bounds. What happens if I pass 0. @mopfel-winrux

If you pass a 0 the opts parser returns c3n and the software crashes

barter-simsum:
Maybe 0 should just disable snapshotting altogether. differing to @joemfb

pkg/vere/main.c
mopfel-winrux: I like rate better than minutes between as well

pkg/vere/main.c
mopfel-winrux: fixed

belisarius222 (approved):

PR #331, opened 2023-04-04
Add conn.c usage guide

### LINKED ISSUE #304: Write Conn/Khan usage guide
The Conn/Khan control plane is still relatively poorly understood, We'll benefit from a usage guide for the control plane, including the different types of requests (i.e. `%fyrd`, `%peek`, etc), how each type of request is used, and plenty of practical examples explaining how the control plane can be used to issue commands and consume data from a running ship. In particular, the guide should clearly explain how to run Dojo commands like `|mass`, `|install`, and `|ota` via Conn/Khan.

Related: #279.

ashelkovnykov:
Resolves #304 

This location is temporary housing until a separate repo for Urbit tools exists (`urbit/tools`?), and also so that other devs can review the contents.


bin/README.md
```diff
@@ -0,0 +1,466 @@
+## Background
+
+Prior to 2022, the only way to interact with a running ship from Earth was via
+HTTP requests sent to the `%lens` agent.In addition to `%lens`, there was a
```
matthew-levan, line 4:
```suggestion
HTTP requests sent to the `%lens` agent. In addition to `%lens`, there was a
```

bin/README.md
```diff
@@ -0,0 +1,466 @@
...
+#### `%ovum`
+
+The argument to an `%ovum` command is a raw kernel move which is injected
+directly into the Arvo event loop. This is a very powerful - and potentially
+dangerous - tool. For example, if a ship somehow got into a state where Clay was
+no longer working properly (meaning new files could not be compiled to fix the
+state of the kernel), the source code for a new, working Clay could be directly
+injected into the ship using an `%ovum`.
+
+The output of a `%ovum` command is:
```
matthew-levan, line 75:
```suggestion
The output of an `%ovum` command is:
```

bin/README.md
```diff
@@ -0,0 +1,466 @@
...
+/path/to/click -kp /path/to/pier/zod $'=/  m  (strand ,vase)  ;<  our=@p  bind:m  get-our  ;<  ~  bind:m  (poke [our %hood] %kiln-install !>([%desk ~sampel-palnet %desk]))  (pure:m !>(\\\'success\\\'))'
+```
+
+#### `|install ~sampel-palnet %desk, =local %my-desk`
+
+```
+echo "[0 %ovum [%g /test [%deal [~zod ~zod] %hood %raw-poke %kiln-install %my-desk ~sampel-palnet %desk]]]" | ./urbit eval -jn | nc -U -W 1 /path/to/pier/zod/.urb/conn.sock | ./urbit eval -cn
+```
+```
+/path/to/clickk -kp /path/to/pier/zod $'=/  m  (strand ,vase)  ;<  our=@p  bind:m  get-our  ;<  ~  bind:m  (poke [our %hood] %kiln-install !>([%my-desk ~sampel-palnet %desk]))  (pure:m !>(\\\'success\\\'))'
```
matthew-levan, line 397:
```suggestion
/path/to/click -kp /path/to/pier/zod $'=/  m  (strand ,vase)  ;<  our=@p  bind:m  get-our  ;<  ~  bind:m  (poke [our %hood] %kiln-install !>([%my-desk ~sampel-palnet %desk]))  (pure:m !>(\\\'success\\\'))'
```

matthew-levan (approved):
This is really great, thanks for writing this. 

The only substantial suggestion I have is to format the click examples with triple  tic `.hoon` file codeblocks instead of long, inline hoon-as-bash-strings. Where you say "even more conveniently" is an example of a start of this. To finish that particular example, you'd just have to write the `poke.hoon` as a codeblock right underneath. 

Perhaps leaving a few of the "inline" examples is wothwhile as well, but I think most serious users (i.e., hosting providers) will want to store their click inputs as Hoon files in folders.

ashelkovnykov:
@matthew-levan I understand what you're saying and agree that it's probably the way in which hosting providers will use the example commands. However, I think it would make the flow of the document worse and it would eliminate the ability for people to test out these commands with a one-line copy+paste.

Beyond that, I can't be certain exactly what each hosting provider wants to do with these commands. I think it best if they develop their own library of `.hoon` threads that they call via click (for now) or via a thick client (eventually).

That said, we could probably make a separate directory in the eventual `urbit/tools` repo where devs and companies share scripts that they're using. I think that would be a great place to populate with `.hoon` versions of the example calls.

mopfel-winrux:
This readme contradicts the code comment below:
https://github.com/urbit/vere/blob/develop/pkg/vere/io/conn.c#L82

I'm looking for guidance as to if injecting an %ovum is fine if understood

ashelkovnykov:
@mopfel-winrux `%ovum` commands are safe to run if understood. "Void your warranty" is a really vague term, meant in jest to get across the idea that in the worst case:
- Bad `%urth` and `%peel` requests can fail
- Bad `%fyrd` and `%peek` requests can crash the ship
- Bad `%ovum` requests can necessitate a breach

joemfb (dismissed):
`bin/` has always been used for pills and other binary artifacts. It's configured to use LFS, but these scripts are not (and should not be) in LFS. This leads to (at least) warnings and other wonkiness.

There is a `sh/` directory, which is not a bad candidate. Or you could make a new `pkg/click`. But please move them out of `bin/`.

bin/README.md
```diff
@@ -0,0 +1,466 @@
...
+### click
+
+[click](https://github.com/urbit/vere/blob/develop/bin/click) is a `bash` thin
+client which auto-formats `-eval` and `-khan-eval` thread calls via `%fyrd`
+requests to `conn.c` and coordinates chaining together the appropriate commands
+to execute those requests on a running ship.
+
+Using click, a call like:
+```
+echo $'[0 %fyrd %base %khan-eval %noun %ted-eval \'=/  m  (strand ,vase)  ;<  ~  bind:m  (poke [~zod %hood] %helm-hi !>(\\\'\\\'))  (pure:m !>(\\\'success\\\'))\']' | ./test-urbit eval -jn | nc -U -W 1 ~/Urbit/test-piers/zod/.urb/conn.sock | ./test-urbit eval -cn
```
mcevoypeter, line 287: I think this would be more readable if split over multiple lines, but maybe you considered that and are optimizing instead for copy-pastability.

mcevoypeter (commented):
Looks great. My only question is the line length of examples and its impact on the tradeoff between readability and copy-pastability.

bin/README.md
barter-simsum: I agree that the examples would be a lot more readable if split. Why not split on `|`? You can terminate with it to avoid backslashes.

Also, @ashelkovnykov looks like you have a couple `./test-urbit` instances here instead of `./urbit`

ashelkovnykov: @mcevoypeter Yeah, I was going for "easiest to copy-paste", but I think you two might be right about splitting the commands along the pipes

@barter-simsum Good catch, will change those

ashelkovnykov:
@joemfb Will do.

I've been talking with @mcevoypeter and @mopfel-winrux about how something like this guide, click, `herb`, etc. should go into a separate repo for helper tools. Anyone have any objections to creating an `/urbit/tools` repo and moving this + click there after review?

matthew-levan:
No objections.

bin/README.md
```diff
@@ -0,0 +1,466 @@
...
+
+Currently, there are a number of minor issues and one major issue impacting
+interactions between Earth and Mars.
+
+The minor issues are:
+- `conn.c`'s simulated namespace for `%peel`
+    - Unprincipled namespace simulation for no reason other than consistency
+      with Arvo scry
+- `/mass` path for `conn.c` `%peel` not implemented
+- No `mass` mark in Arvo, so attempting to scry for `|mass` with `%peek`
```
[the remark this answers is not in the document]
ashelkovnykov: Command:
```
echo "[0 %peek %| %once %$ %base /whey]" |
./urbit eval -jn |
nc -U -W 1 test-piers/zod/.urb/conn.sock |
./urbit eval -cn
```
Output in `~zod`:
```
conn: peek 0v0
clay: read-at-aeon fail [desk=%base care=%c case=[%da p=~2023.4.5..19.39.59..9239] path=/noun/mass]
clay: no files match /mar/mass/hoon
[%error-building-cast %noun %mass]
[%error-building-tube %noun %mass]
peek no tube from noun to mass
conn: moor bail -32 broken pipe
~zod:dojo> newt: write failed broken pipe

...

conn: moor bail -32 broken pipe
~zod:dojo> newt: write failed broken pipe
Segmentation fault (core dumped)

```

pkg/click/README.md
```diff
@@ -0,0 +1,514 @@
...
+
+```
+echo "[0 %ovum [%g /test [%deal [~zod ~zod] %hood %raw-poke %kiln-install %base ~zod %desk]]]" |
+/path/to/urbit eval -jn |
+nc -U -W 1 /path/to/pier/zod/.urb/conn.sock |
+/path/to/urbit eval -cn
+```
+```
+/path/to/click -kp /path/to/pier/zod \
+$'=/  m  (strand ,vase)  ;<  our=@p  bind:m  get-our  ;<  ~  bind:m  (poke [our %hood] %kiln-install !>([%base our %desk]))  (pure:m !>(\\\'success\\\'))'
```
mcevoypeter, line 406: Should this be `[%base ~bus %desk]` instead of `[%base our %desk]`?

ashelkovnykov: You're correct - c-p typo from the above command

PR #334, opened 2023-04-06
ci: run fake ship tests under gc

### LINKED ISSUE #332: fake-ship tests are not run under gc
This looks like a regression during demonorepofication.

matthew-levan:
Adds `--gc` to the boot phase of the fake ship tests as the binary we build for them already uses the correct `copts`.

Resolves #332.

### DIFF
--- a/pkg/vere/BUILD.bazel
+++ b/pkg/vere/BUILD.bazel
@@ -332,7 +332,7 @@ genrule(
 
     set -x
 
-    $(execpath :urbit) --lite-boot --daemon ./pier 2> urbit-output
+    $(execpath :urbit) --lite-boot --daemon --gc ./pier 2> urbit-output
 
     port=$$(grep loopback ./pier/.http.ports | awk -F ' ' '{print $$1}')
 


barter-simsum:
> When I tried to boot an existing pier with the full debug binary, it just printed a bunch of hex addresses and eventually a "boot failed". Is this normal behavior?

No. `--config=dbg` only optimizes at `-O0` and turns on `c3_dessert` debug asserts.

barter-simsum:
@matthew-levan 

`--config=dbg` isn't a full debug configuration and we shouldn't be running it in ci because the unoptimized binary is quite a bit slower.

.github/workflows/shared.yml
```diff
@@ -135,6 +135,7 @@ jobs:
         if: ${{ matrix.target == 'linux-x86_64' && inputs.fake_tests }}
         run: |
           # See https://github.com/urbit/vere/issues/40.
+          bazel build :urbit --config=mem_dbg
```
barter-simsum, line 138: We probably shouldn't use the `--config=` config definitions defined in .bazelrc in ci. These are only meant for local development usage. Much more foolproof and easier to reason about would be the following:

`bazel build --copt='-DU3_CPU_DEBUG` --copt='-DU3_MEMORY_DEBUG' :urbit`

barter-simsum: Actually, I believe all of these options are already getting invoked. https://github.com/urbit/vere/blob/7ef74854aa5c0d68a155e113299e8f1669f73a96/.bazelrc#L34

matthew-levan: Right, so maybe all we need to do is add the `--gc` flag to the `urbit` flags we use for the fake ship tests.

matthew-levan:
Ok, so to recap: we just needed to add `--gc` to the boot phase of the fake ship tests.

barter-simsum (approved):

PR #336, opened 2023-04-06
Updates to Docker build and README

ashelkovnykov:
- Updates README (here and in DockerHub)
- Fixes bug that posted README updates to `tloncorp/urbit` instead of `tloncorp/vere`
- Moves DockerHub deploy tasks in build to separate workflow
- Adds a workflow which allows on-demand GitHub Action to redeploy Docker image to DockerHub


mopfel-winrux:
Can we also include `click.sh` in the docker container? It would make maintaining piers a lot easier

PR #338, opened 2023-04-08
ci: fix `macos-x86_64`-specific build/test `clang` versions

### LINKED ISSUE #337: Outdated `clang` version for `macos-x86_64` CI
GitHub updated their `macos-12` hosted runner image, so our hardcoded `clang` version in `.github/workflows/shared.yml` is now outdated and breaks `macos-x86_64` jobs.

matthew-levan:
Resolves #337.


mcevoypeter (dismissed):

ashelkovnykov:
@matthew-levan @mcevoypeter Can one of you leave a quick comment about the changes:
- why downgrade to `14.0.0` in the workflow
- why mismatch between workflow and Bazel (`14` vs `15`)

mcevoypeter:
The path to `brew`-installed LLVM headers changed, which requires an update to the `sys_includes` attribute of the `brew-clang-macos-x86_64-config` `cc_toolchain_config` target. The `--clang_version` was erroneously specified as `14.0.6` before, when it needs to be `15.0.7`. Interestingly, an incorrect version supplied to the `compiler_version` attribute of `brew-clang-macos-x86_64-config` doesn't seem to have had a negative impact. I've updated the `--clang_version` to match the LLVM install (15.0.7) anyway.

ashelkovnykov (approved):

matthew-levan:
What happened here @mcevoypeter and @ashelkovnykov is that GitHub updated their `macos-12` hosted runner image, thereby removing `clang` version `14.0.6` and adding version `15.0.7`.

PR #341, opened 2023-04-11
docker: use build cache; fix failing flow

ashelkovnykov:
Resolves failing GitHub build action

Tested [here](https://github.com/urbit/vere/actions/runs/4669925658/jobs/8269011549)

### DIFF
--- a/.github/workflows/docker-shared.yml
+++ b/.github/workflows/docker-shared.yml
@@ -16,6 +16,34 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      #
+      # BUILD OR READ FROM CACHE
+      #
+
+      - name: chown /usr/local
+        run: |
+          # See https://github.com/actions/cache/issues/845.
+          sudo chown $(whoami) /usr/local && (chown -R $USER /usr/local/*-musl || true)
+
+      - name: Set up build cache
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-shared.yml
```diff
@@ -16,6 +16,38 @@ jobs:
...
+            # Cache musl libc toolchains.
+            /usr/local/*-musl
+
+      - name: Install toolchains
+        if: steps.build-cache.outputs.cache-hit != 'true'
+        run: bazel run //bazel/toolchain:x86_64-linux-musl-gcc
+
+      - name: Build binary
+        if: steps.build-cache.outputs.cache-hit != 'true'
+        run: bazel build :urbit
```
mcevoypeter, line 45: This can be removed since the `//docker:push*` targets all depend on `//docker:image_x86_64`, which in turn depends on `//pkg/vere:urbit`.

ashelkovnykov: Confirmed, see comment below

ashelkovnykov:
Tested again [here](https://github.com/urbit/vere/actions/runs/4670294521/jobs/8269842190) after @mcevoypeter 's comment. Also confirms that cache is working correctly.

PR #343, opened 2023-04-12
Replace `c3_assert()` with `u3_assert()` or `exit()`

### LINKED ISSUE #273: Replace overuse of `c3_assert()` with `exit()` and sane exit code scheme
We need to decide on a meaningful set of exit codes and use them effectively when encountering a fatal error. We currently use `c3_assert()` to handle many fatal errors, which is less than ideal.

mcevoypeter:
Resolves #273.

This PR is a step toward more robust error handling. It renames `c3_assert()` as `u3_assert()` and replaces a number of `c3_assert()` calls designed to signal fatal errors with `exit()` and an appropriate `errno`-based status code. Some `c3_assert()` calls should not result in `exit()` but instead in `u3m_bail(c3__oops)`. To the best of my knowledge, the changes maintain the original error handling intent of the code.

## Testing

```console
$ bazel build :urbit
$ ./urbit -F zod -B path/to/solid/pill
<snip>
~zod:dojo> |exit
$ ./urbit zod
$ bazel test --build_tests_only ...
```

### DIFF
--- a/pkg/c3/defs.h
+++ b/pkg/c3/defs.h
@@ -21,34 +21,6 @@
 
   /** Random useful C macros.
   **/
-    /* Assert.  Good to capture.
-
-       TODO: determine which c3_assert calls can rather call c3_dessert, i.e. in
-       public releases, which calls to c3_assert should abort and which should
-       no-op? If the latter, is the assert useful inter-development to validate
-       conditions we might accidentally break or not useful at all?
-    */
-
-#     if defined(ASAN_ENABLED) && defined(__clang__)
-#       define c3_assert(x)                       \
-          do {                                    \
-            if (!(x)) {                           \
-              assert(x);                          \
-            }                                     \
-          } while(0)
-#     else
-#       define c3_assert(x)                       \
-          do {                                    \
-            if (!(x)) {                           \
-              fflush(stderr);                     \
-              fprintf(stderr, "\rAssertion '%s' " \
-                      "failed in %s:%d\r\n",      \
-                      #x, __FILE__, __LINE__);    \
-              assert(x);                          \
-            }                                     \
-          } while(0)
-#endif
-
     /* Dessert. Debug assert. If a debugger is attached, it will break in and
        execution can be allowed to proceed without aborting the process.
        Otherwise, the unhandled SIGTRAP will dump core.
@@ -68,10 +40,6 @@
 #else
   #define c3_dessert(x) ((void)(0))
 #endif
-
-    /* Stub.
-    */
-#     define c3_stub       c3_assert(!"stub")
 
     /* Size in words.
     */
@@ -137,7 +105,7 @@
         if ( 0 == rut ) {                                       \
           fprintf(stderr, "c3_malloc(%" PRIu64 ") failed\r\n",  \
                           (c3_d)s);                             \
-          c3_assert(!"memory lost");                            \
+          exit(ENOMEM);                                         \
         }                                                       \
         rut;})
 #     define c3_calloc(s) ({                                    \
@@ -145,7 +113,7 @@
         if ( 0 == rut ) {                                       \
           fprintf(stderr, "c3_calloc(%" PRIu64 ") failed\r\n",  \
                           (c3_d)s);                             \
-          c3_assert(!"memory lost");                            \
+          exit(ENOMEM);                                         \
         }                                                       \
         rut;})
 #     define c3_realloc(a, b) ({                                \
@@ -153,7 +121,7 @@
         if ( 0 == rut ) {                                       \
           fprintf(stderr, "c3_realloc(%" PRIu64 ") failed\r\n", \
                           (c3_d)b);                             \
-          c3_assert(!"memory lost");                            \
+          exit(ENOMEM);                                         \
         }                                                       \
         rut;})
 
--- a/pkg/noun/allocate.c
+++ b/pkg/noun/allocate.c
@@ -2,6 +2,7 @@
 
 #include "allocate.h"
 
+#include "error.h"
 #include "hashtable.h"
 #include "log.h"
 #include "manage.h"
@@ -111,7 +112,7 @@ _box_make(void* box_v, c3_w siz_w, c3_w use_w)
   u3a_box* box_u = box_v;
   c3_w*    box_w = box_v;
 
-  c3_assert(siz_w >= u3a_minimum);
+  u3_assert(siz_w >= u3a_minimum);
 
   box_u->siz_w = siz_w;
   box_w[siz_w - 1] = siz_w;     /* stor size at end of allocation as well */
@@ -132,8 +133,8 @@ _box_make(void* box_v, c3_w siz_w, c3_w use_w)
 static void
 _box_attach(u3a_box* box_u)
 {
-  c3_assert(box_u->siz_w >= (1 + c3_wiseof(u3a_fbox)));
-  c3_assert(0 != u3of(u3a_fbox, box_u));
+  u3_assert(box_u->siz_w >= (1 + c3_wiseof(u3a_fbox)));
+  u3_assert(0 != u3of(u3a_fbox, box_u));
 
 #if 0
   //  For debugging, fill the box with beef.
@@ -176,13 +177,15 @@ _box_detach(u3a_box* box_u)
 
   if ( nex_p ) {
     if ( u3to(u3a_fbox, nex_p)->pre_p != fre_p ) {
-      c3_assert(!"loom: corrupt");
+      fprintf(stderr, "loom: corrupt\r\n");
+      exit(ENOTRECOVERABLE);
     }
     u3to(u3a_fbox, nex_p)->pre_p = pre_p;
   }
   if ( pre_p ) {
     if( u3to(u3a_fbox, pre_p)->nex_p != fre_p ) {
-      c3_assert(!"loom: corrupt");
+      fprintf(stderr, "loom: corrupt\r\n");
+      exit(ENOTRECOVERABLE);
     }
     u3to(u3a_fbox, pre_p)->nex_p = nex_p;
   }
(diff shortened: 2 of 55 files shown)


ashelkovnykov (commented):
@mcevoypeter What was your decision process for replacing a statement with either `u3_assert` or `exit(...)`? In +95% of cases, it seems to be whether the statement is asserting a condition or asserting `0`.

The reason I ask is that these two situations can be considered interchangeable, depending on the context. For example:
```
assert(condition);
```
vs.
```
if (!condition) {
  // big oops
  assert(0);
}
```

Also: is it an important distinction for our use cases whether we `exit()` or `abort()` (a failed `u3_assert` resolves to `abort()`)?

pkg/noun/error.h
```diff
@@ -0,0 +1,21 @@
...
+  do {                                                                         \
+    if ( !(condition) ) {                                                      \
+      fprintf(stderr,                                                          \
+              "\rAssertion '%s' failed in %s:%d\r\n",                          \
+              #condition,                                                      \
+              __FILE__,                                                        \
+              __LINE__);                                                       \
+      u3m_bail(c3__oops);                                                      \
+    }                                                                          \
+  } while ( 0 )
```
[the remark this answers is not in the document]
mcevoypeter: Although I think the `if` statement can be used anywhere a single statement is used, using the `do` wrapper guarantees that the expansion is a single statement. Generally speaking, the `do` wrapper is a part of good macro hygiene.t

mcevoypeter:
> What was your decision process for replacing a statement with either u3_assert or exit(...)? In +95% of cases, it seems to be whether the statement is asserting a condition or asserting 0.

This was the main heuristic combined with knowledge about whether the error was truly fatal or not. `events.c`, for instance, replaces almost all assertions with `exit()`. You bring up a good point about the equivalency between `c3_assert(condition)` and `if (!condition) c3_assert(0)`, but the truth is I didn't consider that. There are so many call sites where we use `c3_assert` to handle errors, and individually considering each one for any longer than a few seconds would've been too much work. We have our work cut out for us in improving our error handling, and this PR is simply supposed to open up the conversation and take a small step in the right direction.

pkg/noun/error.h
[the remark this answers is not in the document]
ashelkovnykov: Thanks, your answer gave me enough to formulate the question better, so as to find a [more involved](https://stackoverflow.com/a/154138/15397371) answer with some example.

ashelkovnykov (approved):
As a replacement for `c3_assert` and the beginnings of a Vere-specific `errno` system, the PR looks good.

In the few places where I have domain specific knowledge, the cases looked correct.

Beyond that, I'll defer to @joemfb and @barter-simsum ("this will do").

@mcevoypeter Ping re: `abort()` vs `exit()` - is this a distinction that we should make right now? Will it affect any downstream users (e.g. hosting companies) who want to register `atexit` calls or expect `EXIT_SUCCESS` vs. `EXIT_FAILURE` vs. `SIGABRT`?

pkg/noun/allocate.c
```diff
@@ -2734,11 +2740,19 @@ u3a_loom_sane()
...
+        exit(ENOTRECOVERABLE);
+      }
+      if (pre_p && pre_u->nex_p != this_p) {
+        fprintf(stderr, "loom: corrupt\r\n");
+        exit(ENOTRECOVERABLE);
+      }
       if (!pre_p                /* this must be the head of a freelist */
-          && u3R->all.fre_p[_box_slot(this_u->box_u.siz_w)] != this_p)
-        c3_assert(!"loom: wack");
+          && u3R->all.fre_p[_box_slot(this_u->box_u.siz_w)] != this_p) {
```
barter-simsum, line 2752: please keep "wack" and don't replace with "corrupt". This makes it easy to differentiate typical corruption errors from those found when the freelist is traced

pkg/noun/error.h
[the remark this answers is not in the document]
barter-simsum: Well it's not exactly a single statement. It's a compound statement. But so is if {...}. The do {...} while(0), just the trick for defining a compound statement that _isn't_ conditional. So I think these should be removed.

barter-simsum (changes requested):
I feel that the `s/c3_assert/u3_assert/` in the macro calls is unnecessary and distracts from the semantic changes of this PR. (and it's harder to review because of this).

I'm not sure why these can't remain in c3. In general I find the u3/c3 distinction somewhat pointless. But basic utility function/macros are definitely not in the u3 "land of nouns". They should be defined in universally #includeable headers.

pkg/c3/defs.h
```diff
@@ -21,34 +21,6 @@
...
-            }                                     \
-          } while(0)
-#     else
-#       define c3_assert(x)                       \
-          do {                                    \
-            if (!(x)) {                           \
-              fflush(stderr);                     \
-              fprintf(stderr, "\rAssertion '%s' " \
-                      "failed in %s:%d\r\n",      \
-                      #x, __FILE__, __LINE__);    \
```
barter-simsum, line 46: __LINE__ can technically overflow :^) I've seen it strigified for this reason. But personally don't care about this. Probably only relevant for generated code

barter-simsum:
In general, I definitely support exit codes > 1. But I think we should define our own and not rely on errno.h for the following reasons:

1) It is useful to separate process error codes from the extraprocess error codes that are attached to errno when we call outside the process. 

e.g.
vere_status.h
```
typedef int VERESTAT;

#define VENOTRECO 1 // not recoverable. SOL
#define VENOTSUP 2 // not supported
#define VENOTIMPL 3  // not implemented

// ...
// other vere specific status codes
// ...
```
events.c
```
VERESTAT f() {...}
```

2) the defined values of errno are limited and particular to kernel statuses. There are a lot of definitions we don't need. e.g. `ENOTBLK` "Block device required.", `ENOSPC` "No space left on device", etc. Moreover, isn't it likely there will be a particular status we want to differentiate that just doesn't fit into the errno set? So errno is too much a superset in some cases and too much a subset in others.

3) Also not sure if this PR does it, but I'm strongly against using errno as a mechanism for interprocess control flow outside of checking its status following a system call and doing the appropriate thing. For instance:

- function A calls function B
- B opens a file and finds errno to have a value of EACCESS (uh oh)

If this is recoverable, B should directly return a vere status code indicating that. A shouldn't ever have to check errno to know what to do because A didn't make a syscall. If not recoverable, B should call an exit routine. This way errno shite is contained to functions making syscalls.

pkg/noun/urth.c
```diff
@@ -728,9 +733,11 @@ u3u_mmap_read(c3_c* cap_c, c3_c* pat_c, c3_d* out_d, c3_y** out_y)
     void* ptr_v;
 
     if ( MAP_FAILED == (ptr_v = mmap(0, len_d, PROT_READ, MAP_SHARED, fid_i, 0)) ) {
+      int err_i = errno;
       fprintf(stderr, "%s: mmap failed (%s): %s\r\n",
-                      cap_c, pat_c, strerror(errno));
+                      cap_c, pat_c, strerror(err_i));
       close(fid_i);
+      errno = err_i;
       return c3n;
```
barter-simsum, line 741: I understand why you're assigning errno to a temporary variable here to prevent its value being clobbered by the fprintf, but we don't care if errno gets clobbered because we've effectively translated its value to c3n (the only vere status code we have for now) and the caller shouldn't care what its value was.

-- right? or am I reading this wrong

barter-simsum:
RE. the equivalence of `assert(cond)` and `if(cond) exit(0)`, @ashelkovnykov raised.

The only real difference as far as we care is that by raising SIGABRT we'll create a coredump. We definitely want to keep that behavior.

barter-simsum:
I just want to add that by defining your own status codes, you have far more flexibility. For instance, errno is used to give higher fidelity to an error, but what if we want more information on a success. You could use a single bit in a status code to do this (probably the msb since that has the property of making failures negative and success positive). Generally I would like to see all routines in vere that return a c3o (failed or not) to start returning a VERESTATUS.

pkg/noun/error.h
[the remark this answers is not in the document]
barter-simsum: NVM the above comment. Realize the inadequacies of defining a compound statement in a macro with `if (cond) stmt` in the case of the macro being expanded inside the if of an if/else without braces

pkg/c3/defs.h
mcevoypeter: If we have over two billion lines (looks like `__LINE__` is `int` according to https://github.com/tloncorp/ylem/issues/822) in a single file, then we have bigger problems.

pkg/noun/urth.c
mcevoypeter: Yeah, good catch.

mcevoypeter:
I'm not claiming this is the final form of Vere error handling. Only a step in the right direction. I agree we want custom error codes, but they're outside the scope of this PR. This is a step toward a goal, not the achievement of the goal itself.

pkg/c3/defs.h
barter-simsum: lol agree this is absolutely pedantic

barter-simsum:
@mcevoypeter the fact that we won't dump core when previously we would have is my biggest concern with this pr https://github.com/urbit/vere/pull/343#issuecomment-1513656975

barter-simsum:
Caring about exit codes implies that there is code outside of vere that is making a decision based on that code's value - i.e. it's an API.

For this reason, I'm against using the errno.h status codes temporarily because we'll break the api when we define our own.

I'd propose something like the following:

```
// true if status x is a success, otherwise a failure
#define VERE_SUCCESS(x) (! (x) & 0x80000000)

typedef uint32_t VSTATUS;

// define replacements for:
// ENOMEM
// ENOTRECOVERABLE
// ENOTSUP
// EFAULT
// EINVAL
// ECANCELED
// EPROTO
// EPERM
// EEXIST


#define V_FAULT ((VSTATUS)0x1L))
#define V_NOMEM ((VSTATUS) 0x2L))
// ...

```

mcevoypeter:
After talking to @barter-simsum offline, I'm closing this because it's undesirable to replace `abort()` with `exit()` for fatal errors. I'll open another PR that moves `c3_assert()` to `pkg/noun` as `u3_assert()`.

PR #346, opened 2023-04-12
epoch system

### LINKED ISSUE #313: `pier`: epoch system
To ameliorate jet mismatching issues (especially during event log replay), an "epoch system" should be created. The epoch system is a new `<pier>/.urb/log` format which enables event log replays to correctly match binary versions with particular subsets of events from the pier's event log.

Related: 
* urbit/urbit#5701
* [Epoch system on the roadmap](https://roadmap.urbit.org/project/epoch-system)
* [Epoch system design WIP](https://gist.github.com/matthew-levan/d4a438f65d5c405de5f76405e8c9ff4b)

matthew-levan:
Resolves #313.


pkg/vere/disk.c
```diff
@@ -882,10 +884,35 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
     //
-    //  per the LMDB docs:
-    //  "[..] on 64-bit there is no penalty for making this huge (say 1TB)."
+
+    //  migrate to the correct disk format
+    //
+    u3_disk_migrate(log_u);
+
+    //  find the latest epoch directory
+    //  XX separate function?
```
belisarius222, line 895: yeah I think this deserves its own function, even if it's only called once -- it's conceptually important enough to live on its own, and I'd say it's likely to be used by other tools too.

pkg/vere/disk.c
```diff
@@ -924,24 +923,22 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
-      return 0;
+    if ( 0 == lat_d ) {
+      c3_d fir_d;
+      if ( c3n == u3_lmdb_gulf(log_u->mdb_u, &fir_d, &log_u->dun_d) ) {
+        log_u->dun_d = 0;
+      }
+      log_u->sen_d = log_u->dun_d;
+    } else {
+      log_u->dun_d = lat_d;
+      log_u->sen_d = log_u->dun_d;
```
joemfb, line 935: `sen_d` and `dun_d` are for "commits in progress" vs "commits completed" -- you should just add a new member to the `u3_disk` struct for "first event in epoch".

pkg/vere/disk.c
```diff
@@ -1020,6 +1027,14 @@ c3_o u3_disk_epoc_init(u3_disk* log_u) {
...
   }
 
+  // write the metadata to the database
+  if ( c3y == eps_o ) {
+    if ( c3n == u3_disk_save_meta(log_u->mdb_u, who_d, fak_o, lif_w) ) {
+      fprintf(stderr, "disk: failed to save metadata\r\n");
+      exit(1);
+    }
+  }
+
```
joemfb, line 1037: this is fine for now, but we're going to get the metadata out of arvo going forward. we can save that change for the end, or for a future version of the epoch format.

pkg/vere/disk.c
matthew-levan: Thanks, but I don't understand this yet. Why in the old code do we set `sen_d` to `dun_d`? Also, I think the incorrectness of my code here is why I can't boot rolled ships yet.

matthew-levan:
Midnight update: I can now boot a fresh fakezod with the epoch format, `roll` it, and start it up again. Events appear to be processing normally in the `dojo` from that state, but only the snapshot (not the database) seems to be getting updated. Thus, when shutting down *again* and trying to reboot, I get a:
```
...
boot: installed 652 jets
pier: eve_d 146, dun_d 133
Assertion 'god_u->eve_d <= log_u->dun_d' failed in pkg/vere/pier.c:1401
Assertion failed: god_u->eve_d <= log_u->dun_d (pkg/vere/pier.c: _pier_on_lord_live: 1401)
```

Thoughts welcome, just taking notes here for my future self.

pkg/vere/disk.c
joemfb: Yeah, they're both event numbers. "commits in flight" would be `sen_d - dun_d`.

matthew-levan:
That makes sense, thank you sir.

matthew-levan:
Another update: got `roll` working nicely. `chop` is next on the.. chopping block. `prep` too. Finally will come the migration code. Stay tuned!

jalehman:
Thanks for the update! This is exciting.

—

~wolref-podlex
https://urbit.org ( https://urbit.org/ )

matthew-levan:
`chop` *was* easy. It's probably not perfect, but it works... Onto `prep`, then onto the migration code.

matthew-levan:
Implemented a very minimal migration routine for existing piers-- still need to validate there's a current snapshot, rollover to a new epoch, and delete the backup snapshot (in that order). 

`_cw_roll` is currently just a private function in `main.c`-- maybe it becomes `u3_disk_epoc_roll` in `disk.c` so I can use it in `disk.c` too. Hm...

pkg/vere/disk.c
```diff
@@ -867,66 +874,370 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+      fprintf(stderr, "disk: failed to get first/last event numbers\r\n");
       return 0;
     }
 
+    //  initialize dun_d/sen_d values
+    if ( 0 == las_d ) {               //  fresh epoch (no events in lmdb yet)
+      if ( 0 == lat_d ) {             //  first epoch
+        log_u->dun_d = 0;
+      } else {                        //  not first epoch
+        log_u->dun_d = lat_d - 1;     //  set dun_d to last event in prev epoch
```
joemfb, line 969: This looks wrong. `lat_d` is the previous epoch number, right? What we need is the number of the last event in the previous epoch.

matthew-levan: Apologies for lack of clarity. `lat_d` is the latest epoch number *after* creating a new one in the migrate function. Therefore it will be an empty epoch and named N+1 (where N is the last event of the previous epoch) or 0 (if it's the first epoch). Hope that makes sense. The code does indeed work.

matthew-levan: Just realized this isn't reflecting the design though... According to our call this past week, we want the epoch numbers to be named N (not N+1) where N is the last event of the previous epoch. Will fix.

pkg/vere/disk.c
```diff
@@ -22,6 +23,12 @@ struct _cd_save {
   struct _u3_disk* log_u;
 };
 
+/* DISK FORMAT
+ */
+
+#define U3D_VER1   1
+#define U3D_VERLAT U3L_VER1
+
```
barter-simsum, line 31: these should be defined in version.h. You're already importing it anyway.

matthew-levan:
Sounds good.

pkg/vere/disk.c
joemfb: Arguably, the pier/log format is a higher-level concern, and shouldn't be part of the core noun package. Doesn't matter match, and will be easy to change later.

matthew-levan:
Ok I'll just revert that commit where I followed David's suggestion.

pkg/vere/disk.c
barter-simsum: version.h is in /noun "just cuz." I don't consider it to be part of any package.

@matthew-levan U3D_VERLAT evaluates to U3L_VER1 which isn't defined. It should be U3D_VER1 instead

pkg/vere/main.c
```diff
@@ -2128,123 +2131,192 @@ _cw_chop(c3_i argc, c3_c* argv[])
...
+  // note: this includes patch applications (if any)
   u3m_boot(u3_Host.dir_c, (size_t)1 << u3_Host.ops_u.lom_y);
 
   // check if there's a *current* snapshot
-  if ( old_u->dun_d != u3A->eve_d ) {
+  if ( log_u->dun_d != u3A->eve_d ) {
     fprintf(stderr, "chop: error: snapshot is out of date, please "
                     "start/shutdown your pier gracefully first\r\n");
-    fprintf(stderr, "chop: eve_d: %" PRIu64 ", dun_d: %" PRIu64 "\r\n", u3A->eve_d, old_u->dun_d);
+    fprintf(stderr, "chop: eve_d: %" PRIu64 ", dun_d: %" PRIu64 "\r\n", u3A->eve_d, log_u->dun_d);
```
barter-simsum, line 2143: NIT: there are a lot of these so it'd probably be easier to fix up at the end. `eve_d`, `dun_d`, etc. are /aliases/ to fixed width 64 bit integers. You shouldn't use the 64 print specifier directly. I added corresponding print specifiers to types.h to use with c3 types. These are convenient because you don't have to think about integer width at all.

Use PRIc3_d for these since they're a c3_d type.

matthew-levan: Great thank you. Done.

pkg/noun/events.c
```diff
@@ -1379,7 +1380,9 @@ u3e_save(u3_post low_p, u3_post hig_p)
     }
   }
 
-  u3e_backup(c3n);
+  c3_c bhk_c[8193];
+  snprintf(bhk_c, sizeof(bhk_c), "%s/.urb/bhk", u3P.dir_c);
+  u3e_backup(bhk_c, c3n);
```
joemfb, line 1385: This heuristic backup should be removed, the epoch system itself is a much better solution.

pkg/noun/manage.c
```diff
@@ -1753,7 +1753,9 @@ _cm_limits(void)
 c3_o
 u3m_backup(c3_o ovw_o)
 {
-  return u3e_backup(ovw_o);
+  c3_c bhk_c[2048];
+  snprintf(bhk_c, 2048, "%s/.urb/bhk", u3P.dir_c);
+  return u3e_backup(bhk_c, ovw_o);
 }
```
joemfb, line 1759: This function should also be removed, the only call site was for `chop` which no longer uses it.

pkg/noun/noun.h
```diff
@@ -15,6 +15,7 @@
 #include "types.h"
 #include "vortex.h"
 #include "zave.h"
+#include "events.h"
```
joemfb, line 18: `events.h` is intentionally excluded here, it's an internal module and should only be imported/called explicitly where needed.

pkg/noun/urth.c
```diff
@@ -8,6 +8,7 @@
 #include <sys/stat.h>
 
 #include "allocate.h"
+#include "events.h"
```
joemfb, line 11: it should also be removed here.

pkg/vere/disk.c
```diff
@@ -2,6 +2,7 @@
 
 #include "noun.h"
 #include "vere.h"
+#include "version.h"
```
joemfb, line 5: It seems like `version.h` should be automatically included in `noun.h`

pkg/vere/db/lmdb.c
```diff
@@ -38,11 +38,20 @@ intmax_t mdb_get_filesize(mdb_filehandle_t han_u);
 /* u3_lmdb_init(): open lmdb at [pax_c], mmap up to [siz_i].
 */
 MDB_env*
-u3_lmdb_init(const c3_c* pax_c, size_t siz_i)
+u3_lmdb_init(const c3_c* pax_c)
```
joemfb, line 41: Lets restore the size argument here, we really should be able to set this parameter at runtime / via commandline flag.

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
 #if defined(DISK_TRACE_JAM) || defined(DISK_TRACE_CUE)
   u3t_trace_open(pax_c);
 #endif
 
   return log_u;
 }
+
+/* u3_disk_epoc_good: check for valid epoch.
+*/
+c3_o u3_disk_epoc_good(u3_disk* log_u, c3_d epo_d) {
```
joemfb, line 961: function signature style: return type should be on its own line.

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+    fprintf(stderr, "disk: epoch 0i%" PRIc3_d "/epoc.txt is not writable\r\n", epo_d);
+    ret_o = c3n;
+  }
+  if ( c3n == biv_o ) {
+    fprintf(stderr, "disk: epoch 0i%" PRIc3_d "/vere.txt is not writable\r\n", epo_d);
+    ret_o = c3n;
+  }
+
+  return ret_o;
+}
```
joemfb, line 1038: This function makes me nervous. The validation that it performs seems sane, but it's then used to decide whether we should overwrite this epoc or not (in `u3_disk_epoc_init()`). It doesn't seem quite constrained enough for that purpose. I think we can only do that if there are no events in the log for the epoc under consideration.

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+ 
+/* u3_disk_epoc_init: create new epoch.
+*/
+c3_o u3_disk_epoc_init(u3_disk* log_u) {
```
joemfb, line 1042: function signature style

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+fail:
+  c3_unlink(epv_c);
+  c3_unlink(biv_c);
+  c3_rmdir(epo_c);
+  return c3n;
+}
+
+/* u3_disk_epoc_kill: delete an epoch.
+*/
+c3_o u3_disk_epoc_kill(u3_disk* log_u, c3_d epo_d) {
```
joemfb, line 1140: function signature style

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+  //  cleanup
+  u3_dire_free(dir_u);
+
+  //  success
+  return c3y;
+}
+
+/* u3_disk_epoc_last: get latest epoch number.
+*/
+c3_o u3_disk_epoc_last(u3_disk* log_u, c3_d* lat_d) {
```
joemfb, line 1173: function signature style

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+  return ret_o;
+}
+
+/* u3_disk_epoc_vere: get binary version from epoch.
+*/
+c3_o
+u3_disk_epoc_vere(u3_disk* log_u, c3_d epo_d, c3_c* ver_w) {
+  //  XX  check implementation (similar code is in king.c, _king_get_pace)
+  //      should probably write a generic function to read a string from 
+  //      a file into malloc'd memory, with an optional whitespace trimming
```
joemfb, line 1198: Bumping this comment, as this function signature is unsafe (`fil_u` is read into `ver_w`, regardless of length).

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+    fprintf(stderr, "disk: failed to read vere.txt in epoch 0i%" PRIc3_d 
+                    "\r\n", epo_d);
+    return c3n;
+  }
+  return c3y;
+}
+
+/* u3_disk_migrate: migrates disk format.
+ */
+c3_o u3_disk_migrate(u3_disk* log_u)
```
joemfb, line 1220: function signature style

pkg/vere/disk.c
```diff
@@ -867,66 +868,494 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+      fprintf(stderr, "disk: migrate: failed to unlink data.mdb\r\n");
+      return c3n;  //  migration succeeds only if we can unlink data.mdb
+    }
+
+    //  success
+    fprintf(stderr, "disk: migrated disk to v%d format\r\n", U3D_VER1);
+  }
+
+  return c3y;
+}
```
joemfb, line 1361: The newline at the end of this file should be restored.

joemfb (changes requested):
I've noted some style nits, and a concern about overwriting incomplete/invalid epochs while rolling-over. Otherwise, this functionality looks good. Still need to do the final round of crash-recovery / idempotence testing, then it should be good to go.

joemfb (changes requested):
I've noted some style nits, and a concern about overwriting incomplete/invalid epochs while rolling-over. Otherwise, this functionality looks good. Still need to do the final round of crash-recovery / idempotence testing, then it should be good to go.

pkg/vere/disk.c
matthew-levan: Good call-- I added a block that checks whether or not there are any events in the epoch's `data.mdb` file.

barter-simsum:
`u3_disk_epoc_init` and several other functions in disk.c should have brace on newline rather than same line

pkg/vere/disk.c
```diff
@@ -867,66 +868,514 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
   return log_u;
 }
+
+/* u3_disk_epoc_good: check for valid epoch.
+*/
+c3_o
+u3_disk_epoc_good(u3_disk* log_u, c3_d epo_d) {
+  /*  a "valid" epoch is currently defined as a writable folder in
+   *  the <pier>/.urb/log directory named with a @ui and contains:
+   *  - a writable data.mdb file which can be opened by lmdb (add db checking?)
```
barter-simsum, line 965: example of the brace nit I mentioned above, though it looks like this applies to all new functions in disk.c

pkg/vere/disk.c
```diff
@@ -867,66 +868,522 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+  //  initialize first epoch "0i0"
+  if ( c3n == u3_disk_epoc_init(log_u) ) {
+    fprintf(stderr, "disk: migrate: failed to initialize first epoch\r\n");
+    return c3n;
+  }
+
+  //  create hard links to data.mdb and lock.mdb in 0i0/
+  c3_c dut_c[8193], luk_c[8193];  //  old paths
+  snprintf(dut_c, sizeof(dut_c), "%s/data.mdb", log_u->com_u->pax_c);
+  snprintf(luk_c, sizeof(luk_c), "%s/lock.mdb", log_u->com_u->pax_c);
```
barter-simsum, line 1334: You actually have unnecessary code duplication here. In the function-level scope at the top, you declare `dat_c` and `lok_c` which contain the same strings as `dut_c` and `luk_c` here. So you should just declare dut and luk at the top and use them here

barter-simsum: Uh, so slight change, because it looks like you collapsed the condition

barter-simsum: Below, you snprintf the new dat and lock locations into `dat_c` and `luk_c` already declared. _Instead_, I would just at the top of `u3_disk_migrate`:
```C
c3_c dut_c[8193], luk_c[8193];  //  old paths
snprintf(dut_c, sizeof(dut_c), "%s/data.mdb", log_u->com_u->pax_c);
snprintf(luk_c, sizeof(luk_c), "%s/lock.mdb", log_u->com_u->pax_c);

// do the stuff with access(dut_c), access(luk_c) etc

// now here, declare dat_c and lok_c
c3_c epo_c[8193], dat_c[8193], lok_c[8193];  //  new paths
snprintf(epo_c, sizeof(epo_c), "%s/0i0", log_u->com_u->pax_c);
snprintf(dat_c, sizeof(dat_c), "%s/data.mdb", epo_c);
snprintf(lok_c, sizeof(lok_c), "%s/lock.mdb", epo_c);

// ...
```

barter-simsum: i.e. at the beginning of this function, dat_c and lok_c store the _old_ paths and later store the new paths when dut_c and luk_c store the old paths. Just always store the old paths in dut_c and luk_c. Swapping them is confusing/unnecessary

matthew-levan:
Yes, will do. Thank you, good eye.

barter-simsum:
pushed minor change in f7722622606a6a2354296bac2832e70bcab3e598. dut_o is pointless, just exit the migration early if data.mdb can't be accessed

pkg/noun/noun.h
matthew-levan: Ok. I was using it to include the `u3e_backup` function. I've moved the `#include` directive to `disk.c` instead-- better?

matthew-levan:
Successfully ran a migration on `~mastyr-bottec` using local-networking mode:
```
matthew@domus:~/src/urbit/vere$ ./urbit --local ~/ships/planets/mastyr-bottec/mastyr-bottec/
~
urbit 2.9-c045119fd3
boot: home is /home/matthew/ships/planets/mastyr-bottec/mastyr-bottec
loom: mapped 2048MB
boot: protected loom
live: mapped: GB/1.345.880.064
live: loaded: KB/16.384
boot: installed 657 jets
loom: mapped 2048MB
lite: arvo formula 2a2274c9
lite: core 4bb376f0
lite: final state 4bb376f0
loom: image backup complete
disk: migrated disk to v1 format
disk: loaded epoch 0i219237052
loom: mapped 2048MB
boot: protected loom
live: mapped: GB/1.345.880.064
live: loaded: KB/16.384
boot: installed 657 jets
vere: checking version compatibility
%unlagging
ames: live on 51723 (localhost only)
conn: listening on /home/matthew/ships/planets/mastyr-bottec/mastyr-bottec/.urb/conn.sock
http: web interface live on http://localhost:8080
http: loopback live on http://localhost:12321
pier (219237063): live
~mastyr-bottec:dojo> 
```

matthew-levan:
Closed in favor of #459, which implements the same and also integrates with `urbit play` a la #420.

PR #356, opened 2023-04-25
Replace `c3_assert()` with `u3_assert()`

### LINKED ISSUE #355: Replace `c3_assert()` with `u3_assert()`
The build system change that occurred alongside Vere's extraction from https://github.com/urbit/urbit required removal of an indirect call to `u3m_bail(c3__oops)` in `c3_assert()` because of a circular dependency between `pkg/c3` and `pkg/noun`. This `u3m_bail(c3__oops)` call be reinstated by renaming `c3_assert()` as `u3_assert()` and moving it to `pkg/noun` so `u3_assert()` can call `u3m_bail(c3__oops)` without issue.

mcevoypeter:
Resolves #355.


barter-simsum (approved):
LGTM

Issue #235, opened 2023-02-16
blit read: Bad address (during meld)


timlucmiptev:
Got the following when melding a ship:

```
loom: mapped 2048MB
loom: image (north) blit read: Bad address
Assertion '0' failed in pkg/noun/events.c:864
Assertion failed: 0 (pkg/noun/events.c: _ce_image_blit: 864)
```

The same error occurs when trying to `.run next`.

Binary: 1.19
OS: Linux

jfranklin9000:
Try adding `--loom 32` when melding, see https://github.com/urbit/vere/issues/233.

timlucmiptev:
OK, good news is that that worked. 

Bad news is that it took up to 15GB of mem (RAM + swap), and also that error message should be more informative.

PR #357, opened 2023-04-25
u3: check that snapshot fits in *current* loom size 

joemfb:
... and not just the compile-time maximum.

Helps with #235, urbit/urbit#6508.


barter-simsum (approved):
"this will do"

PR #362, opened 2023-05-01
u3: optimizes u3i_edit (nock opcode 10)

joemfb:
This PR ports urbit/urbit#6004. It is not urgent.

Benchmarking this against the old implementation would be worthwhile. It might also be worth inlining/decomposing `u3a_is_mutable()`; this is (currently) its only call-site, and the road pointer and seniority predicate cannot change throughout this loop.

### DIFF
--- a/pkg/noun/imprison.c
+++ b/pkg/noun/imprison.c
@@ -626,12 +626,15 @@ u3i_edit(u3_noun big, u3_noun axe, u3_noun som)
     case 1: break;
 
     default: {
-      c3_w dep_w = u3r_met(0, u3x_atom(axe)) - 2;
+      c3_w        dep_w = u3r_met(0, u3x_atom(axe)) - 2;
+      const c3_w* axe_w = ( c3y == u3a_is_cat(axe) )
+                        ? &axe
+                        : ((u3a_atom*)u3a_to_ptr(axe))->buf_w;
 
       do {
         u3a_cell*  big_u = u3a_to_ptr(big);
         u3_noun*     old = (u3_noun*)&(big_u->hed);
-        const c3_y bit_y = u3r_bit(dep_w, axe);
+        const c3_y bit_y = 1 & (axe_w[dep_w >> 5] >> (dep_w & 31));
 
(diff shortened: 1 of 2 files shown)


ashelkovnykov:
This PR came up today during a sync between @joemfb and I. I skimmed it to make sure I understood the motivation and how it spiritually aligns w.r.t. similar noun walks in Ares, but I didn't verify the logic carefully enough to really "review" it.

joemfb:
I've written a benchmark for this code, and run the various versions of this operation through it. The benchmark constructs a list of `1`s, and then edits the list at the axis of each item, replacing the `1` with `2`. This is an absurd operation, and no real-world code we ever run would look like it. But it's the best-case scenario for certain optimizations here, and gives us an opportunity to compare them:

| opcode 10 variation                                             | 1k list | 10k list |
| --------------------------------------------------------------- | ------- | -------- |
| develop (a244cc57a5a60467fb51124770dfdc87a192c512)              | 19ms    | 2.41s    |
| ^ without mutation (a6debdb0940334ce313ae73d8eb177e6a415a128)   | 26ms    | 3.218s   |
| loop (d6988e6f931cfce669ac5542b711f4f2f6937d6f)                 | 1ms     | 147ms    |
| reduce branches (fa84fd88e80829cd0ae4da8b38eb6bd06f36e517)      | 1ms     | 170ms    |
| inline axis bit-math (e182de8ddd762cf822e59b7b8bd07381ec756c82) | 1ms     | 159ms    |
| ^ without mutation                                              | 6ms     | 667ms    |
| inline axis without branch reduction                            | 1ms     | 130ms    |

Notably, the mutation optimization (we can edit a cell in place if it is on our road and has a refcount of 1) is only a ~30% improvement with the old implementation, but is a 5-6x improvement in the new.

I suspect that the branch reduction being slower is an artifact of this benchmark -- we're almost always editing the tail of a cell (lists associate to the right), making branch prediction very effective. So I've left it in for now, but I'm open to being talked out of that.

ashelkovnykov (approved):
Re-reviewed later, and then just now w/ Joe

PR #369, opened 2023-05-04
http: implement request cancellation

joemfb:
This PR implements the `%cancel` effect from %eyre in the http i/o driver. It also properly handles request timeouts when we've already started sending a response (`u3_rsat_ripe`), and corrects handling of invalid effects.

These changes still need more testing. Advice on how to consistently induce %eyre to send %cancel effects would be appreciated.

/cc @pkova

### DIFF
--- a/pkg/vere/io/http.c
+++ b/pkg/vere/io/http.c
@@ -495,7 +495,11 @@ _http_req_kill(u3_hreq* req_u)
 typedef struct _u3_hgen {
   h2o_generator_t neg_u;             // response callbacks
   c3_o            red;               // ready to send
-  c3_o            dun;               // done sending
+  enum {                             //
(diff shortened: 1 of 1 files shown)


pkova:
Open any landscape app that uses `js-http-api` to subscribe to anything, find the get request in chrome dev tools (it's the one with channel-id and 200 response), and double click it. Duplicate get request on a channel causes `%eyre` cancel the old one instantly.

image

joemfb:
Thanks, that works. When I do it for an SSE in talk, it immediately detects the cancelation and sends another GET, stealing the SSE right back.

arthyn:
@joemfb after meeting with @pkova and running this, it seems to clear up issues seen with reconnects making PUT requests take +20s. Will keep running to make sure.

PR #371, opened 2023-05-05
docker: fix all deployments tagged as 'once'

ashelkovnykov:
Resolves [missing tags on DockerHub](https://hub.docker.com/r/tloncorp/vere/tags) (no issue exists for this, but multiple people have mentioned noticing it). This is also responsible for the missing `2.2` tag.


barter-simsum (approved):
lftm

mopfel-winrux (commented):
LGTM

PR #373, opened 2023-05-08
fix u3m_water obo

### LINKED ISSUE #372: u3m_water obo
obobobobobobobo

barter-simsum:
mostly a straight copy of urbit-v1.14 with the exception of adjustments for pointer compression

Resolves #372


pkg/noun/manage.c
```diff
@@ -1027,12 +1027,31 @@ u3m_flog(c3_w gof_w)
 /* u3m_water(): produce watermarks.
 */
 void
-u3m_water(c3_w* low_w, c3_w* hig_w)
+u3m_water(c3_w* low_p, c3_w* hig_p)
```
ashelkovnykov, line 1030: Why the switch from `uint32_t` to `uintptr_t`?

pkg/noun/manage.c
```diff
@@ -1027,12 +1027,31 @@ u3m_flog(c3_w gof_w)
...
+  //
+  if ( !u3R ) {
+    *low_p = 0;
+    *hig_p = u3C.wor_i - 1;
+  }
+  //  in a north road, hat points to the end of the heap + 1 dword,
+  //  while cap points to the top of the stack
+  //
+  else if ( c3y == u3a_is_north(u3R) ) {
+    *low_p = u3R->hat_p - u3C.walign_w;
```
ashelkovnykov, line 1045: Shouldn't this be `+`, not `-` `u3C.walign_w`?

pkg/noun/manage.c
barter-simsum: These are posts aka c3_w

pkg/noun/manage.c
```diff
@@ -1027,12 +1027,31 @@ u3m_flog(c3_w gof_w)
 /* u3m_water(): produce watermarks.
 */
 void
-u3m_water(c3_w* low_w, c3_w* hig_w)
+u3m_water(c3_post* low_p, c3_post* hig_p)
```
ashelkovnykov, line 1030: Should be `u3_post`

ashelkovnykov (changes requested):

pkg/noun/manage.c
ashelkovnykov: Woops, my mistake - it was the `_p` suffix that caused me to misread. Now that these are going to be `u3_post`, I believe they shouldn't have a suffix at all.

barter-simsum: no, posts _do_ have a suffix e.g. hat_p

pkg/noun/manage.c
barter-simsum: I need to think about this a little more and look at the calling code

pkg/noun/manage.c
barter-simsum: oh duh, fixed.

joemfb (changes requested):
A straight port from v1.14 is not viable here -- it changed the meaning of this function and all call-sites at the same time. The old (current) meaning is "number of words in-use, north and south". The new meaning is "innermost word-offsets (posts) in-use". I think that latter definition is more useful, but the change is more disruptive.

The current implementation is consistent, but wrong for snapshots (which is the only call-site). The low water mark is calculated with `u3a_heap()`, which  is `hat_p - rut_p` on the home road (north). That does not give you the number of words in use for north.bin, since `rut_p` is greater than the loombase (see `_pave_home()`).

The smallest change that corrects the north.bin length calculation would be `*low_w = u3R->hat_p - 1;`. `hat_p` still points at the lowest word of the contiguous free space, one less than that is the number of words in use in the north segment. It could also be expressed as `u3a_heap(u3R) + u3C.walign_w` (assuming details of `rut_p` initialization), but that seems like false generality.

barter-simsum:
@joemfb missed the change in semantics -- had just assumed u3m_water was always being called with posts

pkg/noun/manage.c
ashelkovnykov: That's kinda weird. I get it, since it's actually shadowing a `c3` type, but it's confusing because it's the only `u3` structure that does this (that I can think of), and it clobbers a `c3` suffix. This should probably be noted in `u3.md` - there's even a section on `u3_post` that doesn't mention this.

barter-simsum:
@joemfb Maybe I'm missing something. why not simply `*low_w = u3R->hat_p`? This is equal to `u3a_heap(u3R) u3C.walign_w`. i.e. why are you subtracting 1 in `*low_w = u3R->hat_p - 1`?.

joemfb (approved):
@barter-simsum you're right, `hat_p - 1` would be the new,  u3_post semantics, not to the current word-length semantics.

this will do

pkg/noun/manage.c
joemfb: It is kind of weird. `c3_p` and `c3_ps` are unused, I think, and should probably just be removed. `_p` is also incorrectly used for `void*` in a few places -- it should always be `u3_post`.

barter-simsum:
@ashelkovnykov can you reset your vote? Can't merge I think until you do

ashelkovnykov (approved):
That's weird, that GH wouldn't throw away my review - or at least surface the option to do so

PR #375, opened 2023-05-10
ames: minimum viable decrement underflow hotfix

Fang-:
We were incrementing the queue size counter conditionally, only for packets we want to forward. But for all packets, we were issuing a scry, and the callback would unconditionally decrement the queue size counter, leading to underflows. That, in turn, would result in vere thinking the queue was full when it wasn't.

Here, we make the least invasive change that resolves that particular bug. There may or may not still be other things wrong with the logic/factoring around the packet queue and lane scries.

f9e8542f9 changed the factoring here, making normal packet sending and forwarding take the same code path. ~~But also, even before that change, we were always sending the head of the packet queue to whatever lane scry result we got, which feels like it might not be right, _especially given that we don't even put the packet in queue when not forwarding_.~~ (edit: i think that was an incorrect reading of the code, we pass a pointer to the specific packet, not the whole queue.) I suspect we just don't want to scry for the lane in the normal sending codepath. We didn't do that before recent changes either. (edit: as currently written, we keep packets we want to send in memory until the scry gets back to us, without tracking them anywhere.) But I don't know the full context of the recent changes, so I'm only solving the specific symptom/bug here.

Since this bug prevents forwarding from working reliable under certain circumstances, this should go out as a hotfix.

### DIFF
--- a/pkg/vere/io/ames.c
+++ b/pkg/vere/io/ames.c
@@ -1393,8 +1393,6 @@ _ames_lane_scry_cb(void* vod_p, u3_noun nun)
   u3_ames* sam_u = pac_u->sam_u;
   u3_weak    las = u3r_at(7, nun);
 
-  sam_u->sat_u.foq_d--;
-
   //  if scry fails, remember we can't scry, and just inject the packet
   //
   if ( u3_none == las ) {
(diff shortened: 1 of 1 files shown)


joemfb (approved):
this will do

Fang-:
Managed to get a SIGINFO print on ~zod suffering from the bug, fully confirming the description given above:

```
pier: work
  effects: released=2331439778
  drivers:
    ...
    ames: live=&, queue=0
      config:
        filtering: &
         can send: &
         can scry: &
        caches:
          cached lanes: 7726, 1061432 B
          cached meows: 69, 80824 B
      counters:
                 dropped: 0
        forwards dropped: 1129691
        forwards pending: 18446744073709551607
               forwarded: 132960
                     ...
         lane scry fails: 0
            cached lanes: 7726
```

PR #386, opened 2023-05-11
ames.c: fix memory leak in _ames_try_send()

philipcmonk:
Resolves memory leak on galaxies, where we leaked every packet that we're forwarding if we already know the lane.


belisarius222 (approved):

philipcmonk:
Testing on ~zod, previously it was adding several MB per minute, now it's been stable at 28MB for 10 minutes.  So this seems to be working.

PR #391, opened 2023-05-12
http: fallback to Eyre on failed cache scry

lukechampine:
Haven't tested this, pushing early to get feedback on the approach. Basically, we put `[~ ~]` scries into the cache just like any other noun, and if `_http_cache_respond` sees `[~ ~]`, it falls back to `_http_req_dispatch`.

Resolves https://github.com/urbit/urbit/issues/6582

### DIFF
--- a/pkg/vere/io/http.c
+++ b/pkg/vere/io/http.c
@@ -651,14 +651,32 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
 */
 static void
 _http_cache_respond(u3_hreq* req_u, u3_noun nun) {
-  // XX check auth
-  u3_noun auth, response_header, data;
-  u3x_qual(u3k(u3t(u3t(nun))), &auth, 0, &response_header, &data);
-  u3_noun status, headers;
-  u3x_cell(response_header, &status, &headers);
+  h2o_req_t* rec_u = req_u->rec_u;
+  u3_httd* htd_u = req_u->hon_u->htp_u->htd_u;
 
-  req_u->sat_e = u3_rsat_plan;
-  _http_start_respond(req_u, u3k(status), u3k(headers), u3k(data), c3y);
+  if ( u3_nul == nun ) {
+    h2o_send_error_404(rec_u, "Not Found", "not found", 0);
+  }
+  else if ( u3_none == u3r_at(7, nun) ) {
+    h2o_send_error_500(rec_u, "Internal Server Error", "scry failed", 0);
+  }
+  else {
+    u3_noun auth, response_header, data;
+    u3x_qual(u3k(u3t(u3t(nun))), &auth, 0, &response_header, &data);
+    u3_noun status, headers;
+    u3x_cell(response_header, &status, &headers);
+
+    // check auth
+    if ( (c3y == auth)
+      && (c3n == _http_req_is_auth(&htd_u->fig_u, rec_u)) )
+    {
+      h2o_send_error_403(rec_u, "Unauthorized", "unauthorized", 0);
+    }
+    else {
+      req_u->sat_e = u3_rsat_plan;
+      _http_start_respond(req_u, u3k(status), u3k(headers), u3k(data), c3y);
+    }
+  }
   u3z(nun);
 }
 


lukechampine:
Updated: now, instead of falling back to Eyre, vere returns a 404, [matching Eyre's (current) behavior](https://github.com/urbit/urbit/blob/65fd1cc179693bce297bbc5c73cf276c874d1a74/pkg/arvo/sys/vane/eyre.hoon#L829). We discussed modifying this behavior to support truly "unbinding" a URL, as currently it's not possible to make a URL dynamic again after it's been bound in the cache. We also discussed changing Eyre's %cache scry API to properly distinguish between unbound and never-bound paths. Both of these would have to happen in a future kelvin.

Also incorporated the auth checking from https://github.com/urbit/vere/pull/199, so authorization actually works on cache entries now.

pkg/vere/io/http.c
```diff
@@ -650,20 +650,38 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
+      req_u->sat_e = u3_rsat_plan;
+      _http_start_respond(req_u, u3k(status), u3k(headers), u3k(data), c3y);
+    }
+  }
   u3z(nun);
 }
 
 typedef struct _cache_scry_cb_t {
   u3_hreq* req_u;
+  u3_noun  req;
```
joemfb, line 684: It looks like `req` is no longer used, can you remove it from here and `_http_cache_respond()`?

pkg/vere/io/http.c
```diff
@@ -650,20 +650,38 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
+  u3_httd* htd_u = req_u->hon_u->htp_u->htd_u;
 
-  req_u->sat_e = u3_rsat_plan;
-  _http_start_respond(req_u, u3k(status), u3k(headers), u3k(data), c3y);
+  if ( u3_none == nun ) {
+    h2o_send_error_500(rec_u, "Internal Server Error", "scry failed", 0);
+  }
+  else if ( (u3_none == u3t(nun)) || (u3_none == u3t(u3t(nun))) ) {
+    h2o_send_error_404(rec_u, "Not Found", "not found", 0);
+  }
```
joemfb, line 662: These checks won't work: `u3t()` will bail on an atom, not return `u3_none` (which is "not a known", distinct from `u3_nul`).

Because we used a "last" scry, `nun` is `(unit ,[%omen path (cask)])`. The arvo/vere interface collapses the outer unit from the vane scry (`~` and `[~ ~]` both become `~`).  And `%eyre`'s `/cache` scry interface doesn't let us distinguish between failed scrys and empty cache entries.

I think we want `u3_nul == nun` => 404, and everything else 500. We can't match %eyre's request handling behavior exactly wrt 404s, but they're probably preferable regardless.

You can use `u3r_at(7, nun)` to get a `u3_weak` equivalent to `u3t(u3t(nun))` (check against `u3_none`), or you could check `c3n == u3r_trel(nun, 0, 0, &some_noun)`.

joemfb (changes requested):

pkg/vere/io/http.c
lukechampine: fixed, thanks!

pkg/vere/io/http.c
```diff
@@ -683,7 +700,7 @@ _http_cache_scry_cb(void* vod_p, u3_noun nun)
 /* _http_req_cache(): attempt to serve http request from cache
 */
 static c3_o
-_http_req_cache(u3_hreq* req_u)
+_http_req_cache(u3_hreq* req_u, u3_noun req)
```
joemfb, line 703: `req` is superfluous here.

pkg/vere/io/http.c
```diff
@@ -1100,7 +1117,7 @@ _http_rec_accept(h2o_handler_t* han_u, h2o_req_t* rec_u)
   }
   else {
     u3_hreq* req_u = _http_req_prepare(rec_u, _http_req_new);
-    if ( c3n == _http_req_cache(req_u) ) {
+    if ( c3n == _http_req_cache(req_u, req) ) {
```
joemfb, line 1120: and here

joemfb (changes requested):

joemfb:
@lukechampine I pushed a first draft of state-tracking for these requests to `jb/eyre-cache-state`. Haven't had a chance to test it yet.

pkg/vere/io/http.c
```diff
@@ -651,14 +651,31 @@ _http_req_dispatch(u3_hreq* req_u, u3_noun req)
...
+  }
+  else {
+    u3_noun auth, response_header, data;
+    u3x_qual(u3k(u3t(u3t(nun))), &auth, 0, &response_header, &data);
+    u3_noun status, headers;
+    u3x_cell(response_header, &status, &headers);
+
+    // check auth
+    c3_o aut_o = (c3n == auth) || _http_req_is_auth(&htd_u->fig_u, rec_u);
+    if ( c3n == aut_o ) {
```
joemfb, line 671: Mixing boolean/loobean logic in an assignment like this is fraught. `(c3n == auth)` converts a loobean to a boolean, while `_http_req_is_auth()` returns a boolean. You can use the `c3o()` macro for loobean logical-OR, but i think it's clearest to write explicit conditionals:

```
if (  (c3y == auth)
   && (c3n == _http_req_is_auth(&htd_u->fig_u, rec_u)) )
{
  //  404
}
```

joemfb (changes requested):

joemfb (approved):
I've confirmed that this fixes the crash on null cache entries, and adds functioning authentication. I'll open another PR with lifecycle tracking and some refcount fixes.

PR #397, opened 2023-05-17
jets: refactors and optimizes tree-math jets

joemfb:
Building on urbit/urbit#6592, this PR adds jets for `+pin` and `+hub`. Additionally, it fixes some unsafe `c3_w` -> `u3_atom` conversions and optimizes `+peg` and `+mas`.

This is a draft PR as further testing of these changes is still needed.

/cc @eamsden


pkg/noun/jets/c/hub.c
```diff
@@ -0,0 +1,61 @@
...
+    u3_atom c, d, e, f, g;
+
+    c = u3i_word(b_w - a_w);
+    d = u3qc_rsh(0, c, b);
+
+    if ( c3n == u3r_sing(a, d) ) return u3m_bail(c3__exit);
+
+    e = u3qc_bex(c);
+    f = u3qa_dec(e);
+    g = u3qc_dis(b, f);
```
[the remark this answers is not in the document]
eamsden: Looks like your math is right and mine is off by one.

pkg/noun/jets/c/hub.c
```diff
@@ -0,0 +1,61 @@
...
+  b_w = u3r_met(0, b);
+
+  if ( b_w < a_w ) return u3m_bail(c3__exit);
+
+  c_w = b_w - a_w;
+
+  if ( (c3y == u3a_is_cat(a)) && (c3y == u3a_is_cat(b)) ) {
+    if ( a != (b >> c_w) ) return u3m_bail(c3__exit);
+
+    return b & ((1 << c_w) - 1);
```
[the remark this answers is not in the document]
eamsden: Yup that's right

eamsden (dismissed):
LGTM now

ashelkovnykov (dismissed):

joemfb:
If urbit/urbit#6819 is ever restored, 2d696e3bf570e3b7b442dc73fea601e2f50649c7 should be reverted.

pkg/noun/jets/c/mas.c
```diff
@@ -1,45 +1,49 @@
...
+      c3_d a_d = u3r_chub(0, a);
+      a_d  &= ~((c3_d)1 << (b_w - 1));
+      a_d  |=  ((c3_d)1 << (b_w - 2));
+      return u3i_chub(a_d);
+    }
+    else {
+      u3i_slab sab_u;
+      u3i_slab_from(&sab_u, a, 0, --b_w);
+
+      sab_u.buf_w[(b_w >> 5)] &= ((c3_w)1 << (b_w & 31)) - 1;
```
joemfb, line 36: NB: slab_from rounds *up* to word boundaries in the slice specified, so we need to mask off any more significant bits than we want.

pkg/noun/jets/c/mas.c
```diff
@@ -1,45 +1,49 @@
...
+      b_w--;
+      sab_u.buf_w[(b_w >> 5)] |= ((c3_w)1 << (b_w & 31));
```
eamsden, line 38: This mask and store logic is wrong in the boundary condition. @joemfb is going to put masking logic for sub-word blocksizes in `u3i_slab_from()`

eamsden (changes requested):

pkg/noun/jets/c/mas.c
joemfb: Thanks, this was a great catch

joemfb:
I've updated `u3i_slab_from()` to correctly mask off any unwanted bits from the most significant word when asked to truncate its input atom.  All prior call sites used at least the exact word-length of the input atom, so the old behavior was correct in all those cases. But it the API was riskier than needed (and my local work around for it was wrong).

I've corrected the `+mas` jet, and added some trivial tests to sanity-check the boundary condition.

pkova (approved):

Issue #405, opened 2023-05-22
vere uses too much memory


joemfb:
The current release (v2.6) and prior maintain the entire persistent state in memory. As of #402, clean, persistent pages of the home-road heap (north.bin) are in a file-backed mapping, reducing the resident set and improving performance under memory pressure (no need to swap those pages). When the snapshot is updated, the full home-read heap is remapped to north.bin, and all ephemeral memory (contiguous free space within the loom) is discarded with `MADV_DONTNEED`. But much more can be done.

------

To reduce ephemeral memory usage:

- unused pages can be returned to the host OS more aggressively

`MADV_FREE` or `MADV_DONTNEED` could be issued as often as after every event.

- consume memory more incrementally

unused pages are currently initialized to dirty, read-write state, which reduces page faults.
an option to initialize to a clean, read-only state would reduce pre-commitment overhead (RAM + swap),
allowing larger loom sizes by default.

- roll-your-own swap

for environment were configuring/managing swap space is impossible or honerous (apparently including kubernetes),
the ephemeral portions of the loom could be placed in an ephemeral file-backed mapping (and never sync'd).

- improve tooling efficiency

mass, pack, and meld each make the entire persistent state resident. additionally, meld uses absurd amounts of off-loom heap.

- adapt the *ares* PMA design to vere

long term, this is the most efficient/promising design for a single-level store

------

To reduce persistent memory usage

- reclaim from cold jet state

unreferenced registrations could be deleted (after a full `|meld`, to ensure that refcounts are semantically precise).

- compress cells

a special representation for small allocations would allow us to reduce cells from 24 to 16 bytes.
(aligning on 16 bytes would then allow doubling the loom maximum to 16GB)

see %6599

philipcmonk:
It is useful to be able to put a (mostly) hard limit on the amount of memory that a ship can use.  Demand paging is relatively good at increasing the number of ships that can be run in a given amount of memory, but relatively bad at reducing the amount of memory that a ship needs, at least when interpreted naively.  If we roll our own swap, then we can say something like: as long as you have a few tens of MB for vere's non-loom state, you can restrict memory as much as you want, and it will only impact performance (and ability to meld).

To roll our own swap, we would need to map the ephemeral memory to a file, and I suppose also any time a clean page is dirtied, that would also get mapped into that file?  And yeah, never sync it.  Is that as easy as:
- creating such a file
[...]

PR #410, opened 2023-05-23
vere: map ephemeral space to file in pier

philipcmonk:
See #405

This implements a "roll-your-own swapfile", where we designate a file (.urb/ephemeral.bin) to file-back any parts of the loom which are not backed by the snapshot.  These are pages in the contiguous free space, and pages that have been dirtied since the last snapshot.

With these changes, you should be able to safely reduce the allotted memory to a VM running a machine without relying on separate swap space.

The main disadvantages to this approach are:
- ephemeral.bin is large (the size of the loom).  We could delete it on graceful shutdown, or we could make it a sparse file.
- _ce_flaw_protect has to copy the contents of dirtied pages twice.  I'm skeptical that this introduces significant slowness, but it could.

It may be reasonable to put this behavior behind a flag simply because of the increased disk requirements.

I've successfully tested various scenarios in a limited memory environment, but this is not well enough tested to be ready for merging.  Main things I've tested (all on a machine with 1GB RAM and no swap):
- Boot
- Allocate 8 128MB atoms in the dojo in a row, verifying |mass is 1.2GB afterward
- Free those atoms, run |pack, recreate them, free them again, |pack again
- Restart ship, recreate those atoms, free them, and |pack

All of this worked fine on that machine, even though the ephemeral memory usage must have been greater than the RAM on the machine, and there were no swap files.  During most of this, the htop-reported "resident memory" for the serf process was around 400-600MB, while the amount of memory "used" overall per `free -h` was under 180-250MB every time I checked.  This is consistent with the OS deciding how much of the backing files to keep resident at any given time, while the amount of memory strictly required stays fairly small.

Testing on a machine with plenty of memory, I didn't notice any slowdown.  On the 1GB machine, it was somewhat slow to allocate the large atoms, but it felt about right considering they must have been written to the backing file.  Graceful shutdown was somewhat slow sometimes, presumably when it needed to copy from the ephemeral file to the snapshot.


pkg/noun/events.c
```diff
@@ -181,12 +181,45 @@ u3e_check(c3_c* cap_c)
 static inline c3_i
 _ce_flaw_protect(c3_w pag_w)
 {
+  // NB: if this is not static, sometimes we SIGSEGV in the memcpy,
+  // and I don't know why.
+  //
+  static c3_y con_y[16384];
```
joemfb, line 187: The stack is grown via page faults, and a non-static declaration would be allocated on the stack at call-time.

pkg/noun/events.c
```diff
@@ -181,12 +181,45 @@ u3e_check(c3_c* cap_c)
...
+
+  // a guard page is PROT_NONE, so the memcpy will fail.  even if we
+  // excluded the current guard page, sometimes the guard page is moved
+  // without resetting PROT_NONE.  we handle all such cases by
+  // mprotecting ahead of the memcpy.
+  //
```
joemfb, line 193: > sometimes the guard page is moved without resetting PROT_NONE

That should no longer be the case, have you seen otherwise?

pkg/noun/events.c
```diff
@@ -988,7 +1051,7 @@ _ce_loom_mapf_north(c3_i fid_i, c3_w pgs_w, c3_w old_w)
     if ( MAP_FAILED == mmap(_ce_ptr(0),
                             _ce_len(pgs_w),
                             PROT_READ,
-                            (MAP_FIXED | MAP_PRIVATE),
+                            (MAP_FIXED | MAP_SHARED),
```
joemfb, line 1054: `MAP_SHARED` here breaks the atomicity of snapshot updates, allowing the kernel to flush modifications to disk unpredictably. What motivated the change?

pkg/noun/events.c
```diff
@@ -314,6 +348,35 @@ _ce_image_stat(u3e_image* img_u, c3_w* pgs_w)
...
+  c3_mkdir(ful_c, 0700);
+
+  snprintf(ful_c, 8192, "%s/.urb/ephemeral.bin", u3P.dir_c);
+  if ( -1 == (*eph_i = c3_open(ful_c, mod_i, 0666)) ) {
+    fprintf(stderr, "loom: ephemeral c3_open %s: %s\r\n", ful_c,
+            strerror(errno));
+    return c3n;
+  }
+
+  if ( ftruncate(*eph_i, u3a_bytes) < 0 ) {
```
joemfb, line 372: `u3a_bytes` is the compile-time maximum (8 GB). you can use `_ce_len(u3P.pag_w)` to for "`--loom` in bytes"

pkg/noun/events.c
joemfb: It occurs to me that another way to trade off disk/memory usage would be to double-buffer north.bin on disk, which would allow it to always be read-write/MAP_SHARED. But that would be a more invasive change.

joemfb (commented):
This is very interesting. I hadn't thought to remap dirty, persistent pages onto the ephemeral store, and just assumed they would have to contribute to the resident set.

I do think this feature should be opt-in, behind a command-line flag.

As for limiting disk i/o, you can apparently use `msync(MS_INVALIDATE)` to "discard" writes through a MAP_SHARED mapping: https://groups.google.com/g/comp.os.linux.development.system/c/BZoz6GBc7Vw. I don't know if that's an practical alternative to MADV_DONTNEED, or if we'd want both.

pkg/noun/events.c
philipcmonk: I did see this on page 0, possibly related to the "guard on remap" case, which I ended up deleting anyway.  That case started getting triggered because I changed `old_w` to not be 0 on start, so that we map everything to the ephemeral file immediately.  Could just explicitly mmap everything to the ephemeral file in `u3e_live` instead of changing that logic, though.

The guard on remap case doesn't make sense to me, can you remember why it was there?

pkg/noun/events.c
philipcmonk: Oh, of course.  I was not seeing the memory numbers I expected, so I changed this, but I'm pretty sure that was because the ephemeral space wasn't yet getting fully mapped in.  I'll change it back and see if the memory numbers look the same (I think they will).

Technically, it probably doesn't make a difference because `PROT_READ` should mean we never modify those addresses while they're mapped to file anyway, but `MAP_PRIVATE` is more correct.

pkg/noun/events.c
philipcmonk: that's better

pkg/noun/events.c
philipcmonk: Is that a general property of C stacks?  And the reason it makes a difference here is that we're already inside a SIGSEGV handler?

joemfb: It's an OS property; I believe all the major ones do something similar. I've written a buggy SIGSEGV handler before, and had it trigger itself endlessly. Was that what you were seeing, or did it just crash?

pkg/noun/events.c
joemfb: The guard page can be anywhere between the current home-road boundaries (cap and hat). When the heap shrinks, we remap an anonymous/ephemeral segment off the end of the home heap, over all pages between our current boundary (hat) and the previously persisted boundary. This remapping is to make sure that we never SIGBUS (due to touching a page mapped off the end of the now truncated `north.bin`). But we have no guarantee that the guard page wasn't placed somewhere in that segment, so we need to check and re-protect it if necessary.

pkg/noun/events.c
joemfb: You'd still get the fault in userspace, but without CoW semantics in the kernel.

pkg/noun/events.c
philipcmonk: It crashed with SIGSEGV (ie the king printed that the serf crashed with signal 11).  I went on a wild goose chase for a while assuming the memcpy was SIGSEGV'ing because the source was PROT_NONE, but eventually I narrowed it down to writing anything to the 7608th byte in `con_y`, which is consistent with "your stack was about halfway through this page, and you filled up the rest of the page".  Should have checked whether that was on a page boundary, but most likely it was.

pkg/noun/events.c
philipcmonk: Ah, right.  Then I should re-add that check.  I wonder if the issue is that `gar_w` defaults to 0 and isn't set until the end of `u3e_live`.  If so, there are several ways around the issue, including an explicit `mmap` in u3e_live to initialize the ephemeral mapping (analagous to the one in u3m_init).

philipcmonk: That seems to work in initial testing: 5d22c38d5f9f8e8cfb3624ae6231e0aa6a7de0f6

philipcmonk:
If we wanted to be really clever, we would mmap the whole loom in one file, and then when a page gets dirtied we would mmap that page directly to the patch file. Then to save, just msync the dirty pages, write a checksum to the patch files, apply them, and remap the whole loom.

philipcmonk:
We're getting good results testing this in our hosting product.  Other than testing, I think what I want to do is:

- Add a flag to turn this on, perhaps `--use-ephemeral-file`
- Add a flag to set the path for the ephemeral file, perhaps `--ephemeral-file=/foo`.  This will make it easier to put the ephemeral file on a separate drive if desired, since it has such different usage characteristics.
- Delete the ephemeral file on graceful shutdown
- Research sparse files to see if it's possible to use them for the ephemeral file

joemfb:
I think the args should clearly be `--swap` and `--swap-to /path/to/file` (might as well lean in to the stolen valor).

Ignore what i said/linked re:`MS_INVALIDATE`, that thread contradicts other discussions and docs. I think we want more frequent `MADV_DONTNEED`, possibly after every event. If sparse files can work, then `MADV_REMOVE` would be preferable on linux.

pkg/noun/events.c
```diff
@@ -179,12 +179,47 @@ u3e_check(c3_c* cap_c)
...
+      fprintf(stderr, "loom: fault mmap failed (%u): %s\r\n",
+                       pag_w, strerror(errno));
+      return 1;
+    }
+
+    // restore contents of page
+    //
+    if ( gar_w != pag_w ) {
+      memcpy(_ce_ptr(pag_w), con_y, _ce_page);
+    }
```
joemfb, line 222: `_ce_flaw_protect()` has two call sites, with three scenarios:

- u3e_ward(): retire the guard page (since we're moving it preemptively)
- u3e_fault():
  - retire the guard page (since we hit it)
  - dirty a persistent page

The old definition is sufficient for the guard page case; all that's need is to establish read-write protections. And the double-copy + remapping is only needed when dirtying a persistent page. So i think they should just be two different functions, with a little refactoring of the `u3e_fault()` call site (further distinguishing guard-page handling). That way, no pre-protection or guard-page logic is needed in the copy + remap function.

joemfb (changes requested):

pkg/noun/events.c
philipcmonk: Makes sense.  FYI, I'm out the rest of the week, so I'll do these tasks next week.

justinlynn:
As we're implementing what is effectively a swap system in user space, as an operator of vere in a potentially memory contended environment, it would be extremely useful to have metrics such as those in https://www.kernel.org/doc/html/latest/accounting/psi.html exposed by our custom allocator. This way I can detect when our allocated amount of memory (cgroupsv2 / k8s on cgroupsv2 will limit available page cache to the containers memory request/limit which can be quite small) is too low to permit vere to run without stalling on IO. Would it be possible to design (and/or implement) this aspect of user-space-swap as a part of this work?

justinlynn:
Given that this might also increase shutdown times, as mentioned by Philip, is there a way we can add a shutdown mode that can optionally always toss out the ephemeral loom while leaving the system snapshot in a consistent state, even if a short log replay is necessary on restart (perhaps minimising IO as much as possible, while capping the number of events that would likely need to be replayed? I'm just thinking how we can avoid huge bursts of IO during the system lifecycle (esp. at shutdown when the host is likely to be doing a lot of other runtime shutdowns at the same time, and needs to do them as quickly as possible - say, within 20 seconds in the case of power failure or "spot market" vm termination).

philipcmonk:
Another thought that occurred to me is that a specific backing file for|meld could be nice, because that's one of the few operations that takes much more memory.

pkg/noun/events.c
chc4: The stack is grown via page faults, but those aren't delivered as SIGSEGV's and are handled by the kernel transparently. The reason is probably that signal delivery happens on a seperate stack from normal execution, and I believe most libcs set a fixed sized memory region for that instead of an actual MAP_GROWDOWN page.

chc4: Having it as a static means you're wasting 16kb of memory all the time, though, instead of it being available to the system outside the sigsegv handler. It could be replaced with a seperate mmap'd and cleaned up temporary memory region instead.

joemfb:
@justinlynn we're not fully implementing swap here, we don't have faults to track reads and establish an LRU ordering of pages. It'd be nice to monitor the write volume to this ephemeral backing file, but i haven't found a good way to do that. Maybe you could back that info out of `vm_stat` or something in procfs? Otherwise, the simplest thing would probably be to put the ephemeral backing file on a dedicated partition and just monitor disk i/o generally on that.

For quick shutdown, i think you want MADV_DONTNEED on the ephemeral mapping, then to SIGTERM/SIGKILL as normal. We might need a signal handler (or `atexit()`) to do that reliably.

justinlynn:
Thanks for the context @joemfb - will look into your suggestions and see what I can contribute 👍🏻

pkg/noun/events.c
philipcmonk: Done

philipcmonk:
I've completed the tasks I listed above, so this is ready to go pending additional testing.  @joemfb can you give it a review before I do final testing?

Ignore the failing linux-aarch64 CI failure, that's broken on develop.

joemfb (changes requested):
This looks good, thanks for the refactoring. I'm surprised the backing file is sparse with explicit fallocate or hole-punching syscalls.

I don't think it should be on by default, at least unless/until we've verified that it's no worse than using *actual* swap, and ideally better. (And, fwiw, I don't expect that to be the case: https://github.com/urbit/vere/pull/422#issuecomment-1591736200)

The name `ephemeral.bin` is unaesthetic. north/south are 5 characters, so "limbo" would work. memory/control are 6 and 7, so "swap" would make a contiguous sequence.

A couple more nits:

philipcmonk:
> I'm surprised the backing file is sparse with explicit fallocate or hole-punching syscalls.

Yeah, apparently `ftruncate` creates sparse files on most filesystems.

> The name ephemeral.bin is unaesthetic. north/south are 5 characters, so "limbo" would work. memory/control are 6 and 7, so "swap" would make a contiguous sequence.

I like limbo.  Right now it's directly in `.urb/`, should that be `.urb/chk/limbo.bin` or `.urb/limbo.bin`?

pkg/noun/events.c
```diff
@@ -270,7 +306,20 @@ u3e_fault(u3_post low_p, u3_post hig_p, u3_post off_p)
...
+  }
+  else
+#endif
+  if ( u3P.eph_i ) {
+    if ( _ce_flaw_mmap(pag_w) ) {
+      return u3e_flaw_base;
+    }
+  }
+  else if ( _ce_flaw_mprotect(pag_w) ) {
```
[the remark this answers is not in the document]
philipcmonk: I think this is correct.  We duplicate the `dit_w` handling, but I think it's still better.

pkg/noun/events.c
```diff
@@ -296,24 +296,29 @@ u3e_fault(u3_post low_p, u3_post hig_p, u3_post off_p)
       fprintf(stderr, "loom: strange guard (%d)\r\n", pag_w);
       return u3e_flaw_sham;
     }
+
+    if ( u3P.dit_w[blk_w] & (1 << bit_w) ) {
+      fprintf(stderr, "loom: strange guard page (%d): %x\r\n", pag_w, off_p);
+      return u3e_flaw_sham;
+    }
+
+    u3P.dit_w[blk_w] |= (1 << bit_w);
```
joemfb, line 305: Clearing the bit here is redundant, it's an error if it's set.

philipcmonk:
I put it in `.urb/chk/limbo.bin`, and addressed the other review comments.

I think this is ready to go now.

joemfb (changes requested):
I'd be inclined to put the backing file in `.urb/chk/` for consistency with the rest of the snapshot system. But I don't feel strongly about it, it should be gone after a graceful shutdown regardless.

One more nit:

pkg/noun/events.c
philipcmonk: Good catch, that refactor was wrong.  I tested that this version failed whenever we move the guard page, and deleting this made it work in the same case.

joemfb: Glad to see CI caught that as well: https://github.com/urbit/vere/actions/runs/5326673779

joemfb (approved):
this will do

PR #210, opened 2023-02-08
`urbit play`: replay event log subcommand

sigilante:
Resolves #157; see PR #158 for discussion.

Playback is normal speed and Ctrl+Z works to exit without hanging.  This may have inadvertently clobbered some of your replay stuff @mcevoypeter.

After `urbit play` succeeds, it boots into a prompt—not sure if this is desirable behavior or not.

PR #192, opened 2023-02-01
Implement simple replay command

mcevoypeter:
## Description

Resolves #157. As described in https://github.com/urbit/vere/issues/157#issuecomment-1409520195, this PR adds support for a `play` subcommand, which piggybacks off the existing replay implementation. This PR also augments the existing replay implementation with per-batch snapshots and a command line configurable batch size., which makes replaying large event logs a less painful experience for users. Snapshotting after each replay batch of course incurs a performance hit because we're writing to disk more frequently during replay, but replay is already slow, and in practice, users should not expect to see any difference. Additionally, this PR adds command line configurable replay batch size gives users a knob for increasing or decreasing the frequency with which snapshots are taken.

## Testing

```
$ bazel build :urbit
<snip>
$ ./urbit -F zod -u https://bootstrap.urbit.org/urbit-v1.17.pill
<snip>
$ rm zod/.urb/chk/*.bin
$ rm zod/.urb/chk/*.bin
$ ./urbit --replay-to 103 zod
~
urbit 1.20-c98600a5ed
boot: home is /home/tlon/code/gh/urbit/vere/zod
loom: mapped 2048MB
lite: arvo formula 2a2274c9
lite: core 4bb376f0
lite: final state 4bb376f0
loom: mapped 2048MB
boot: protected loom
live: logical boot
boot: installed 652 jets
pier: replay till 103
---------------- playback starting ----------------
pier: replaying events 1-103
<snip>
pier: (103): play: done
---------------- playback complete ----------------
pier: replay barrier reached, cramming
serf (103): saving rock
hash-cons arena:
  root: B/160
  atoms (6939):
    refs: KB/218.920
    data: MB/16.597.260
    dict: KB/206.720
  total: MB/17.022.900
  cells (2598252):
    refs: MB/70.491.560
    dict: MB/174.264.720
  total: MB/244.756.280
total: MB/261.779.340

%group-store: on-peek on path /x/whey
clay: read-at-aeon fail [desk=%base care=%c case=[%da p=~2023.2.1..17.47.34..2cb5] path=/noun/mass]
clay: no files match /mar/mass/hoon
[%error-building-cast %noun %mass]
[%error-building-tube %noun %mass]
peek no tube from noun to mass
serf: measuring memory:

%arvo:
  %hoon:
    %one: KB/29.160
    %two: KB/188.064
    %tri: KB/280.708
    %qua: KB/683.716
    %pen: MB/3.092.604
  --MB/4.274.252
  %hex: KB/562.264
[...]

Issue #157, opened 2023-01-24
Restore event log replay functionality.


sigilante:
> [Pull] the event log replay functionality out of v1.14 of Vere with the intention of it landing in v1.19.

> 1.14 was reverted due to some kind of corruption error but the error was from a different part of the release so the idea here is to extract the replay improvements from 1.14 and make a PR that just contains those

sigilante:
> None of the meld/memory paging/printing stuff, just the replay.
>
> Not this, for instance: https://github.com/urbit/urbit/pull/6159/files#diff-9b3877485f99b0b6591908d549b6ac16cf7e327e33d941018ca339e590622a47R648
>
>Hmm, it seems to rely on automeld. Moving convo to issue.

`_cw_play` in `main.c` uses `u3o_auto_meld` and thus implicates the whole meld set of changes.

sigilante:
@belisarius222 what was actually causing the memory corruption issue?  It wasn't the meld stuff was it?

joemfb:
Automeld was added (ported) for that feature, specifically. On that branch / as of that pr, it wasn't used anywhere else.

sigilante:
What I'm worried about is that I'm going to inadvertently regress something. There are a lot of data structures with single elements added because of all the automeld and CLI options.

Git diffs don't work because the repo rearchitecture changed so many minor things as to be unwieldy.

Thoughts?

Example: the code I'm adding back in refers to gar_w in _u3e_pool.

https://github.com/urbit/urbit/blob/a7880717e3c77c21e185ffe52a4982b0b3bd52d1/pkg/urbit/include/noun/events.h#L45

So I have to add this back in, or I have to backtrace the entire logic for why it's not there now.

Include all of the guard page stuff as well then?

sigilante:
Here's the direct comparison:  https://github.com/urbit/vere/compare/urbit-v1.14-rc3...vere-v1.17

I've been working off of https://github.com/urbit/urbit/pull/6159/files to make changes thus far.

matthew-levan:
@sigilante what's the difference between the functionality you're referring to here and the functionality provided by the existing `--replay-from NUMBER`, `--replay-to NUMBER`, and `--replay` options?

sigilante:
@matthew-levan This restores the 1.14 replay stuff:

> - adds the `play` command, for improved replay of large event logs,
with automatic retry of bail:meme after pack (or meld with --auto-meld),
snapshot updates after errors or interruption (via ctrl-c), and
[...]

PR #413, opened 2023-05-30
`urbit play`: updated with working batch size / loom settings

ryjm:
Updated #210. A few things still need to be unclobbered and I haven't tested the pill issue discussed in #157, but the `play` subcommand now works with custom batch sizes and loom (which was missing in  #192)  Needed this to speed up ~littel-wolfur's replay, and I'm running it on a copy now.


ryjm:
Superseded by #420

PR #420, opened 2023-05-30
play: dedicated event replay, revived

### LINKED ISSUE #157: Restore event log replay functionality.

joemfb:
This PR ports urbit/urbit#6159, fixing a performance problem that plagued previous porting attempts. Fixes #157, supersedes #210 and #413.

The poor performance observed in #210 and elsewhere was not due to any issue matching or dispatching jets. It coincided with the switch from hoon %140 to %139, but only incidentally. It was caused by a change to the `+solid` pill generator, which inadvertently broke the structural sharing in the lifecycle sequence (see https://github.com/urbit/urbit/pull/5989/files#diff-2f8df9d079ccb58c0a9a9c46f2f7dbd943dabaa21ba658c839de757bbac999f1L108-L116). The problem was unnoticed because, in normal (ie, king/serf) boot and replay, events are sent over IPC in batches, which had the side effect of recovering the necessary structural sharing. This new replay implementation does not involve IPC, but instead reads and computes events synchronously, in a single process.

The issue did not arise until ships booted from pills created with the updated generator were replayed using this new implementation, and that happened to coincide with the release of hoon %139. The absence of structural shared lead to jets being registered with one copy of the kernel, but dispatched from a separate copy, resulting in absurdly expensive equality comparisons. Since both copies were already allocated on the home-road, unification could not be performed. And since the problem manifested during the initial phase (lifecycle sequence) of the boot process, `|meld` could not be used.

This PR includes a trivial hack to work around such event logs: the lifecycle sequence is read in an inner road, jammed, and then cue'd, thus recovering structural sharing before any nock computation, jet registration, &c. The solid pill generator should also be fixed, but workarounds will still be needed to account for existing piers. Longer-term, home-road unification should clearly be explored to avoid such fragility.


pkg/vere/main.c
```diff
@@ -1960,55 +1962,163 @@ _cw_pack(c3_i argc, c3_c* argv[])
...
-        fprintf(stderr, "%s\r\n", usage_c);
+      case c3__loom: {
+        c3_w lom_w;
+        c3_o res_o = _main_readw(optarg, u3a_bits + 3, &lom_w);
+        if ( (c3n == res_o) || (lom_w < 20) ) {
+          fprintf(stderr, "error: --loom must be >= 20 and <= %u\r\n", u3a_bits + 2);
+          exit(1);
+        }
+        u3_Host.ops_u.lom_y = lom_w;
+      } break;
```
ryjm, line 2019:
```suggestion
      case c3__loom: {
        if (_main_readw_loom("loom", &u3_Host.ops_u.lom_y)) {
          exit(1);
        }
      } break;
```

Fails on `--loom 33` otherwise.

barter-simsum: It should use `u3a_bits_max + 1` as the comparison instead. We're already doing this but was probably missed when old changes were ported

joemfb: Thanks, I had just ported the old version straight over.

pkg/vere/mars.c
```diff
@@ -0,0 +1,288 @@
...
+            u3a_print_memory(stderr, "mars: meld: gained", u3u_meld());
+          }
+          else {
+            u3a_print_memory(stderr, "mars: pack: gained", u3m_pack());
+          }
+        } break;
+
+        //  XX handle any specifically?
+        //
+        case _play_int_e:
```
ashelkovnykov, line 271: A separate message for an interrupted replay would be nice, since it didn't fail so much as become paused

ashelkovnykov (commented):
LGTM

pkg/vere/main.c
```diff
@@ -1960,55 +1962,159 @@ _cw_pack(c3_i argc, c3_c* argv[])
...
   static struct option lop_u[] = {
-    { "batch-size", required_argument, NULL, 'b' },
-    { "replay-to",  required_argument, NULL, 'n' },
-    { NULL,         0,                 NULL, 0   }
+    { "loom",      required_argument, NULL, c3__loom },
+    { "no-demand", no_argument,       NULL, 6 },
+    { "auto-meld", no_argument,       NULL, 7 },
+    { "full",      required_argument, NULL, 'f' },
+    { "replay-to", no_argument,       NULL, 'n' },
+    { "snap-at",   no_argument,       NULL, 's' },
```
joemfb, line 2003: s/b `required_argument`

mopfel-winrux (commented):
@joemfb walked @matthew-levan and I through this today and we added changes as we found them. Once those are fixed it LGTM

zalberico:
Looking forward to this one

pkg/vere/mars.c
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    mar_u->mug_l = u3r_mug(u3A->roc);
+
+    if ( tac_u.mug_l && (mar_u->mug_l != tac_u.mug_l) ) {
+      fprintf(stderr, "play (%" PRIu64 "): mug mismatch "
+                      "expected %08x, actual %08x\r\n",
+                      tac_u.eve_d, tac_u.mug_l, mar_u->mug_l);
+
+      if ( c3y == mug_o ) {
```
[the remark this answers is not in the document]
joemfb: Mug mismatch checks have been soft/advisory since mid-2020. They had been added and mandatory roughly a year prior to that. The problem was that they actually occurred with some frequency, and there was no way for people to work around / get past them.

They're actually quite rare now (as far as I know), and always indicate a serious problem. But people still need to be able to get past them in some scenarios (attempt to recover data, for instance). So I've made the checks hard by default, and added an argument to soften them up and continue replay.

joemfb:
This PR is done and tested, ready for final review.

ryjm:
for posterity: 
I was able to use this PR to replay ~littel-wolfur from 1.2b to 1.8b events with great success. events/sec were fairly variable - started at a brisk 2k/s but 1.7-1.8b was a slog.

mopfel-winrux (dismissed):
LGTM, Major issues we discussed in the review were done.

pkg/noun/vortex.c
```diff
@@ -50,7 +59,8 @@ u3v_boot(u3_noun eve)
       return c3n;
     }
 
-    u3A->roc = u3k(u3t(pro));
+    u3A->roc   = u3k(u3t(pro));
+    u3A->eve_d = len_d;
```
matthew-levan, line 63: Just for my own clarity, what are these changes in `u3v_boot` for?

matthew-levan (commented):
This looks good. I left a few small comments. Also, I noticed there are a number of `XX` comments left here though, especially in `u3_mars_play`-- are any of them ripe for actually implementing and/or resolving before merging this, in your opinion?

pkg/vere/main.c
```diff
@@ -1960,55 +1962,168 @@ _cw_pack(c3_i argc, c3_c* argv[])
...
-                                "| --batch-size <event_cnt>] <pier>`";
   static struct option lop_u[] = {
-    { "batch-size", required_argument, NULL, 'b' },
-    { "replay-to",  required_argument, NULL, 'n' },
-    { NULL,         0,                 NULL, 0   }
+    { "loom",      required_argument, NULL, c3__loom },
+    { "no-demand", no_argument,       NULL, 6 },
+    { "auto-meld", no_argument,       NULL, 7 },
+    { "soft-mugs", no_argument,       NULL, 8 },
+    { "full",      required_argument, NULL, 'f' },
```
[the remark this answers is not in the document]
joemfb: Good catch, thanks. I missed this because I was testing with `urbit play -f $pier`, where `$pier` was doubling up as the argument to `-f`, then ignored.

pkg/noun/vortex.c
joemfb: Calculate the length of the lifecycle sequence, store that length and result globally. "boot and persist the result". This function was old and unused prior to this PR, the in-use equivalent is `_serf_play_life()` in `serf.c` (which also includes IPC response details).

pkg/vere/mars.c
```diff
@@ -0,0 +1,354 @@
...
+u3_mars_play(u3_mars* mar_u, c3_d eve_d, c3_d sap_d)
+{
+  u3_disk* log_u = mar_u->log_u;
+
+  if ( !eve_d ) {
+    eve_d = log_u->dun_d;
+  }
+  else if ( eve_d <= mar_u->dun_d ) {
+    u3l_log("mars: already computed %" PRIu64 "", eve_d);
+    u3l_log("      state=%" PRIu64 ", log=%" PRIu64 "",
```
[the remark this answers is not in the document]
joemfb: those are optional imo

matthew-levan (approved):

PR #422, opened 2023-05-31
u3: "toss" ephemeral pages after every event

joemfb:
Highly relevant to #410 (I think this behavior should be always-on in the presence of "swap"), but also likely useful as a standalone option for low-memory deployments. This is a draft PR as the behavior is hardcoded, not controlled by command-line arguments.

### DIFF
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -1501,17 +1501,34 @@ u3e_save(u3_post low_p, u3_post hig_p)
     _ce_loom_mapf_north(u3P.nor_u.fid_i, u3P.nor_u.pgs_w, nod_w);
   }
 
-  {
-    void* ptr_v = _ce_ptr(u3P.nor_u.pgs_w);
-    c3_w  pgs_w = u3P.pag_w - (u3P.nor_u.pgs_w + u3P.sou_u.pgs_w);
-
-    if ( -1 == madvise(ptr_v, _ce_len(pgs_w), MADV_DONTNEED) ) {
-        fprintf(stderr, "loom: madvise() failed for %u pages at %p: %s\r\n",
-                        pgs_w, ptr_v, strerror(errno));
-    }
-  }
+  u3e_toss(low_p, hig_p);
 
   u3e_backup(c3n);
+}
+
+/* _ce_toss_pages(): discard ephemeral pages.
+*/
+static void
+_ce_toss_pages(c3_w nor_w, c3_w sou_w)
+{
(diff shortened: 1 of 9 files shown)


joemfb:
This is (most of) the first option in #405

philipcmonk:
```
              After a successful MADV_DONTNEED operation, the semantics
              of memory access in the specified region are changed:
              subsequent accesses of pages in the range will succeed,
              but will result in either repopulating the memory contents
              from the up-to-date contents of the underlying mapped file
              (for shared file mappings, shared anonymous mappings, and
              shmem-based techniques such as System V shared memory
              segments) or zero-fill-on-demand pages for anonymous
              private mappings.
```
If I'm reading this correctly, `MADV_DONTNEED` on shared mappings (such as the ephemeral file) will make us read in the contents again from disk if we tried to read from that memory again.  We always write before reading -- do you know if that will be sufficient to avoid reading in garbage from the ephemeral file?

ashelkovnykov:
@philipcmonk Seems to me that writing before flushing should always be sufficient to avoid data loss

Two questions:
1. This improvement is aimed at hosting providers, correct? Hosting providers need to reduce the active memory set at all times to reduce costs. However, it doesn't matter on my local machine if the inner-road pages are considered as "actively being used" by the kernel, especially if they're backed by the ephemeral file, since they can be LRU bumped to swap/disk (which is what I would want to happen, if I'm running so many applications on my PC that I can't fit them all + an 8GB Urbit into memory).
2. Have you run a test on the performance impact caused by the first few inner-road pages thrashing (I would assume that they are being advised `MADV_DONTNEED`, then being hit and loaded from disk in short order frequently)? I can't imagine it being significant, but it's gotta be non-zero.

The code itself looks good

philipcmonk:
I'm not worried about data loss, since this is only the contiguous free space. I'm wondering if we'll unnecessarily read data off the disk at the beginning of each event, which could add significant latency.

ashelkovnykov:
@philipcmonk Ah okay, I see now. We're curious about the same thing, then

joemfb:
I've retargeted this PR against #410 to avoid options/arguments/ipc merge conflicts. I've made the "toss" option independent of "swap", including a configurable number of words (on reflection, this should probably be pages) to keep and avoid thrashing, and wired up a commandline argument. (only for normal operation, it's not obvious when to trigger this in subcommands, and also not a priority.)

philipcmonk (approved):
This looks good.  I look forward to trying this out to see how it affects disk IO and perceived speed.

PR #402, opened 2023-05-18
demand paging redux

joemfb:
This PR is a forward-port / rewrite of the demand-paging implementation from v1.14 (see urbit/urbit#6063, urbit/urbit#6126, urbit/urbit#6127, and urbit/urbit#6152). The original scope has been decreased, and the implementation simplified: i/o errors are not retried, the dirty page bitmap is manipulated with much simpler code, page offsets/pointers are calculated with macros, &c. There are additional layers of snapshot validation for updates (controlled at compile time, always on as of this PR); clean pages are compared to disk both before and after update. (This validation should stay on for pre-release testing, and possibly for initial release as well.)

This PR has been tested extensively on live ships; the corruption issues that plagued v1.14 cannot be reproduced.

Fixes #188. Supersedes #277. (Was previously opened as #401, but a typo in the branch name was preventing updates.)

PR #423, opened 2023-05-31
u3: specify madv_random for file-backed snapshot mapping

joemfb:
Since #402 was released, there have been isolated reports of unusually high disk i/o, specifically reads. This is unexpected behavior -- our current, best guess is the kernel's readahead logic for file-backed mappings. This PR implements the standard solution to overeager readahead (MADV_RANDOM), as a draft since the diagnosis is not confirmed, and the advice may need to be behind a flag.

### DIFF
--- a/pkg/noun/events.c
+++ b/pkg/noun/events.c
@@ -995,6 +995,11 @@ _ce_loom_mapf_north(c3_i fid_i, c3_w pgs_w, c3_w old_w)
                       pgs_w, strerror(errno));
       u3_assert(0);
     }
+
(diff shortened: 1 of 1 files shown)


ashelkovnykov:
For posterity: the actual issue was #427

joemfb:
Thanks @mopfel-winrux, @ashelkovnykov.

I think this advice is the right thing for north.bin regardless. And I'd rather not add yet-another flag to control it. But it's not pressing, and should be measured first. Leaving as a draft for now.

PR #441, opened 2023-06-05
build: ensure statically linked `openssl`

### LINKED ISSUE #440: bazel build fails on macOS when some of our external dependencies are also present in `/usr/local/include` or other system directories
@pkova @mopfel-winrux @tadad @mikolajpp and I have experienced this with `gmp`, `openssl`, and others. The workaround for now is to remove/move the overlapping files in the system directory and build again. 

Here's an example of an error message we can see when there's a `/usr/local/include/openssl` folder on the system. Bazel *should* be ignoring this folder and using our own bundled `openssl` to statically link it, but it doesn't for some reason.

We should probably find a command that explicitly directs Bazel to *ignore* system-level files and folders for some (or all) external dependencies. cc: @mcevoypeter 

```
matt@mbp14 vere % bazel build :urbit
INFO: Analyzed target //:urbit (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /private/var/tmp/_bazel_matt/657ca343b546ec28817b7a65bf3b8da7/external/aes_siv/BUILD.bazel:10:11: Compiling aes_siv.c failed: undeclared inclusion(s) in rule '@aes_siv//:aes_siv':
this rule is missing dependency declarations for the following files included by 'aes_siv.c':
  '/usr/local/include/openssl/cmac.h'
  '/usr/local/include/openssl/evp.h'
  '/usr/local/include/openssl/opensslconf.h'
  '/usr/local/include/openssl/opensslv.h'
  '/usr/local/include/openssl/ossl_typ.h'
  '/usr/local/include/openssl/e_os2.h'
  '/usr/local/include/openssl/symhacks.h'
  '/usr/local/include/openssl/bio.h'
  '/usr/local/include/openssl/crypto.h'
  '/usr/local/include/openssl/safestack.h'
  '/usr/local/include/openssl/stack.h'
  '/usr/local/include/openssl/cryptoerr.h'
  '/usr/local/include/openssl/bioerr.h'
  '/usr/local/include/openssl/evperr.h'
  '/usr/local/include/openssl/objects.h'
  '/usr/local/include/openssl/obj_mac.h'
  '/usr/local/include/openssl/asn1.h'
  '/usr/local/include/openssl/asn1err.h'
  '/usr/local/include/openssl/bn.h'
  '/usr/local/include/openssl/bnerr.h'
  '/usr/local/include/openssl/objectserr.h'
Target //pkg/vere:urbit failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.615s, Critical Path: 0.44s
INFO: 10 processes: 9 internal, 1 darwin-sandbox.
FAILED: Build did NOT complete successfully
```

matthew-levan:
Resolves #440.


matthew-levan:
@mcevoypeter @barter-simsum looks like we might have to do something like this to ensure we're actually statically linking some of our third-party dependencies, but I'm not sure. Since upgrading to macOS 13.4, I can't build `vere` thanks to `bazel`'s misconfiguration (or plain refusal?) that leads to its inability to link our packages with our own build of `openssl`. This could be happening with `gmp` as well (cc: @mopfel-winrux).

mikolajpp:
On my macOS 13.4 I too suffer from this problem. For me, it is the `libuv` which, when installed with via homebrew breaks everything. I can't run `nvim` without it, so to build Vere I need to manually uninstall my system `libuv` every time bazel builds it.

matthew-levan:
@mikolajpp ah, sorry to hear that. Not a good experience at all. I will be continuing to look into this and keep this thread updated.

PR #444, opened 2023-06-08
fixes +fitz jet mismatch

mikolajpp:
With the recent discovery of a bug in `+fitz`, https://github.com/urbit/urbit/pull/6647, a jet mismatch was also uncovered. 
The `+fitz` jet erronously excludes all uppercase letters but the last one from the compatibility check.

```
> +fitz ['AB' 'CB']
%.n
> (fitz 'AB' 'CB')
%.y
> +fitz ['AC' 'CB']
%.n
> (fitz 'AC' 'CB')
%.y
```

Here the `+fitz` generator is the fixed Hoon version.

### DIFF
--- a/pkg/noun/jets/f/fitz.c
+++ b/pkg/noun/jets/f/fitz.c
@@ -5,7 +5,6 @@
 
 #include "noun.h"
 
-
 static u3_noun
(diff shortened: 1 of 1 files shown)


joemfb (approved):
this will do

PR #454, opened 2023-06-14
fine: Loosen path parsing and bump page size

philipcmonk:
These are the vere changes that accompany urbit/urbit#6669, see that for a description.

### DIFF
--- a/pkg/vere/io/ames.c
+++ b/pkg/vere/io/ames.c
@@ -5,7 +5,7 @@
 #include "noun.h"
 #include "ur.h"
 
-#define FINE_PAGE       512             //  packets per page
+#define FINE_PAGE      4096             //  packets per page
 #define FINE_FRAG      1024             //  bytes per fragment packet
 #define FINE_PATH_MAX   384             //  longest allowed scry path
 #define HEAD_SIZE         4             //  header size in bytes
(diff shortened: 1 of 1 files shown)


philipcmonk:
I've marked this as ready for review, and updated the description of urbit/urbit#6669 to describe the changes there and here.  While they do not need to be released in lockstep, it's better to understand these changes in that context.

pkg/vere/io/ames.c
```diff
@@ -5,7 +5,7 @@
 #include "noun.h"
 #include "ur.h"
 
-#define FINE_PAGE       512             //  packets per page
+#define FINE_PAGE      4096             //  packets per page
 #define FINE_FRAG      1024             //  bytes per fragment packet
```
ashelkovnykov, line 9: What is a Fine page? Is it the maximal file size? Just trying to figure out to which change described in Arvo 6669 this corresponds.

philipcmonk: It's the max number of packets we read out in a scry result. Eg there's a 15MB file in our clay history. Before, that would require about 30 scry requests into arvo, now it requires 4, and each one loads up to 4096 entries into the cache.

joemfb (approved):
this will do

PR #455, opened 2023-06-14
Transfer cap permissions to new binary

### LINKED ISSUE #312: `next` utility: previous setcap port 80/443 rule overwritten on linux
every time I upgrade the binary with `next` I need to re-run `sudo setcap 'cap_net_bind_service=+ep' <pier>/.run`

this is kinda annoying

is there a way for it to not do this?

mopfel-winrux:
`next` now transfers netcap permissions to new binary on Linux. We use magic numbers so not to include libcap. This fixed #312

### DIFF
--- a/pkg/vere/king.c
+++ b/pkg/vere/king.c
@@ -1204,12 +1204,13 @@ u3_king_vere(c3_c* pac_c,  // pace
 
   #if defined(U3_OS_linux)
 
-  	c3_w value[5];
-	  value[0] = 0x02000001;
-	  value[1] = 0x80000400;
-	  value[2] = 0;
-	  value[3] = 0;
-	  value[4] = 0;
+    // This runs `sudo setcap 'cap_net_bind_service,cap_setfcap=+ep'` on the new binary
+    c3_w value[5];
+    value[0] = 0x02000001;
+    value[1] = 0x80000400;
+    value[2] = 0;
+    value[3] = 0;
+    value[4] = 0;
     ret_i = setxattr(bin_c, "security.capability", value, 5*sizeof(c3_w), 0);
     if( -1 == ret_i ) {
       u3l_log("netcap permissions failed error: %s", strerror(errno));
--- a/pkg/vere/king.c
+++ b/pkg/vere/king.c
@@ -11,6 +11,7 @@
 #include "ur.h"
 #include "uv.h"
 #include "version.h"
+#include <sys/xattr.h>
 
 // It's unclear which file owns this variable, so put it here for now.
 u3_host u3_Host;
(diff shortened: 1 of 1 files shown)


pkg/vere/king.c
```diff
@@ -1201,6 +1202,23 @@ u3_king_vere(c3_c* pac_c,  // pace
...
+    c3_w value[5];
+    value[0] = 0x02000001;
+    value[1] = 0x80000400;
+    value[2] = 0;
+    value[3] = 0;
+    value[4] = 0;
+    ret_i = setxattr(bin_c, "security.capability", value, 5*sizeof(c3_w), 0);
+    if( -1 == ret_i ) {
+      u3l_log("netcap permissions failed error: %s", strerror(errno));
+    }
```
pkova, line 1217: I don't think we care enough to print if `setcap` fails. For users that don't care about binding low ports because they are behind a reverse proxy or something this can just silently fail.

pkg/vere/king.c
```diff
@@ -1201,6 +1202,23 @@ u3_king_vere(c3_c* pac_c,  // pace
...
 
+  #if defined(U3_OS_linux)
+
+    // This runs `sudo setcap 'cap_net_bind_service,cap_setfcap=+ep'` on the new binary
+    c3_w value[5];
+    value[0] = 0x02000001;
+    value[1] = 0x80000400;
+    value[2] = 0;
+    value[3] = 0;
+    value[4] = 0;
```
pkova, line 1213: I wonder if this works if we just remove these zeroes and do `c3_w value[2]`.

pkg/vere/king.c
```diff
@@ -1201,6 +1202,23 @@ u3_king_vere(c3_c* pac_c,  // pace
 
   u3l_log("vere: saved to %s", bin_c);
 
+  #if defined(U3_OS_linux)
+
+    // This runs `sudo setcap 'cap_net_bind_service,cap_setfcap=+ep'` on the new binary
```
joemfb, line 1207: Does this only work if you `sudo urbit dock $pier` or `sudo $pier/.run next`? And if not, how?

pkova: The plan is to have the docs say

`sudo setcap 'cap_net_bind_service,cap_setfcap=+ep' <pier>/.run` 

instead of

`sudo setcap 'cap_net_bind_service=+ep' <pier>/.run`

Then you only ever have to do it once.

joemfb: That's a lot of capability ... I wonder if it'd be better just recommend that people run as root if they don't want to deal with capabilities and/or proxy ports 80/443.

mopfel-winrux: I think either way is equally annoying. People tend to not like running software as root though

belisarius222:
Closing due to consensus that the security implications are too severe to be offset by the user experience win, which can be gotten in other ways.

PR #459, opened 2023-06-15
Epoch System

### LINKED ISSUE #313: `pier`: epoch system

matthew-levan:
This PR implements a new format for how piers store their event logs on disk. 

Resolves #313.

### Design

Existing format:
```
./zod/.urb/log
├── data.mdb
└── lock.mdb
```

New format:
```
./zod/.urb/log
├── 0i0             # epoch dirnames specify the last event of the previous epoch
│   ├── data.mdb    # lmdb file containing events 1-132
│   ├── epoc.txt    # disk format version (this PR starts versioning at 1)
│   ├── lock.mdb    # lmdb lock file
│   └── vere.txt    # binary version this set of events was originally run with
└── 0i132
    ├── data.mdb
    ├── epoc.txt
    ├── lock.mdb
    ├── north.bin   #
    ├── south.bin   # snapshot files (state as of event 132), strictly read-only
    └── vere.txt
```

The new format introduces *epochs*, which are simply "slices" or "chunks" of a ship's complete event log. Above, you can see the ship's event log chunked into two epochs: `0i0` and `0i132`. 

New ships booted with the code in this PR instantiate their `log` directories with the new format. Existing piers are automatically migrated on boot.

Epoch "rollovers" (when the current epoch is ended and a new, empty epoch is created) occur under three conditions:
1. The pilot uses the new `roll` subcommand to manually rollover.
2. The pilot runs the `chop` subcommand.
3. We detect a different running binary version than the one pinned in the current epoch.

Both migrations and epoch rollovers ensure there's a current snapshot before running.

A few TODOs left:
- [x] Iron out small kink in migration behavior for previously chopped piers
- [x] Make sure correct binary version gets pinned to first epoch of migrated piers
- [x] Rollover to new epoch when a new binary version is detected
- [x] Make sure manual migration logic is idempotent
- [x] ~~Update `prep` command~~
- [x] Fix `chop` so it works when there are 3 epochs starting with `0i0`
- [x] ~~Reproduce and fix partially-deleted epoch `0i0` after `chop`~~
- [x] Pair with someone to run manual GDB testing for migration idempotency and rollover logic
- [x] Take a look at @joemfb's replay code and compare/find overlaps
- [x] Document final system design in this PR
- [x] Correct epoch naming scheme
- [x] Make `chop` leave the latest two epochs
- [x] Better error handling
- [x] Better cleanup
- [x] Test migration with real ships running on local-networking mode
- [x] Test epoch rollover idempotency
- [x] Test fresh boot
- [x] Handle case where snapshot has been deleted from `chk/`
- [x] Ensure `u3_disk_epoc_good()` is implemented and used how we want
- [x] Ensure `u3_disk_epoc_init()` is implemented and used how we want
- [x] Replay works with `urbit play` and `urbit`
- [x] Replay works in edge case where only epoch 0 and no valid snapshot exist
- [x] Move new-epoch-on-vere-version-mismatch logic to `_pier_wyrd_init()` 
- [x] Make subcommands which call `u3_disk_init()` auto-migrate
  - [x] `info`
  - [x] `cram`
  - [x] `queu`
  - [x] `meld`
  - [x] `pack`
  - [x] `play`
  - [x] `chop`
  - [x] `roll`
- [x] Make replay on boot use `u3_mars_play()`
- [x] Test migration from an old pier (again)
- [x] Test migration from an old pier that needs a full replay (i.e., from beginning of its event log) first works
- [x] Test that `./urbit roll zod` with an updated binary version *and* an empty latest epoch, it does not roll but instead just updates the `vere.txt` file


joemfb:
@matthew-levan this branch has conflicts with `develop`.

matthew-levan:
Thanks, things move fast! Just merged `develop` again.

pkg/noun/events.c
```diff
@@ -398,23 +398,41 @@ _ce_ephemeral_open(c3_i* eph_i)
...
+        if ( 0 != mkdir(pax_c, 0777) && EEXIST != errno ) {
+          fprintf(stderr, "loom: mkdir %s: %s\r\n", pax_c, strerror(errno));
+          return c3n;
+        }
+
+        *nxt_c = '/';
+        cur_c = nxt_c + 1;
+      }
+    }
+  }
```
barter-simsum, line 430: wtf is this? cur_c starts at the *end* of the string and thus this while loop doesn't do anything useful. I guess you're trying to do the equivalent of `mkdir -p /my/new/path` ?

barter-simsum: moreover, since we're in .urb, I don't think we should be doing this at all

pkg/noun/events.c
```diff
@@ -1345,54 +1363,73 @@ _ce_image_copy(u3e_image* fom_u, u3e_image* tou_u)
   return c3y;
 }
 
-/* u3e_backup(): copy snapshot to .urb/bhk (if it doesn't exist yet).
+/* u3e_backup(): copy snapshot from [pux_c] to [pax_c], 
+ * overwriting optionally. note that image files must 
+ * be named "north" and "south".
 */
```
barter-simsum, line 1369: cleanup trailing whitespace

pkg/noun/events.h
```diff
@@ -74,10 +74,11 @@
 
   /** Functions.
   **/
-    /* u3e_backup(): copy the snapshot from chk to bhk. 
+    /* u3e_backup(): copy the snapshot from [pux_c] to [pax_c], 
+     * overwriting optional.
```
barter-simsum, line 78: trailing whitespace

pkg/vere/main.c
```diff
@@ -2231,20 +2232,41 @@ _cw_play(c3_i argc, c3_c* argv[])
...
+  //     a. if it is valid and up-to-date with the log
+  //        i. if we executed via `urbit play`, exit 0
+  //        ii. if we executed via `urbit`, boot into live mode
+  //     b. if it is invalid, try loading a snapshot from
+  //        i. the snapshot in the latest epoch
+  //           1. if it is valid, replay the events and save the resulting
+  //              snapshot into chk/, exit 0 or boot into live mode as before
+  //           2. if it is invalid, return an error that informs the user of 
+  //              his need for a serious surgical operation
+  //
```
barter-simsum, line 2255: more trailing whitespace. Haven't highlighted all instances either. Please remove this. Maybe look into this? https://stackoverflow.com/questions/591923/make-git-automatically-remove-trailing-white-space-before-committing I don't personally use it, but my editor doesn't make it easy to leave trailing whitespace, so it's never been an issue for me

barter-simsum (changes requested):
took a quick look. Not quite finished yet though

pkg/noun/events.c
matthew-levan: Removed this entirely as it is unnecessary, thank you.

pkg/noun/events.c
```diff
@@ -398,23 +398,31 @@ _ce_ephemeral_open(c3_i* eph_i)
...
+    while ( dir_c < end_c ) {
+      if ( '/' == *dir_c ) {
+        *dir_c = 0;
+        c3_mkdir(ful_c, 0700);
+        *dir_c = '/';
+      }
+      dir_c++;
+    }
+    c3_mkdir(ful_c, 0700);
+  }
```
barter-simsum, line 420: compared to the previous change, this should correctly create the parent directories, but I think it's wholly unnecessary and also this "traverse path and replace / with \0, mkdir path, re-replace \0 with /" is just kinda janky and more generalized than we need.

rather than call _ce_image_open on a path + file that may not exist, we should just ensure the required parent directories already exist.

Those would be "./urb/log" (this necessarily must already exist) and "./urb/log/0iN" -- this WILL need to be created. An explicit mkdir 0iN would certainly be more readable.

barter-simsum: i.e. *retain* the behavior of creating images (files) that do not exist, but don't add extra behavior that creates parent directories if those also do not exist.

pkg/vere/disk.c
```diff
@@ -1019,66 +1031,415 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+    //  initialize dun_d/sen_d values
+    if ( 0 == las_d ) {               //  fresh epoch (no events in lmdb yet)
+      if ( 0 == lat_d ) {             //  first epoch
+        log_u->dun_d = 0;
+      } else {                        //  not first epoch
+        log_u->dun_d = lat_d;         //  set dun_d to last event in prev epoch
+      }
+    } else {                          //  not fresh epoch
+      log_u->dun_d = las_d;           //  set dun_d to last event in lmdb
+    }
```
barter-simsum, line 1089: The extra `if (0 == lat_d)` is unnecessary.

more concisely:
```
if (las_d != 0) log_u->dun_d = las_d;
else log_u->dun_d = lat_d;
```

pkg/vere/disk.c
```diff
@@ -1019,66 +1031,415 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+   *     c. creates binary version file
+   *     d. initializes database
+   *     e. reads metadata from old database
+   *     f. writes metadata to new database
+   *     g. loads new epoch directory and sets it in log_u
+   *  2. create hard links to data.mdb and lock.mdb in 0i0/
+   *  3. rollover to new epoch (second call to u3_disk_epoc_init()) *
+   *     a. same as 1a-g but also copies current snapshot between c/d steps
+   *  4. delete backup snapshot (c3_unlink() and c3_rmdir() calls) *
+   *  5. delete old data.mdb and lock.mdb files (c3_unlink() calls) *
```
barter-simsum, line 1327: The "(* indicates breakpoint set" and "*" instances, you'll want to probably remove since they were for debugging purposes

pkg/vere/foil.c
```diff
@@ -106,13 +106,17 @@ u3_foil_folder(const c3_c* pax_c)
...
   */
   while ( UV_EOF != uv_fs_scandir_next(&ruq_u, &den_u) ) {
     if ( UV_DIRENT_FILE == den_u.type ) {
       u3_dent* det_u = u3_dent_init(den_u.name);
       det_u->nex_u = dir_u->all_u;
       dir_u->all_u = det_u;
+    } else if ( UV_DIRENT_DIR == den_u.type ) {
+      u3_dent* det_u = u3_dent_init(den_u.name);
+      det_u->nex_u = dir_u->dil_u;
+      dir_u->dil_u = det_u;
```
barter-simsum, line 119: why is this necessary? I can't tell

pkg/noun/events.c
matthew-levan: Also `.urb/chk`, which is actually the only time we actually create the directories in `_ce_image_open` at the moment.

matthew-levan: Fair enough proposal to require the caller to ensure the parent directories exist before calling `_ce_image_open`-- I will implement it today.

pkg/vere/foil.c
matthew-levan: So this just adds a list of directories (`dil_u`) to the `u3_dire` type:
```
typedef struct _u3_dire {
  c3_c*    pax_c;                     //  path of directory
  uv_file  fil_u;                     //  file, opened read-only to fsync
  u3_dent* all_u;                     //  file list
  u3_dent* dil_u;                     //  directory list
} u3_dire;
```

matthew-levan: I need `dil_u` to work with epoch directories.

barter-simsum:
test

.gitignore
matthew-levan, line 1: This would be helpful for a lot of people (including myself), but obviously isn't a necessary part of this PR. Figured I could sneak it in.

pkg/vere/main.c
```diff
@@ -1925,11 +1924,11 @@ _cw_meld(c3_i argc, c3_c* argv[])
     exit(1);
   }
 
+  u3_Host.eve_d = u3m_boot(u3_Host.dir_c, (size_t)1 << u3_Host.ops_u.lom_y);
   u3_disk* log_u = _cw_disk_init(u3_Host.dir_c); // XX s/b try_aquire lock
   c3_w     pre_w;
 
   u3C.wag_w |= u3o_hashless;
```
joemfb, line 1931: This needs to be before the `u3m_boot()` call

pkg/vere/main.c
```diff
@@ -2213,6 +2245,7 @@ _cw_play(c3_i argc, c3_c* argv[])
 
   //  XX handle SIGTSTP so that the lockfile is not orphaned?
   //
+  u3_Host.eve_d = u3m_boot(u3_Host.dir_c, (size_t)1 << u3_Host.ops_u.lom_y);
```
joemfb, line 2248: This won't restore the right snapshot on a full replay, or on one where `chk` is empty or missing.

pkg/vere/main.c
```diff
@@ -342,7 +342,7 @@ _main_getopt(c3_i argc, c3_c** argv)
           return c3n;
         } else {
           u3_Host.ops_u.sap_w = arg_w * 60;
-          if ( 0 == u3_Host.ops_u.sap_w) 
+          if ( 0 == u3_Host.ops_u.sap_w)
```
joemfb, line 345: s/b space before close-paren

pkg/vere/main.c
```diff
@@ -1196,7 +1197,7 @@ _cw_disk_init(c3_c* dir_c)
   u3_disk*  log_u = u3_disk_init(dir_c, cb_u);
 
   if ( !log_u ) {
-    fprintf(stderr, "unable to open event log\n");
+    fprintf(stderr, "unable to open event log\n");;
```
joemfb, line 1200: extra `;`

joemfb (changes requested):

belisarius222:
CI failure on linux x86_64: https://pastebin.com/nUAQu4Ap

belisarius222:
@joemfb PTAL.  @matthew-levan says it's ready for (hopefully) final review.

pkg/noun/events.c
```diff
@@ -1345,54 +1336,75 @@ _ce_image_copy(u3e_image* fom_u, u3e_image* tou_u)
...
-  c3_i mod_i = O_RDWR | O_CREAT; // XX O_TRUNC ?
-  c3_c ful_c[8193];
+  //  source image files from [pux_c]
+  u3e_image nux_u = { .nam_c = "north" };
+  u3e_image sux_u = { .nam_c = "south" };
+
+  //  destination image files to [pax_c]
+  u3e_image nax_u = { .nam_c = "north" };
+  u3e_image sax_u = { .nam_c = "south" };
+
```
barter-simsum, line 1353: It looks like .pgs_w was originally zeroed. Can you please just append `,0` so the remaining members are explicitly zeroed?

pkg/vere/disk.c
```diff
@@ -1019,66 +1031,450 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
-      #endif
-
-      if ( 0 == (log_u->mdb_u = u3_lmdb_init(log_c, siz_i)) ) {
-        fprintf(stderr, "disk: failed to initialize database\r\n");
-        c3_free(log_c);
-        c3_free(log_u);
-        return 0;
+    if ( c3y == u3_disk_need_migrate(log_u) ) {
+      if ( c3y == mig_o ) {
+        if ( c3n == u3_disk_migrate(log_u) ) {
```
barter-simsum, line 1048: I would simplify the 3 deep condition using `&&` if necessary

pkg/vere/disk.c
```diff
@@ -1019,66 +1031,450 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+/* u3_disk_epoc_vere: get binary version from epoch.
+*/
+c3_o
+u3_disk_epoc_vere(u3_disk* log_u, c3_d epo_d, c3_c* ver_w)
+{
+  struct stat buf_u;
+  c3_c* ver_c;
+  c3_w red_w, len_w;
+  c3_i ret_i, fid_i;
+  ret_i = asprintf(&ver_c, "%s/0i%" PRIc3_d "/vere.txt",
```
barter-simsum, line 1273: should vere.txt be called something else?

barter-simsum (commented):
quick pass over. a few nits mostly.

pkg/vere/disk.c
barter-simsum: but why add .txt? I would just call it vere

matthew-levan:
I think it's worth having the file extension to keep it separate from other
instances of "vere" in the codebase.

pkg/vere/disk.c
barter-simsum: oh i suppose epoc.txt does this as well. ok no prob

pkg/vere/disk.c
```diff
@@ -1019,66 +1031,448 @@ u3_disk_init(c3_c* pax_c, u3_disk_cb cb_u)
...
+   *     e. reads metadata from old database
+   *     f. writes metadata to new database
+   *     g. loads new epoch directory and sets it in log_u
+   *  2. create hard links to data.mdb and lock.mdb in 0i0/
+   *  3. rollover to new epoch (second call to u3_disk_epoc_init())
+   *     a. same as 1a-g but also copies current snapshot between c/d steps
+   *  4. delete backup snapshot (c3_unlink() and c3_rmdir() calls)
+   *  5. delete old data.mdb and lock.mdb files (c3_unlink() calls)
+   */
+
```
barter-simsum, line 1343: I would add a basic printf here to indicate a migration is started to more easily triage crashes as was done in u3m_migrate -- along with any additional printfs preceding concrete steps of the migration that might fail for some reason.

barter-simsum (approved):
PMTS,N

barter-simsum:
@joemfb to reapprove

matthew-levan:
*Ravioli, ravioli, give me the formuoli!*

pkg/vere/main.c
```diff
@@ -2105,6 +2105,38 @@ _cw_play_slog(u3_noun hod)
...
+/* _cw_play_snap(): prepare snapshot for full replay.
+*/
+static void
+_cw_play_snap(u3_disk* log_u)
+{
+  c3_c chk_c[8193], epo_c[8193];
+  snprintf(chk_c, 8193, "%s/.urb/chk", u3_Host.dir_c);
+  snprintf(epo_c, 8192, "%s/0i%" PRIc3_d, log_u->com_u->pax_c, log_u->epo_d);
+
+  if ( 0 == log_u->epo_d ) {
```
joemfb, line 2117: Does this condition also handle full replay before migration?

pkg/vere/main.c
matthew-levan: Thanks, done.

barter-simsum (approved):

joemfb (approved):
this will do

PR #468, opened 2023-06-22
u3: improve snapshot corruption size checks

joemfb:
These checks were introduced in v2.7, aborting the process if the snapshot metadata indicated that truncation had occurred. But the check as written is unnecessarily strict, and also aborts the process if the snapshot was larger than necessary. This PR prints a warning in that case and otherwise continues.

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -617,15 +617,18 @@ _find_home(void)
     nor_w = (low_p + ((1 << u3a_page) - 1)) >> u3a_page;
     sou_w = u3P.pag_w - (hig_p >> u3a_page);
 
-    if ( (nor_w > u3P.nor_u.pgs_w) || (sou_w > u3P.sou_u.pgs_w) ) {
+    if ( (nor_w > u3P.nor_u.pgs_w) || (sou_w != u3P.sou_u.pgs_w) ) {
       fprintf(stderr, "loom: corrupt size north (%u, %u) south (%u, %u)\r\n",
                       nor_w, u3P.nor_u.pgs_w, sou_w, u3P.sou_u.pgs_w);
       u3_assert(!"loom: corrupt size");
     }
 
-    if ( (nor_w < u3P.nor_u.pgs_w) || (sou_w < u3P.sou_u.pgs_w) ) {
-      fprintf(stderr, "loom: strange size north (%u, %u) south (%u, %u)\r\n",
-                      nor_w, u3P.nor_u.pgs_w, sou_w, u3P.sou_u.pgs_w);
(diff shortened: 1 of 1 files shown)


pkg/noun/manage.c
```diff
@@ -617,11 +617,16 @@ _find_home(void)
...
     sou_w = u3P.pag_w - (hig_p >> u3a_page);
 
-    if ( (nor_w != u3P.nor_u.pgs_w) || (sou_w != u3P.sou_u.pgs_w) ) {
+    if ( (nor_w > u3P.nor_u.pgs_w) || (sou_w > u3P.sou_u.pgs_w) ) {
       fprintf(stderr, "loom: corrupt size north (%u, %u) south (%u, %u)\r\n",
                       nor_w, u3P.nor_u.pgs_w, sou_w, u3P.sou_u.pgs_w);
       u3_assert(!"loom: corrupt size");
     }
+
+    if ( (nor_w < u3P.nor_u.pgs_w) || (sou_w < u3P.sou_u.pgs_w) ) {
```
barter-simsum, line 626: means: (north|south) road are smaller than the (north|south) snapshot. Why is that not an error? When should the size of a road's snapshot be allowed to exceed the road itself when applying it?

joemfb: It's strange, and shouldn't happen, so we print a warning. But it's possible that the snapshots are otherwise correct, and we don't want to prevent their restoration in that case.

(As i'm writing this, I'm realizing that's only true for north.bin -- south.bin is reversed on disk and always 0-1 pages, so any mismatch will be fatal. I'll revise)

pkg/noun/manage.c
```diff
@@ -617,11 +617,19 @@ _find_home(void)
     nor_w = (low_p + ((1 << u3a_page) - 1)) >> u3a_page;
```
barter-simsum, line 617: One more thing, though it's not exactly particular to this PR. Do snapshots include the extra 2 zero words at the beginning of the memory arena (for alignment), or are they included in the snapshot? If they're not included, this isn't exactly correct, since nor_w is just hat_p in this case (and not the north road length).

barter-simsum: -- looks like they do include the leading zeroes? Just want to confirm...

joemfb: The pages are aligned in memory and on disk, so the initial padding words *are* included. `low_p` is `u3R->hat_p - 1;` (the home road is always north); `nor_w` is the number of pages in the north segment (rounded up).

barter-simsum (dismissed):
J,TLPGTMAHMA

barter-simsum (approved):

PR #469, opened 2023-06-22
add -flto=thin to mac aarch64 build cflags for ez 10% gain

fighet-parnet:
Adds somewhere between 5-10% improvement to wall time.

I've ran all tests, booted several fakeships and comets with this installed, and all seems to work properly.

Mac x86 is coming, I just don't have access to an intel mac right now [but soon...].
Linux will come later once I figure out some bugs.

The following is time to boot a fakeship from a brass pill:
```
Without LTO:
________________________________________________________
Executed in  180.46 secs    fish           external
usr time   17.83 secs    0.10 millis   17.83 secs
sys time    0.23 secs    1.86 millis    0.23 secs


With LTO:
________________________________________________________
Executed in  164.87 secs    fish           external
usr time   15.65 secs    0.12 millis   15.65 secs
sys time    0.19 secs    1.94 millis    0.19 secs
```
180/164 -> 1.09

---

~barter-simsum

## Brass pill boot:

x86 linux without lto: ~140s
x86 linux with    lto: ~128s

~ 8.6% improvement - note we may be able to squeeze a bit more out if we can
apply `-flto` to all dependencies and not just urbit binaries. This is currently
an issue for x86 linux though due to some weird uninvestigated behavior with
libsigsegv

## For those curious, the following hints at what was inlined

```
readelf -s ./bazel-bin/pkg/vere/urbit | grep lto_priv
  1233: 000000000042d0f0   361 FUNC    LOCAL  DEFAULT    2 _rebalance.lto_priv.1
  1237: 000000000042d6c0   435 FUNC    LOCAL  DEFAULT    2 _rebalance.lto_priv.0
 13002: 0000000000d24a00   200 OBJECT  GLOBAL HIDDEN    14 u3_Signal.lto_priv.0
 13358: 0000000000d36de0     8 OBJECT  GLOBAL HIDDEN    14 sec_u.lto_priv.0
 13429: 0000000000d24ae0    80 OBJECT  GLOBAL HIDDEN    14 u3V.lto_priv.0
 13785: 00000000004516e0   110 FUNC    GLOBAL HIDDEN     2 _tap_in.lto_priv.0
 14094: 0000000000d3aeb0     4 OBJECT  GLOBAL HIDDEN    14 sag_w.lto_priv.0
 14100: 0000000000469af0   971 FUNC    GLOBAL HIDDEN     2 _lord_stop.lto_priv.0
[...]

PR #471, opened 2023-06-26
factor release/debug dispatch logic out of .bazelrc

### LINKED ISSUE #472: Factor release/debug dispatch logic out of .bazelrc
We "dispatch" on a number of things in the .bazelrc and it's becoming very
tedious to make further changes (e.g. adding -flto to only release builds, but
changing the form of the argument between macos/linux).

We should minimally factor release/debug dispatch into two new cc_binary targets:
- urbit-rel
- urbit-dbg

Sample commands:

build urbit release binary:
```
bazel build :urbit
```
build urbit debug binary with default defines:
```
bazel build :urbit-dbg
```
build urbit debug binary with memory + cpu debug:
```
bazel build --config=mem_dbg --config=cpu_dbg :urbit-dbg
```

So essentially, instead of

```
bazel build [...] --config=dbg :urbit
```

we

```
bazel build [...] :urbit-dbg
```

barter-simsum:
move release/dbg dispatch logic out of .bazelrc and into separate cc_binary targets

Resolves #472 

Draft for now. Still need to:
[] update docs

[] confirm no changes to macos/linux release+debug builds

[] Fake ship tests are running about 3x slower due to removal of `build
--host_copt='-O3'` from .bazelrc. Unclear why as all third party dependencies
specify an optimization level in their cc_library targets and vere now specifies
it for both vere-rel and vere-dbg. -- Oh duh, probably specifying a copt in
bazel with `--copt`/`--host_copt` takes precedence over a `copts` "rule". Should
we just be building every dependency at -O3 then??? -- Answer: the other
cc_library targets in vere should also specify copts of -O3 for rel and -O0 for
dbg... which complicates things. By providing these in .bazelrc, it's easy to
send them to all targets in pkg/...

... may also try getting this stuff out of the .bazelrc file:

```
build:mem_dbg --per_file_copt='pkg/.*@-DU3_MEMORY_DEBUG'
build:cpu_dbg --per_file_copt='pkg/.*@-DU3_CPU_DEBUG'
build:snp_dbg --per_file_copt='pkg/.*@-DU3_SNAPSHOT_VALIDATION'
```

This sort of thing in other builds systems is just done by defining an env var
like:

```
MEM= CPU= bazel build :urbit
```

But I don't think this is possible in bazel. What we have now _may_ be the
simplest solution

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,16 +20,7 @@ build --@io_bazel_rules_docker//transitions:enable=false
 build --flag_alias=clang_version=//:clang_version
 build --flag_alias=gcc_version=//:gcc_version
 
-# Don't include source level debug info on macOS. See
-# https://github.com/urbit/urbit/issues/5561 and
-# https://github.com/urbit/vere/issues/131.
-build:linux --per_file_copt='pkg/.*@-g'
-build:linux --host_copt='-g'
 build --strip=never
-
-# Use -O3 as the default optimization level.
-build --per_file_copt='pkg/.*@-O3'
-build --host_copt='-O3'
 
 # Turn on CPU and memory debug for exec config, which we only use to run the
 # fake ship tests.  Also turn on extra snapshot validation.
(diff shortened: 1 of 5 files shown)


barter-simsum:
will probably close in favor of #469

fighet-parnet:
This can be closed now.

PR #479, opened 2023-06-29
u3: move the guard page earlier when making roads

joemfb:
... preventing spurious `loom: ward bogus` assertion failures. The guard page was being moved if necessary after initializing an inner road (accounting for road padding), but it could still be hit while initializing the road struct itself, in which case it would trigger an guard-page sanity assertion.

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -518,7 +518,7 @@ _pave_road(c3_w* rut_w, c3_w* mat_w, c3_w* cap_w, c3_w siz_w)
    len_w - size of your loom in words
 */
 static u3_road*
-_pave_north(c3_w* mem_w, c3_w siz_w, c3_w len_w)
+_pave_north(c3_w* mem_w, c3_w siz_w, c3_w len_w, c3_o kid_o)
 {
   //  in a north road, the heap is low and the stack is high
   //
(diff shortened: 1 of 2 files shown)


pkg/noun/manage.c
```diff
@@ -901,7 +915,6 @@ u3m_leap(c3_w pad_w)
   */
   {
     u3R = rod_u;
-    u3m_ward();
     _pave_parts();
```
barter-simsum, line 918: So, if I'm reading this right, we're no longer calling u3m_ward at all on leap beacuse of its dependency on u3R which may be unintialized. Instead, we have a call to u3e_ward in _pave_north. But there's no ward call in _pave_south. Shouldn't there be one?

joemfb: This is about calling ward *before* initializing the inner road, not after. Otherwise, the stores in `_pave_road()` can hit the guard page without the road boundaries having been updated.

joemfb:
@mopfel-winrux still hits the assertion in question when running this PR, I will review and revise.

joemfb:
[14be92e](https://github.com/urbit/vere/pull/479/commits/14be92ef3586319da581ef406c8bfccde710070f) completes the fix for the bogus assertion. The fact that our bounds check in `u3e_ward()` was incorrect wrt to the south segment means that it was still possible to jump the guard page if it were adjacent to that side of our road, which means that this issue could also cause heap/stack corruption. That would usually just crash the process, but it's possible that it could escape to the home road and affect persistent state. This issue was introduced in vere-v2.7, and needs to be released in a hotfix.

barter-simsum (approved):
LGTM

PR #481, opened 2023-06-30
ci: fix error check in version upload

joemfb:
It looks like this has been broken since the repository split. The version upload works, but the error check does not.


matthew-levan:
Yes I've known about this for a long time, thanks for fixing it. Approved. Let's wait until passing CI to merge just in case.

matthew-levan (approved):

barter-simsum (approved):
IA

joemfb:
It won't run in this PR build, only in the push to `develop`.

joemfb:
worked: https://github.com/urbit/vere/actions/runs/5425664779/jobs/9867110484#step:5:58

PR #496, opened 2023-07-20
UIP-0103: Persistent Nock Caching

matthew-levan:
Implements [UIP-0103](https://github.com/urbit/UIPs/blob/main/UIPS/UIP-0103.md). 

TODO: Write a roadmap for merging this. What else has to be done?
- [x] Loom migration
- [x] Affordance for future specialized caches in `u3a_road` struct
- [x] PRs for use cases ready in Arvo
  - [x] urbit/urbit#6770
  - [x] urbit/urbit#6773
  - [x] ~~Ford Lightning~~
- [x] Manage new persistent state like we manage the rest of our persistent state (`|mass` breaks before we do this)
- [x] Update command-line args for cache sizes
- [x] Determine why `bazel build //pkg/vere:test-fake-ship` seems to hang indefinitely (I ran it for ~16h on my own dedicated box and it still didn't finish)

Tested working in all vanes I tested (Iris, etc. are a little more cumbersome):
- [x] Ames
- [x] Arvo
- [x] Behn
- [x] Clay
- [x] Dill
- [x] Eyre
- [x] Gall

### DIFF
--- a/pkg/noun/nock.c
+++ b/pkg/noun/nock.c
@@ -2634,7 +2634,8 @@ _n_burn(u3n_prog* pog_u, u3_noun bus, c3_ys mov, c3_ys off)
       o   = *top;
       if ( ( u3z_memo_toss == u3h(o) )
          ? ( &(u3H->rod_u) != u3R )
-         : ( 0 == u3R->ski.gul ) ) {  //  prevents userspace from persistence
+        //  : ( 0 == u3R->ski.gul ) ) {  //  prevents userspace from persistence
+         : ( 1 ) ) {  //  prevents userspace from persistence
         u3z_save_m(u3h(o), 144 + c3__nock, u3t(o), x);
       }
       else if ( ( u3z_memo_keep == u3h(o) ) &&
--- a/pkg/noun/nock.c
+++ b/pkg/noun/nock.c
@@ -619,7 +619,7 @@ _n_melt(u3_noun ops, c3_w* byc_w, c3_w* cal_w,
 
         case SKIB: case SLIB: {
           c3_l tot_l = 0,
-               sip_l = u3h(u3t(op));
+               sip_l = u3h(u3t(u3t(op)));
           c3_w j_w, k_w = i_w;
           for ( j_w = 0; j_w < sip_l; ++j_w ) {
             tot_l += siz_y[++k_w];
@@ -830,7 +830,9 @@ _n_prog_asm(u3_noun ops, u3n_prog* pog_u, u3_noun sip)
           _n_prog_asm_inx(buf_y, &i_w, mem_s, cod);
           mem_u        = &(pog_u->mem_u.sot_u[mem_s++]);
           mem_u->sip_l = sip_l;
-          mem_u->key   = u3k(u3t(u3t(op)));
+          // [op_y, cid, mem_w, nef]
+          mem_u->key   = u3k(u3t(u3t(u3t(op))));
+          mem_u->cid   = u3h(u3t(op));
           break;
         }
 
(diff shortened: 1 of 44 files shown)


pkg/noun/zave.c
```diff
@@ -90,3 +92,13 @@ u3z_uniq(u3_noun som)
...
+
+/* u3z_reap(): promote memoization cache state.
+*/
+void
+u3z_reap(u3p(u3h_root) har_p)
+{
+  // u3h_walk_with(har_p, _cz_merge_memo_cb, &u3R->cax.har_p);
+  // NB *not* u3z_free; _cz_merge_memo_cb() transfers u3z_memo's
+  // u3h_free(har_p);
+}
```
joemfb, line 104: For a proof-of-concept, all you have to do here is

```
u3h_uni(u3R->cax.har_p, har_p);
u3h_free(har_p);
```

There's no need to touch bytecode programs.

pkg/noun/nock.c
```diff
@@ -1117,16 +1119,23 @@ _n_bint(u3_noun* ops, u3_noun hif, u3_noun nef, c3_o los_o, c3_o tel_o)
...
-        // memoizing code always loses TOS because SAVE needs [pro key]
         mem_w += _n_comp(&mem, nef, c3y, c3n);
         ++mem_w; _n_emit(&mem, SAVE);
 
         op_y   = (c3y == los_o) ? SLIB : SKIB; // overflows to SLIS / SKIS
-        ++tot_w; _n_emit(ops, u3nt(op_y, mem_w, u3k(nef)));
+        u3z_cid cid = u3z_memo_toss;
+        // skip over spot hints
+        // XX actually evaluate hod some day
+        while ( 11 == u3h(hod) && 1 == u3h(u3t(u3h(u3t(hod))))) {
```
matthew-levan, line 1132: @frodwith if you take a look at the Nock produced from `!=  !:  ~>  %memo.[42]  42` you can see that `%spot` hints are wrapped and nested outside/within our `%memo` hint. The `while` loop here is meant to "ignore" those so we can parse the clue correctly.

matthew-levan: Without this logic, using the new `%memo` format (i.e., with a clue other than `[1 0]`) inside most vanes (i.e., the ones that use `!:` calls that wrap our `%memo` hints) doesn't work as intended. I'm open to a better algorithm here, but the code as written did solve the issue when I tested the new `%memo` hint in the same vanes I previously mentioned.

frodwith: It would be nice if you also skipped static hints in addition to dynamic hints with a [1 ...] clue when trying to resolve it statically. It would also be totally fine to ignore this hint when the clue is dynamic (modulo static hints). It's always fine to ignore (recursively) safe static hints. 

As an aside, it might be nice to make this a utility function. I haven't tried to compile this, but as a beginning:

```
/* extract a constant from a formula,
   ignoring safe/static hints,
   doing no computation */
u3_weak u3r_skip(u3_noun fol) {
  while ( 1 ) {
    if ( c3y == u3du(fol) ) {
      /* we -could- recurse on head and tail ... */
      return u3_none;
    }
    else switch ( u3h(fol) ) {
      default:
        /* we -could- try to constant fold... */
        return u3_none;
      case 1:
        return u3k(u3t(fol));
      case 11: {
        /* but we're only skipping hints here. */
        u3_noun arg = u3t(fol),
                hod = u3h(arg);
        
        if ( c3y == u3du(hod) ) {
          u3_weak dug = u3r_skip(u3t(hod));
          if ( u3_none == dug ) {
            /* only recursively safe dynamic ones. */
            return u3_none;
          }
          u3z(dug);
        }
        fol = u3t(arg);
        continue;
      }
    }
  }
}
```

matthew-levan:
Benchmarking how persistent caching changes performance of scry type checks in Arvo:
|**scry**|**uncached**|**cached**|**boost**|
|--------|------------|----------|---------|
|`.^((list [@da duct]) %bx /=//=/debug/timers)`|ms/4.478|ms/3.142|1.42x|
|`.^(cord %gx /(scot %p our)/reel/(scot %da now)/service/noun)`|ms/3.161|ms/1.446|2.18x|
|`.^((list [binding:eyre duct action:eyre]) %e /=bindings=)`|ms/6.607|ms/4.362|1.51x|

Note that, due to the existing scry cache in Ames, I made sure that these are all measurements of "first time scry" executions. 

In other words, I used two fresh fake zods: the first to benchmark the three scry calls *without* scry type checking memoization, and the second *with* scry type checking memoization.

I also made sure that we were actually using the persistent cache with a simple `printf` in Vere.

Here is the code used in Arvo:
```
++  look
  ~/  %look
  |=  [rof=roof lyc=gang]
  ^-  root
  ~/  %in
  |=  [ref=* raw=*]
  ?~  pax=((soft path) raw)  ~
  ?~  mon=(de-omen u.pax)  ~
  ?~  dat=(rof lyc u.mon)  ~
  ?~  u.dat  [~ ~]
  =*  vax  q.u.u.dat
  ?.  =>  [ref=ref vax=p=p.vax hoon-version=hoon-version wa=wa worm=worm]
      ~>  %memo./arvo/look                  ::  with memoization
      ?&  ?=(^ ref)
          =(hoon-version -.ref)
          -:(~(nets wa *worm) +.ref p.vax)
      ==
    ~>(%slog.[0 leaf+"arvo: scry-lost"] ~)
  :: ?.  ?&  ?=(^ ref)                      ::  without memoization
  ::         =(hoon-version -.ref)
  ::         -:(~(nets wa *worm) +.ref p.vax)
  ::     ==
  ::   ~>(%slog.[0 leaf+"arvo: scry-lost"] ~)
  [~ ~ q.vax]
```

Are these results indicative of much gain? I'm not sure. Seems like it.

PR #508, opened 2023-09-01
UIP-0103: Persistent Nock Caching + Loom Migration Framework + Pointer Compression Migration Refactor

matthew-levan:
Implements [UIP-0103](https://github.com/urbit/UIPs/blob/main/UIPS/UIP-0103.md), establishes a pattern for implementing loom migrations, and refactors the pointer compression as a compile-time feature instead of a runtime one. 

See #496 for original work on UIP-0103.

TODO:
- [x] Loom migration
- [x] Affordance for future specialized caches in `u3a_road` struct
- [x] PRs for use cases ready in Arvo
  - [x] urbit/urbit#6770
  - [x] urbit/urbit#6773
  - [x] ~~Ford Lightning~~
- [x] Manage new persistent state like we manage the rest of our persistent state (`|mass` breaks before we do this)
- [x] Update command-line args for cache sizes
- [x] Determine why `bazel build //pkg/vere:test-fake-ship` seems to hang indefinitely (I ran it for ~16h on my own dedicated box and it still didn't finish)

Tested working in all vanes I tested (Iris, etc. are a little more cumbersome):
- [x] Ames
- [x] Arvo
- [x] Behn
- [x] Clay
- [x] Dill
- [x] Eyre
- [x] Gall


matthew-levan:
Notes from @barter-simsum:
* Consider defining a macro to generate versioned symbols, and rename `u3j_vX...`, etc. as `_mig_vX...`

Example of a versioned macro:
```
#define VERSIONED(x) _mig_v ## x
VERSIONED(_cj_free_hank) (params) { defn }
```

pkg/noun/allocate.h
```diff
@@ -374,69 +380,35 @@
...
-    return (c3_w *)u3a_to_ptr(som);
-  }
+#   define u3a_to_wtr(som)  ((c3_w *)u3a_to_ptr(som))
 
   /* u3a_to_pug(): set bit 31 of [off].
    */
-  inline c3_w u3a_to_pug(c3_w off) {
-    c3_dessert((off & u3C.walign_w-1) == 0);
-    return (off >> u3C.vits_w) | 0x80000000;
-  }
```
matthew-levan, line 432: Restore the inline functions that have `c3_dessert` calls.

pkg/noun/allocate.c
```diff
@@ -2170,6 +2146,7 @@ u3a_reclaim(void)
 /* u3a_rewrite_compact(): rewrite pointers in ad-hoc persistent road structures.
 */
 void
+// u3a_v1_rewrite_compact(u3a_v1_road *rod_u)
```
matthew-levan, line 2149: delet

pkg/noun/jets.c
```diff
@@ -1069,18 +1069,18 @@ _cj_prog(u3_weak loc, u3_noun fol)
 /* cj_hank_find(): find cached hook information, keyed by arbitrary
  *                 prefix and term cords. RETAIN.
  */
-static _cj_hank*
+static _cj_v2_hank*
```
matthew-levan, line 1072: Restore the original names.

pkg/noun/jets.h
```diff
@@ -3,6 +3,8 @@
 #ifndef U3_JETS_H
 #define U3_JETS_H
 
+#include "pkg/noun/v2/jets.h"
```
matthew-levan, line 6: Remove.

pkg/noun/nock.h
matthew-levan, line 1: Revert all these changes.

pkg/noun/v1/jets.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef U3_JETS_V1_H
+#define U3_JETS_V1_H
+
+#include "pkg/noun/jets.h"
+#include "pkg/noun/v2/jets.h"
+
+  /**  Aliases.
+  **/
+#     define u3j_v1_fink       u3j_fink
```
matthew-levan, line 11: Alias these to `v2` and add the aliases required in `v2`.

pkg/noun/v1/allocate.h
```diff
@@ -0,0 +1,61 @@
...
+  **/
+#     define u3R_v1         u3R_v2
+#     define u3a_v1_botox   u3a_v2_botox
+#     define u3a_v1_box     u3a_v2_box
+#     define u3a_v1_cell    u3a_v2_cell
+#     define u3a_v1_free    u3a_v2_free
+#     define u3a_v1_into    u3a_v2_into
+#     define u3a_v1_is_cat  u3a_v2_is_cat
+#     define u3a_v1_is_pom  u3a_is_pom
+#     define u3a_v1_wfree   u3a_v2_wfree
```
matthew-levan, line 17: The deallocation functions will need to be pulled in from the commit prior to pointer compressoin.

pkg/noun/v1/nock.c
```diff
@@ -0,0 +1,63 @@
...
+  //    Note that the hank cache *must* also be freed (in u3j_v1_reclaim())
+  //
+  u3n_v1_free();
+}
+
+/* _cn_prog_free(): free memory retained by program pog_u
+**              NB: copy of _cn_prog_free in pkg/noun/nock.c
+*/
+static void
+_cn_prog_free(u3n_v1_prog* pog_u)
```
matthew-levan, line 28: Add `_v1` for consistency.

matthew-levan: Copy and add the pointer recalculation logic from `_n_ream`: https://github.com/urbit/vere/blob/e98746d8856a4967eab79f716285ca71be62ed28/pkg/noun/nock.c#L2972

pkg/noun/v1/nock.h
```diff
@@ -0,0 +1,24 @@
...
+    */
+      void
+      u3n_v1_reclaim(void);
+
+    /* u3n_v1_free(): free bytecode cache.
+     */
+      void
+      u3n_v1_free(void);
+
+#endif /* ifndef U3_NOCK_V1_H */
```
matthew-levan, line 24: Restore end of file newlines.

pkg/noun/v3/manage.c
```diff
@@ -0,0 +1,111 @@
...
+  c3_w  suz_w = c3_wiseof(u3v_v2_home);
+  c3_w *mut_w = c3_align(mem_w + len_w - suz_w, u3a_v3_balign, C3_ALGLO);
+
+  //  old road
+  u3v_v2_home* hum_u = (u3v_v2_home*)mut_w;
+  u3a_v2_road* rud_u = &hum_u->rod_u;
+  size_t ruz_t = sizeof(u3a_v2_road);
+
+  //  set v2 globals
+  u3H_v2 = (void *)mut_w;
```
matthew-levan, line 35: This is kind of a hack because `u3[H,R]_v2` don't contain all the members in the current home/road struct versions (in the main `pkg/noun` folder), and yet we're casting them to the current struct versions anyways and using them to set the global pointers. The reason for this is that the `u3[j,n]_v2_reclaim` functions which we call on lines 42 and 43 are actually aliases to `u3[j,n]_reclaim` (the current versions of the functions). The current functions require the global `u3H` and `u3R` pointers to be set. It just _so happens_ that these functions don't touch any of the struct members which were added to the `v3` (i.e., latest) versions of the home and road structs, so effectively everything just works. If we wanted to avoid this sort of "lying" we'd have to make `v2` versions of all the functions that rely on those global pointers instead of using aliases.

pkg/noun/v2/manage.c
```diff
@@ -0,0 +1,177 @@
...
+  c3_w ver_w = *(u3_Loom + len_w);
+
+  u3_assert( U3V_VER1 == ver_w );
+
+  c3_w* mem_w = u3_Loom + 1;
+  c3_w  siz_w = c3_wiseof(u3v_v2_home);
+  c3_w* mat_w = (mem_w + len_w) - siz_w;
+
+  u3H_v2 = (void *)mat_w;
+  u3R_v2 = &u3H_v2->rod_u;
```
joemfb, line 161: This should set the v1 pointers. They happen to be aliased to v2, but that's an implementation detail. Conceptually, we're initializing a v1 loom here.

pkg/noun/v2/manage.c
```diff
@@ -0,0 +1,177 @@
...
+_cm_pack_rewrite(void)
+{
+  //  XX fix u3a_v2_rewrite* to support south roads
+  //
+  u3_assert( &(u3H_v2->rod_u) == u3R_v2 );
+
+  u3v_v2_rewrite_compact();
+  u3j_v2_rewrite_compact();
+  u3n_v2_rewrite_compact();
+  u3a_v2_rewrite_compact();
```
joemfb, line 26: Technically, all these rewrite_compact functions are v1->v2 migration functions. They're *not* the original rewrite_compact functions as they were in v2 (for pack). We might want to rename them, or at least add comments.

pkg/noun/v3/manage.c
joemfb: Setting v2 here is not a hack, we're initializing a v2 loom. The hack is below, setting `u3R` to alias `u3R_v2`. The reason we're doing that is that `u3{j,n}_v2_reclaim` are aliases, and need a current road pointer to function. It might be better to implement the v2 reclaim functions as wrappers around v3 functions that first set `u3R` -- that encodes the hack more precisely.

pkg/noun/v3/manage.c
```diff
@@ -0,0 +1,111 @@
...
+  //  place the new home over the old one
+  c3_w  siz_w = c3_wiseof(u3v_v3_home);
+  c3_w *mat_w = c3_align(mem_w + len_w - siz_w, u3a_v3_balign, C3_ALGLO);
+  memcpy(mat_w, &hom_u, sizeof(u3v_v3_home));
+
+  //  set globals
+  u3H_v3 = (void*)mat_w;
+  u3R_v3 = &u3H_v3->rod_u;
+  u3H = (u3v_home*)u3H_v3;
+  u3R = (u3a_road*)u3R_v3;
```
joemfb, line 105: This is another instance of the same kind of hack, setting `u3R` so that u3h_v3_new_cache` will work, since it's just an alias.

pkg/noun/v3/vortex.h
```diff
@@ -0,0 +1,35 @@
...
+      } u3v_v3_home;
+
+  /**  Functions.
+  **/
+    /* u3v_v3_rewrite_compact(): rewrite arvo kernel for compaction.
+    */
+      void
+      u3v_v3_rewrite_compact();
+
+#endif /* ifndef U3_VORTEX_V3_H */
```
joemfb, line 35: I think everything in v3 vortex.{c,h} can just be an alias.

pkg/noun/manage.c
```diff
@@ -597,14 +595,25 @@ static void
 _find_home(void)
 {
   c3_w ver_w = *(u3_Loom + u3C.wor_i - 1);
-  u3a_config_loom(ver_w);
+
+  switch ( ver_w ) {
+    case 1: u3m_v2_migrate();
+    case 2: u3m_v3_migrate();
+    case 3: break;
```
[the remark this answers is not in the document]
joemfb: Also, we have version symbols defined (`U3V_VER1`, &c), so we should use them in the case statements here.

joemfb (changes requested):

pkg/noun/v2/nock.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "pkg/noun/nock.h"
+
+#include "types.h"
+
+  /**  Aliases.
+  **/
+#     define  u3n_v2_memo     u3n_memo
+#     define  u3n_v2_prog     u3n_prog
+#     define  u3n_v2_reclaim  u3n_reclaim
```
joemfb, line 14: This also needs to fix up pointers for portability. Also, these v2 aliases should point to v3.

pkg/noun/v1/nock.c
```diff
@@ -0,0 +1,71 @@
...
+  for (dex_w = 0; dex_w < pog_u->reg_u.len_w; ++dex_w) {
+    u3j_v1_rite_lose(&(pog_u->reg_u.rit_u[dex_w]));
+  }
+
+  // fix up pointers for loom portability
+  pog_u->byc_u.ops_y = (c3_y*) ((void*) pog_u) + sizeof(u3n_v1_prog);
+  pog_u->lit_u.non   = (u3_noun*) (pog_u->byc_u.ops_y + pog_u->byc_u.len_w);
+  pog_u->mem_u.sot_u = (u3n_v1_memo*) (pog_u->lit_u.non + pog_u->lit_u.len_w);
+  pog_u->cal_u.sit_u = (u3j_v1_site*) (pog_u->mem_u.sot_u + pog_u->mem_u.len_w);
+  pog_u->reg_u.rit_u = (u3j_rite*) (pog_u->cal_u.sit_u + pog_u->cal_u.len_w);
```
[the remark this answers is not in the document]
matthew-levan: Hmm, doing this actually results in an external fault.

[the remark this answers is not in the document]
joemfb: Can you get a trace for the fault?

joemfb (approved):
this will do

PR #510, opened 2023-09-02
Jet for jub:by

xiphiness:
jub:by is a variant of jab:by, which applies a gate to a value at a certain key. It crashes if it's not there, and also does not allow it to be deleted. This variant allows a create, update, or delete operation with one gate call and one tree traversal. The gate passed to jub takes a unit and returns a unit. The unit it is passed is ~ if the kv pair does not exist, and `val if it does. The unit returned causes a deletion if it is ~, and inserts otherwise.

https://github.com/urbit/urbit/pull/6782 is the corresponding PR with the hoon.

I added a jet for jib, referenced in the PR by amsden these pair of PR's would replace. I haven't written the hoon for it yet. Maybe we don't want to add it, as jub is exhaustive and would only be an ergonomic addition.

### DIFF
--- a/pkg/noun/jets/d/by_jib.c
+++ b/pkg/noun/jets/d/by_jib.c
@@ -1,0 +1,129 @@
+/* j/4/by_jib.c
+**
+*/
+#include "jets/q.h"
+#include "jets/w.h"
+
+#include "noun.h"
+
+u3_noun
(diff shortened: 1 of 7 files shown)


ashelkovnykov:
@xiphiness This PR is pending a final decision on https://github.com/urbit/urbit/pull/6782, is that correct?

PR #511, opened 2023-09-04
mdns: add mdns for local network routing

pkova:
This PR implements [mDNS](https://en.wikipedia.org/wiki/Multicast_DNS) for vere, significantly improving Urbit routing inside local networks. On startup vere will now broadcast a service with the format `dinleb-rambep._ames._udp.local`. Vere will also listen for any services on the local network with the signature of `_ames._udp`. Whenever new services are found vere will send the lane to arvo. 

On macOS we can use [Bonjour](https://developer.apple.com/bonjour/) and don't need to pull down any dependencies at all. On Linux we can use [Avahi](https://www.avahi.org/), specifically the bonjour compatibility layer. The header we're using is [dns_sd.h](https://github.com/lathiat/avahi/blob/master/avahi-compat-libdns_sd/dns_sd.h), [documented here](https://developer.apple.com/library/archive/documentation/Networking/Conceptual/dns_discovery_api/Introduction.html#//apple_ref/doc/uid/TP40002475-SW1).

Tested on macos aarch64 and linux x86_64 so far. To try it out, build vere from this PR, start a fakezod and observe:

```
# macos
dns-sd -B _ames._udp

# linux
avahi-browse -r _ames._udp
```

You can also start another fake ship on the same network and do `|ames-verb %rcv %odd` to see your fakeships find eachother. For this second part you need the companion arvo PR, here's a pill that has it already:

```
./urbit -F zod -u https://urbit-foundation.s3.us-east-2.amazonaws.com/dear.pill
```


pkg/vere/vere.h
```diff
@@ -1238,6 +1244,12 @@
         u3_auto*
         u3_lick_io_init(u3_pier* pir_u);
 
+    /**  mdns, routing for the local network
+    **/
+      /* mdns_init(): initialize mdns.
+      */
+        void
+        mdns_init(uint16_t port, char* our, mdns_cb* cb, void* context);
```
pkova, line 1252: Where should I put this signature? `vere.h` seems wrong but don't really know where it should go.

pkova:
I wonder what we should do in the following scenario:

Someone boots a fresh ship on your local network. You receive the mDNS notification and receive a `%dear` to arvo, but since this is a fresh boot you won't have `peer-state` for the ship yet and the lane will get dropped on the ground. You won't receive a mDNS notification about this lane until either ship restarts vere.

pkg/vere/io/ames.c
```diff
@@ -2060,6 +2060,50 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+  if (u3_nul == shp_u) {
+    u3l_log("ames: strange ship from mdns: %s", ship);
+    return;
+  }
+
+  u3_noun our = u3i_chubs(2, sam_u->pir_u->who_d);
+  if (our == u3t(shp_u)) {
+    u3z(our);
+    return;
+  }
```
joemfb, line 2093: This early return leaks `shp_u`

pkg/vere/io/ames.c
```diff
@@ -2060,6 +2060,50 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+
+  u3z(our);
+
+  u3_noun wir = u3nc(c3__ames, u3_nul);
+  u3_noun cad = u3nt(c3__dear, u3k(u3t(shp_u)), u3nc(c3n, u3_ames_encode_lane(lan)));
+
+  u3_auto_peer(
+               u3_auto_plan(&sam_u->car_u,
+                            u3_ovum_init(0, c3__a, wir, cad)),
+               0, 0, _mdns_dear_bail);
```
joemfb, line 2103: We don't need to track or handle failures with these events.

pkg/vere/mdns.c
```diff
@@ -0,0 +1,220 @@
...
+
+  if (err != kDNSServiceErr_NoError) {
+    u3l_log("mdns: service register error %i", err);
+  } else {
+    u3l_log("mdns: %s registered on all interfaces", name);
+  }
+  uv_poll_stop(payload->poll);
+  c3_free(payload->poll);
+  c3_free(payload->our);
+  c3_free(payload);
```
joemfb, line 170: leave a comment explaining why we leak `sref`.

pkg/vere/mdns.c
```diff
@@ -0,0 +1,220 @@
...
+  #   endif
+
+  memmove(our, our+1, strlen(our)); // certain url parsers don't like the sig
+
+  mdns_payload* register_payload = (mdns_payload*)c3_malloc(sizeof(mdns_payload));
+
+  register_payload->cb = cb;
+  register_payload->context = context;
+  register_payload->our = our;
+
```
joemfb, line 187: It looks like you don't need to save `our` in `payload`. I would try removing that, and also just passing `our + 1` to `DNSServiceRegister()` (maybe after checking that `'~' = *our`).

pkg/vere/mdns.c
```diff
@@ -0,0 +1,220 @@
...
+#include "dns_sd.h"
+
+typedef struct _mdns_payload {
+  mdns_cb*      cb;
+  DNSServiceRef sref;
+  char*         who;
+  char*         our;
+  uint16_t      port;
+  uv_poll_t*    poll;
```
joemfb, line 13: It looks like `poll` always has the same lifetime as `payload`, so it can just be `uv_poll_t` (no `*`).

pkg/vere/mdns.c
```diff
@@ -0,0 +1,220 @@
...
+}
+
+static void dns_sd_cb(DNSServiceRef sref, mdns_payload* payload) {
+  int fd = DNSServiceRefSockFD(sref);
+  uv_loop_t* loop = uv_default_loop();
+  uv_poll_t* poll = (uv_poll_t *)c3_malloc(sizeof(uv_poll_t));
+  poll->data = (void *)sref;
+  payload->poll = poll;
+  uv_poll_init(loop, poll, fd);
+  uv_poll_start(poll, UV_READABLE, poll_cb);
```
joemfb, line 111: If `poll` has the same lifetime as `payload` (see above)`, you can remove the malloc here and just `uv_poll_init(loop, &(payload->poll), fd);`

pkg/vere/io/ames.c
pkova: I get real ugly crashes on arvo versions that don't support `%dear` unless I handle the bail myself.

pkg/vere/mdns.c
pkova: Done

pkg/vere/mdns.c
pkova: Done

pkg/vere/mdns.c
pkova: Done

pkg/vere/mdns.c
pkova: And done.

pkg/vere/io/ames.c
```diff
@@ -2060,6 +2060,50 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+  u3l_log("mdns: %%dear failure;");
+  u3z(lud);
+  u3_ovum_free(egg_u);
+}
+
+/* _ames_put_dear(): send lane to arvo after hearing mdns response
+*/
+static void
+_ames_put_dear(char* ship, unsigned long s_addr, uint16_t port, void* context)
+{
```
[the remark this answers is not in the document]
pkova: Done

PR #512, opened 2023-09-04
jet `+mate` arm

sidnym-ladrut:
These changes introduce a jet for the [`+mate` arm](https://developers.urbit.org/reference/hoon/stdlib/2a#mate), and were tested with the following commands:

```
> (mate ~ ~)
~
> (mate ~ `1)
[~ u=1]
> (mate `2 ~)
[~ u=2]
> (mate `3 `3)
[~ u=3]
> (mate `4 `5)
(error)
took ms/2.118
> ~>  %bout  (roll (gulf 1 1.000) |=([n=@ a=(unit @)] (mate (mate `n ~) (mate ~ `n))))
[~ 1.000]
> =hoon-mate |*([a=(unit) b=(unit)] ?~(b a ?~(a b ?.(=(u.a u.b) ~>(%mean.'mate' !!) a))))
took ms/4.603
> ~>  %bout  (roll (gulf 1 1.000) |=([n=@ a=(unit @)] (hoon-mate (hoon-mate `n ~) (hoon-mate ~ `n))))
[~ 1.000]
```

The `+mate` error text (`bail: fail` followed by stack trace) looks different from the `+need` error text (`dojo: hoon expression failed`) despite using the same jet error routine (i.e. `u3m_bail(c3__fail)`). Feedback on this would be appreciated!

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -358,10 +358,10 @@ versioned_http_archive(
 versioned_http_archive(
     name = "zlib",
     build_file = "//bazel/third_party/zlib:zlib.BUILD",
-    sha256 = "b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30",
+    sha256 = "ff0ba4c292013dbc27530b3a81e1f9a813cd39de01ca5e0f8bf355702efa593e",
     strip_prefix = "zlib-{version}",
     url = "https://www.zlib.net/zlib-{version}.tar.gz",
-    version = "1.2.13",
+    version = "1.3",
 )
 
 #
--- a/pkg/noun/jets/b/mate.c
+++ b/pkg/noun/jets/b/mate.c
@@ -14,10 +14,10 @@
       return u3k(a);
     } else if ( u3_nul == a ) {
       return u3k(b);
-    } else if ( c3y == u3r_sing(a, b) ) {
+    } else if ( c3y == u3r_sing(u3t(a), u3t(b)) ) {
       return u3k(a);
     } else {
-      return u3m_bail(c3__fail);
+      return u3m_error("mate");
     }
   }
   u3_noun
--- a/pkg/noun/jets/b/mate.c
+++ b/pkg/noun/jets/b/mate.c
@@ -1,0 +1,30 @@
+/// @file
+
+#include "jets/q.h"
+#include "jets/w.h"
+
+#include "noun.h"
+
+
+  u3_noun
+  u3qb_mate(u3_noun a,
+            u3_noun b)
+  {
+    if ( u3_nul == b ) {
+      return u3k(a);
+    } else if ( u3_nul == a ) {
+      return u3k(b);
+    } else if ( c3y == u3r_sing(a, b) ) {
+      return u3k(a);
+    } else {
+      return u3m_bail(c3__fail);
+    }
+  }
+  u3_noun
+  u3wb_mate(u3_noun cor)
+  {
+    u3_noun a, b;
+    u3x_mean(cor, u3x_sam_2, &a, u3x_sam_3, &b, 0);
+    return u3qb_mate(a, b);
+  }
+
--- a/pkg/noun/jets/tree.c
+++ b/pkg/noun/jets/tree.c
@@ -2139,6 +2139,7 @@ static c3_c* _k140_ha[] = {
 
 /* new jets
 */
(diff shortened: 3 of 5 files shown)


pkg/noun/jets/b/mate.c
```diff
@@ -0,0 +1,30 @@
...
+
+  u3_noun
+  u3qb_mate(u3_noun a,
+            u3_noun b)
+  {
+    if ( u3_nul == b ) {
+      return u3k(a);
+    } else if ( u3_nul == a ) {
+      return u3k(b);
+    } else if ( c3y == u3r_sing(a, b) ) {
```
joemfb, line 17: To avoid a jet mismatch, you have to do this comparison explicitly against `u3t(a)` and `u3t(b)`.

pkg/noun/jets/b/mate.c
```diff
@@ -0,0 +1,30 @@
...
+      return u3k(a);
+    } else {
+      return u3m_bail(c3__fail);
```
joemfb, line 20: You can match the trace exactly here with `u3m_error("mate");`

WORKSPACE.bazel
```diff
@@ -358,10 +358,10 @@ versioned_http_archive(
 versioned_http_archive(
     name = "zlib",
     build_file = "//bazel/third_party/zlib:zlib.BUILD",
-    sha256 = "b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30",
+    sha256 = "ff0ba4c292013dbc27530b3a81e1f9a813cd39de01ca5e0f8bf355702efa593e",
     strip_prefix = "zlib-{version}",
     url = "https://www.zlib.net/zlib-{version}.tar.gz",
-    version = "1.2.13",
+    version = "1.3",
```
joemfb, line 364: I've seen this changed in at least a couple PRs. @matthew-levan can you follow up with this dependency update?

joemfb (changes requested):

pkg/noun/jets/b/mate.c
sidnym-ladrut: I now see something like this:

```
> (need ~)
dojo: hoon expression failed
> (mate `4 `5)
mate
dojo: hoon expression failed
```

Is there any way to omit the extra message? I also tried `u3m_error("")` and `u3m_error(u3_nul)`, but to no avail.

pkg/noun/jets/b/mate.c
sidnym-ladrut: Yes, you're quite correct. Thank you!

WORKSPACE.bazel
sidnym-ladrut: I think @matthew-levan fixed this already in `develop`, but I needed to cherry-pick the commit because I'm working from `vere-v2.11` (I couldn't find a working bootstrap pill for `2.12`).

matthew-levan:
Often, you can boot a new fakezod with an old pill. For instance, you could
probably boot one with `vere-v2.12` with a `2.11` pill and still be okay.
Just FYI for future reference.

pkg/noun/jets/tree.c
```diff
@@ -2139,6 +2139,7 @@ static c3_c* _k140_ha[] = {
 
 /* new jets
 */
+  static u3j_harm _139_two_mate_a[] = {{".2", u3wb_mate, c3y}, {}};
```
ashelkovnykov, line 2142: Style nit: This line should go on line ~`2274` (in between `_139_two__in_d` and `_139_two_d`)

ashelkovnykov (changes requested):
LGTM, pending another once over by the UF and the one style nit I flagged

pkg/noun/jets/tree.c
sidnym-ladrut: Done, thanks!

pkg/noun/jets/b/mate.c
joemfb: Sorry i missed this. The extra message is consistent with the hoon/nock for +mate. Arguably that's too verbose, but I don't think anyone has complained yet.

To produce a deterministic error without including "mate" in the stack trace, you'd just call `u3m_bail(c3__exit);`. Notably, the hoon for `+need` includes an error message, but the jet does not.

I try to keep jets consistent wrt to error messages, but system doesn't enforce that to the degree it should, and it's much less of a problem than any other kind of mismatch.

ashelkovnykov (approved):

joemfb (approved):
this will do

PR #514, opened 2023-09-11
`noun`: actually use `u3`'s version number

matthew-levan:
A draft working towards proper versioning of the `noun` library, which is essentially the library which implements the Noun and how it is represented in-memory and on-disk. 

This work starts with renaming `pkg/noun` to `pkg/u3`.

This work is in the interest of establishing a proper "loom migration framework" and design pattern which makes future upgrades to the loom format relatively simple when compared with how difficult it is to upgrade the loom format today. 

With this strategy, Vere's file hierarchy would look something like this:
```
pkg/
├── c3
├── ent
├── noun        # aliases to current functions, structs, etc.
├── u3          # noun library as of before pointer compression
├── u4          # noun library as of pointer compression
├── u5          # noun library for persistent memoization
├── ...         # future noun library versions
├── migrations  # migrations between sequential pairs of noun library versions
├── ur
├── urcrypt
└── vere
```


matthew-levan:
Impractical.

PR #529, opened 2023-09-29
Fixed next/kelvin workflow to not upload to docker

mopfel-winrux:
The next/kelvin workflow uploads to docker edge. This shouldn't happen as it messes up edge testing if the next/kelvin branch isn't in sync with develop


joemfb (approved):
this will do

PR #535, opened 2023-10-07
BLAKE3 jets

lukechampine:
See https://github.com/urbit/urbit/pull/6802 and https://github.com/urbit/urcrypt/pull/8

Depends on https://github.com/urbit/vere/pull/524


lukechampine:
Anyone on a beefy amd64 machine want to benchmark this? I'm on an M2, so I can't see those juicy AVX-512 speedups. 😭 Still, ~1.2 GB/s ain't bad -- that's 2x faster than sha3, and 4x faster than blake2b.

belisarius222:
It looks like there's some git churn here.  Maybe merge develop into your branch?

pkg/noun/jets/e/blake.c
```diff
@@ -54,6 +53,110 @@
...
+    if ( !u3r_word_fit(&wid_w, wid) ) {
+      return u3m_bail(c3__fail);
+    } else {
+      c3_y  cv_y[32], block_y[64], block_len;
+      c3_y *dat_y = u3r_bytes_alloc(0, wid_w, dat);
+      c3_d counter_d = u3r_chub(0, counter);
+      c3_y flags_y = u3r_byte(0, flags);
+      u3r_bytes(0, 32, cv_y, cv);
+      urcrypt_blake3_chunk_output(wid_w, dat_y, cv_y, block_y, &block_len, &counter_d, &flags_y);
+      return u3i_cell(u3i_bytes(32, cv_y), u3i_qual(u3i_chub(counter_d), u3i_bytes(64, block_y), block_len, u3i_bytes(1, &flags_y)));
```
lukechampine, line 113: worth defining `u3i_quil`? 🤔

joemfb (changes requested):
These look good. I've noted a couple issues with one allocation, and an assertion that should be removed.

These jets will go into 411 (or later), so they should not be declared in the hoon %140 cores. I've setup the declarations for 411 (hoon 138) in #567, it'll be simplest for this PR to target that one.

joemfb (approved):
this will do

PR #536, opened 2023-10-09
u3: inline cell deconstruction

joemfb:
This PR converts `u3r_cell()`/trel/&c inline functions and `u3x_cell()`/trel/&c to macros. This is mostly just a straightforward optimization of frequently used deconstructors, as prelude to optimizing all the jet-interface (`u3w*`) functions.

In the case of `u3x_atom()` the previously-defined macro repeated its argument, which is a bad pattern as that "argument" is often a call to another function. It has now been switched to an inline function.

### DIFF
--- a/pkg/noun/retrieve.c
+++ b/pkg/noun/retrieve.c
@@ -9,6 +9,35 @@
 #include "trace.h"
 #include "xtract.h"
 
+
+//  declarations of inline functions
+//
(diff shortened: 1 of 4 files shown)


matthew-levan (approved):

matthew-levan (approved):
LGTM.

PR #538, opened 2023-10-10
u3: restore old-style, efficient reap to u3m_love()

joemfb:
This PR effectively reverts urbit/urbit#1753, which simplified cache promotion when falling back to an outer road, at the expense of allocation efficiency.

Early testing shows significant positive effects mitigating snapshot patch size (see urbit/urbit#6805)


joemfb:
For posterity, [f747d86](https://github.com/urbit/vere/pull/538/commits/f747d862f1d22d1eb39d90dee08f10af0ccec5bd) is extending the fix from urbit/urbit#1059 to hank cache promotion.

ashelkovnykov (approved):
Went through this @joemfb . Only part I didn't 100% understand was the `hank` stuff, but I understood it well enough based on cargo-culting of form to approve this PR.

ashelkovnykov (approved):

PR #543, opened 2023-10-19
ames: resolve galaxy ips using ipv4 only

pkova:
Some ships experience a lot of printfs like `ames: czar at zod.urbit.org: not found (b)` for galaxies that are up and have no problems with their DNS config. I initially thought this was something else, but the problem I found by mucking around with `tcpdump` is that we issue both a IPv4 A record query and a IPv6 AAAA record query. When the IPv6 query resolves it thrashes all kinds of state, sometimes causing the spurious printf above.

This PR configures `uv_getaddrinfo` to return just IPv4 addresses. I also removed the unnecessary while loop.


belisarius222 (approved):

joemfb (approved):
great catch!

PR #544, opened 2023-10-23
vere: bring your own boot sequence

Fang-:
Adds support for `--prop-file`, `--prop-url` and `--prop-name`, which let one augment the boot sequence of a ship with additional events.

Back-ports the vere changes from urbit/urbit#5470 onto current vere. See that for additional details/history. This is a mostly straightforward, copy-paste port of the changes from that original PR, with the following exceptions:
- Boot sequence construction now happens in `_pier_boot_make`.
- We don't modify the way the default boot sequence is acquired (yet).

We will want to follow up on the latter, eventually. I've left `TODO`s in the relevant places. Figured it might be better to be less disruptive to release processes initially. cc @pkova

The %autoprop "prop builder" that was included in urbit/urbit#5470 should still operate fine and be useful for automatically keeping certain kinds of props up to date. (Save for minimal kelvin compatibility disrepair it has fallen into due to the flow of time.)


pkg/vere/king.c
```diff
@@ -158,6 +160,49 @@ _king_boot(u3_noun bul)
...
+        u3_atom jam = _king_get_atom(vex_u->loc_c);
+        mor = u3nc(u3ke_cue(jam), mor);
+      } break;
+
+      case 3: {  //  name
+        //NOTE  this implementation limits us to max 38 char prop names
+        c3_c url_c[80];
+        sprintf(url_c,
+                //TODO  should maybe respect ops_u.url_c
+                "https://bootstrap.urbit.org/props/" URBIT_VERSION "/%s.jam",
```
Fang-, line 188: I think this might be correct in the long run, but not right now due to interactions between `-B` and `-u`, and the fact that we hardcode `bootstrap.urbit.org` in a handful of other places as well.

pkg/vere/pier.c
```diff
@@ -1872,7 +1873,48 @@ _pier_boot_make(u3_noun who,
     bot_u.use = u3nc(u3nc(wir, cad), bot_u.use);
   }
 
+  //  prepend & append additional boot enhancements to the userspace sequence
+  //
+  {
```
Fang-, line 1878: Was thinking about inlining the `$prop` type definition here for clarity, but feels kind of weird/ugly to just chuck in there. Not sure what the preferred style for including a "spec" like this is.

```hoon
+$  prop  [%prop meta tier (list ovum)]
+$  meta  term
+$  tier  ?(%fore %hind)  ::  before or after userspace
```

pkg/vere/king.c
joemfb: This should use snprintf to check against stack overflow. And you should bump the max size to 256.

pkg/vere/pier.c
joemfb: just putting that in a comment here is good

pkg/vere/main.c
```diff
@@ -215,6 +215,19 @@ _main_pier_run(c3_c* bin_c)
...
 }
 
+/* _main_add_prop(): add a boot prop to u3_Host.ops_u.vex_u.
+*/
+u3_even*
+_main_add_prop(c3_i kin_i, c3_c* loc_c)
+{
+  u3_even* nex_u = c3_calloc(sizeof(*nex_u));
+  nex_u->kin_i = kin_i;
+  nex_u->loc_c = loc_c;  //  XX _main_repath where appropriate?
```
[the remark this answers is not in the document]
joemfb: The "path canonicalization" stuff here is kind of a mess. And there's no need to call `_main_repath()` here. If you want to figure out what's going on anyway, you could try printing the path first ...

joemfb (changes requested):
time is a flat circle.

thanks for reviving/restaging this, a couple nits:

joemfb (approved):
this will do

PR #547, opened 2023-11-07
vere: refactors event log metadata reading to avoid the loom

joemfb:
This PR resolves a bug that I introduced in #531. I added stronger validation during event log initialization by checking for the presence of metadata, but that was allocating on the loom (in the case of 32-bit or greater ships and/or lives). There are various chicken-and-egg problems with initializing old event logs for replay, so it seems best to just remove this whole operation from the loom.

### DIFF
--- a/pkg/vere/db/lmdb.c
+++ b/pkg/vere/db/lmdb.c
@@ -414,7 +414,7 @@ void
 u3_lmdb_read_meta(MDB_env*    env_u,
                   void*       ptr_v,
                   const c3_c* key_c,
-                  void (*read_f)(void*, size_t, void*))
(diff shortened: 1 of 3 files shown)


matthew-levan (dismissed):
Tested with a fake `~dostec-risfen` (`0xffff.ffff`); this fixes the bug discussed earlier this morning. I booted it, rolled it, then restarted it successfully. Only question: How come we don't have to check if `fak_o` is non-zero on line 761?

joemfb:
@matthew-levan if the length is non-zero, the fake-bit should always be 1. but it's probably better to read that out explicitly regardless. thanks

matthew-levan (approved):

PR #548, opened 2023-11-08
disk: check for null pointer when writing through fak_o

pkova:
#547 was not quite enough to get `~binnec-dozzod-marzod` back online. This one is. Thanks to @mopfel-winrux for the solution.


joemfb (approved):
brutal, thanks for catching this.

PR #552, opened 2023-11-14
u3: redeclare inline u3x_atom for linkability

joemfb:
This PR fixes my mistake in #536. Thanks for @liam-fitzgerald for flagging that this broke debug builds.


matthew-levan (approved):
`bazel build :urbit --compilation_mode=dbg` works now, thanks for fixing this. @mopfel-winrux @pkova and @yosoyubik were running into this yesterday.

PR #557, opened 2023-11-17
Replace unsafe C functions like strcpy with safer alternatives

### LINKED ISSUE #556: Replace unsafe C functions like strcpy with safer alternatives
Inspired by https://github.com/urbit/archaeology/pull/71, there are numerous calls to `strcpy()`, `sprintf()`, and other "unsafe" functions. Many of these are harmless...but a few actually seem feasibly exploitable! Kinda!

In any case, in my fork (https://github.com/darighost/vere) I replaced all of the calls to `strcpy()`, `sprintf()` with calls to `snprintf ` instead. It's also possible I'm breaking a bunch of stuff without realizing. My fork compiles and I ran it and did Urbit stuff but who knows, there's an infinite amount of behavior I didn't test so yeah.

If this is desired, I'd also like to replace some other functions as well. But I wanted to check, is this even a change you guys want? Also, I'm using this as a chance to "practice" contributing to this repo, so please let me know if I should be doing something differently.

Thanks!

PS: I know, I know, we're all awaiting Ares with bated breath. What can I say. I also like PHP and Visual Basic 5, sue me.

narskidan:
Resolves https://github.com/urbit/vere/issues/556

So far I'm just replacing strcpy and sprintf (using snprintf to replace both), but there are others I'd like to replace as well.

Just starting with this because I'd like to be corrected on mistakes I'm making now, before I continue.

I'm able to compile and run with these changes, but maybe there's something I should be testing that I'm not. Please do let me know.


ashelkovnykov:
Everything seems in order so far. @joemfb @barter-simsum ?

PR #559, opened 2023-11-25
c3: new mote macro

mrdomino:
I was browsing through the Ares code and realized you could do the `tas!` macro as a normal C constant expression.

`c3_s5` through `c3_s8` were unused, so I removed them.


mrdomino:
Going to mark this ready for review, but I have not yet gotten bazel working, so this remains untested on my machine; however, the `c3_tas` macro has been tested via Compiler Explorer: https://godbolt.org/z/dscrda5c4

mrdomino:
Ah, my mistake — I forgot to use C rather than C++. Apparently `c3_tas` is not a compile-time C constant. :(

PR #563, opened 2023-12-04
lick: Shut bug

mopfel-winrux:
There is a bug in lick where if a port still has a connection and lick gets a %shut task the port can still send %soaks to arvo. Arvo poops when this happens. This bug also shows up when the port is disconnected and a %disconnect %soak is sent. This PR fixes both of these bugs


joemfb (approved):
this will do

PR #565, opened 2023-12-05
nix: update flake.lock to fix FHSUserEnv

eamsden:
This updates nixpkgs in `flake.lock` to bring in a fix for the `FHSUserEnv` required to run Bazel on NixOS. Without this fix, on recent NixOS, an improper LD_LIBRARY_PATH will be set, causing glibc to crash processes with "stack smashing detected."

See NixOS/nixpkgs#263201

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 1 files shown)


joemfb (approved):
assuming this is fine

PR #569, opened 2023-12-13
ames, http: declare gang correctly when scrying

pkova:
Also refactors fetching the protocol version twice.


joemfb (approved):
this will do

PR #570, opened 2023-12-18
ames: populate gang with fine sender when scrying into arvo

pkova:
This is what we actually want to do.

@joemfb am I leaking something here?


pkova:
Right that all makes sense. Just the responses are signed..

PR #574, opened 2024-01-02
ames: independent galaxy dns resolution for stun

yosoyubik:
DNS galaxy resolution needs to be treated differently when dealing with STUN packets to our sponsor and when this is meant to be for another galaxy, as part of sending an ames packet.


joemfb (approved):
this will do

PR #577, opened 2024-01-03
jets: make a real jet for swp

pkova:
Previously running `(swp 3)` on a 70MB atom would always exhaust a 2gb loom. With this we can easily `(swp 3)` and even `(swp 0)` a 500MB atom. Note that fares a lot better even with no special casing of blocksizes 0, 3 and 5.


joemfb (approved):
This looks great. I think we'll want to special-case bloq `3` (and maybe `0`). And `+rev` should get a similar rewrite.

PR #579, opened 2024-01-04
build: recent macOS updates clang

mrdomino:
Not sure what you want to do with this, but on Sonoma, clang 15.0.0 is what is installed.

I'm not currently able to build on my OSX machine, seemingly because I can't get bazel and my nix libtool package to cooperate.


matthew-levan:
Thanks, I've been meaning to do this. It seems reasonable to keep the version pegged to the version that ships with the latest stable macOS.

matthew-levan (approved):

PR #580, opened 2024-01-04
conn: fchdir, non-fatal error

mrdomino:
- Uses `open` / `fchdir` instead of `getcwd` / `chdir` to return to the
  old cwd after setting up the socket.

- If we can't `fchdir` back (either because we couldn't open `.`
  `O_EXEC` or because `fchdir` itself failed), we don't kill the
  process; we just keep going with cwd at the pier directory.

This fixes an issue where if urbit is started with `sudo -u pier` from a
directory that the pier user doesn't have access to (e.g. a different
user's home directory that is chmod 0750), the process would die. It is
also more robust in the face of unusual or changing paths.


mrdomino:
I actually think you can do this without `chdir` - just make the socket in `/tmp` and symlink it from the pier.

PR #581, opened 2024-01-05
conn: do not chdir for conn.sock

mrdomino:
Instead, we create the socket in a temporary directory - `TMPDIR` if it is not too long, or else `/tmp`. The directory is made via `mkdtemp` with the template `urbit-XXXXXX`. The socket is located there (getting around the length limit) and `$pier/.urb/conn.sock` is symlinked to it.

This fixes an issue where if urbit is started with `sudo -u pieruser` from a directory that the pier user doesn't have access to (e.g. a different user's home directory that is chmod 0750), the process would die trying to `chdir` back to the old directory.

It also gives clients more flexibility; previously, they would have had to `chdir` to the pier directory to connect to the socket without risking overrunning the length limit. They can still do that (and use `".urb/conn.sock"` as their `sun_path`), but they now also have the option of using `readlink` and connecting to the full absolute path of the socket without changing directories.

Replaces calls to `uv_strerror` with `strerror` for system calls; the former does not print system call errors properly.

Supersedes #580.

### DIFF
--- a/pkg/vere/io/conn.c
+++ b/pkg/vere/io/conn.c
@@ -693,55 +693,63 @@ _conn_init_sock(u3_shan* san_u)
   u3l_log("conn: listening on %s", pip_c);
 
 #else   //  _WIN32
-  //  the full socket path is limited to about 108 characters,
-  //  and we want it to be relative to the pier. save our current
-  //  path, chdir to the pier, open the socket at the desired
-  //  path, then chdir back. hopefully there aren't any threads.
-  //
-  c3_c pax_c[2048];
-  c3_i err_i;
+  c3_c tad_c[104];
+  c3_i err_i, fid_i;
+  c3_c *ven_c;
 
-  if ( NULL == getcwd(pax_c, sizeof(pax_c)) ) {
-    u3l_log("conn: getcwd: %s", uv_strerror(errno));
+  if ( (ven_c = getenv("TMPDIR")) ) {
+    if ( strlen(ven_c) >= 80 ) {
+      u3l_log("conn: TMPDIR too long");
+      ven_c = 0;
+    }
+  }
+  if ( !ven_c ) {
+    ven_c = "/tmp";
+  }
+  snprintf(tad_c, sizeof(tad_c), "%s/urbit-XXXXXX", ven_c);
+  if ( !mkdtemp(tad_c) ) {
+    u3l_log("conn: mkdtemp: %s", strerror(errno));
     u3_king_bail();
   }
-  if ( 0 != chdir(u3_Host.dir_c) ) {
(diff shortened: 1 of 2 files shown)


mrdomino:
I'm actually indifferent as to whether this or #580 is taken. I think this is a slightly better solution, since it allows clients to always be able to specify the full absolute path to the socket when connecting to it. But either way is fine.

mrdomino:
@pkova Are you the right person to review this?

mrdomino:
Oh, I guess now there's also `lick.c`.

In that case the temporary directory should probably be stored on `u3_Host` or `u3C` or something and created in `main.c`. Right?

mrdomino:
@mopfel-winrux how do you want to handle `%lick`? I don't have much insight yet into what it's doing. Would it make sense for it to follow suit with this change and use a tmp directory for its sockets?

If not, then the next best thing is probably to reopen #580 and also apply that to `lick.c`.

pkg/vere/io/conn.c
```diff
@@ -693,55 +693,79 @@ _conn_init_sock(u3_shan* san_u)
...
-  if ( 0 != unlink(URB_SOCK_PATH) && errno != ENOENT ) {
-    u3l_log("conn: unlink: %s", uv_strerror(errno));
-    goto _conn_sock_err_chdir;
+  memcpy(tad_c, ven_c, len_w);
+  memcpy(tad_c + len_w, "/urbit-XXXXXX", sizeof("/urbit-XXXXXX"));
+  if ( !mkdtemp(tad_c) ) {
+    u3l_log("conn: mkdtemp: %s", strerror(errno));
+    u3_king_bail();
   }
+  memcpy(tad_c + len_w + sizeof("/urbit-XXXXXX") - 1, "/conn.sock",
```
ashelkovnykov, line 724: The string "/urbit-XXXXXX" gets reused a lot, we should make it a constant or macro

ashelkovnykov (approved):
LGTM, but flagged one thing

mopfel-winrux:
@mrdomino I'm sorry I completely missed this. We should also do this for `lick.c`

mrdomino:
Yeah. How does lick work? Does it make one socket or many?

I guess it's going to be necessary to track and manage the temp dir from some location that is neither conn.c nor lick.c if we go this route.

I also gave a little thought to putting the socket directory under /run/urbit/sampel-palnet rather than /tmp/urbit-XXXXXX. I'm leaning towards /tmp at the moment, but could be persuaded either way.

mopfel-winrux:
It makes multiple sockets. One per `[%spin name=path]` card it is given. It puts them in `.urb/dev/ folder under the name path. we could use that directory

PR #582, opened 2024-01-05
vere: disable systemd and apparmor for the dbus dependency

pkova:
Some people have been having problems building vere because of these.


joemfb (approved):
this will do

PR #584, opened 2024-01-16
mdns: distinguish fake ships in the mdns namespace

pkova:
Also print a better error message for linux users if avahi is not installed.

I'm targeting this wonky branch to get this into 3.0.


joemfb (approved):
this will do

PR #593, opened 2024-01-29
ames: add libnatpmp for automatic port forwarding

pkova:
Same as https://github.com/urbit/urbit/pull/3261 but for vere. From what I can tell this NAT-PMP stuff is fairly well supported by routers, works on my machine at least.

### DIFF
--- a/pkg/vere/io/ames.c
+++ b/pkg/vere/io/ames.c
@@ -2762,21 +2762,21 @@ static void natpmp_init(uv_timer_t* handle);
 
 static void
 natpmp_cb(uv_poll_t* handle,
-          int        status,
-          int        events)
+          c3_i        status,
+          c3_i        events)
 {
   u3_ames* sam_u = handle->data;
 
   natpmpresp_t response;
-  c3_w r = readnatpmpresponseorretry(&sam_u->nat_u.req_u, &response);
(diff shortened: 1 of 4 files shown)


joemfb (commented):
this will do

joemfb (approved):

joemfb (changes requested):
I've noted a couple things that could be cleaner with the libuv handle lifetime.

Also, the function signatures should be adjusted in the normal style.

pkg/vere/io/ames.c
```diff
@@ -2752,6 +2758,51 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+
+static void
+natpmp_cb(uv_poll_t* handle,
+          int        status,
+          int        events)
+{
+  u3_ames* sam_u = handle->data;
+
+  natpmpresp_t response;
+  c3_w r = readnatpmpresponseorretry(&sam_u->nat_u.req_u, &response);
```
joemfb, line 2771: This assignment converts a signed `int` to an unsigned `c3_w`. You should `c3_i` (and suffix the variable name with `_i`).

joemfb (approved):
this will do

PR #596, opened 2024-01-31
epoc: fix log metadata write bug, make pre-release repair more robust

joemfb:
- event log metadata was serialized through loom-dependant apis. this could cause crashes in certain scenarios of partial-initialization.
- the pre-release "repair" on disk initialization handled missing but not empty  top-level lmdb environments. these can be created by crashes during the repair, so it's best to handle them as well.


matthew-levan (approved):
LGTM. Nice catch on handling inter-repair failures. I successfully tested this with a fake `~dister-dozzod-marzod`.

PR #608, opened 2024-02-14
v3-migrate: ream before reclaiming bytecode caches

pkova:
Some ships migrating to vere-v3.0 experienced "loom: corrupt" when reclaiming the bytecode caches when running the v3 migration. This PR does fix the issue, but it may or may not be exactly right. @joemfb will have to take a look.


joemfb (approved):
tyfys

PR #612, opened 2024-03-01
Revert "u3: restore old-style, efficient reap to u3m_love()"

pkova:
Reverts urbit/vere#538

My ship ran into a refcounting bug during pre-release testing of vere-v3.0. The only part of the release that touches the relevant codepath is this PR, so we have decided to revert it.


joemfb (approved):
with sorrow

PR #617, opened 2024-03-06
xmas: new runtime driver for directed messaging

liam-fitzgerald:


joemfb:
I've updated this driver to target the WIP vane from urbit/urbit#6932 (`%mesa`). Note that most of these changes were first staged on the `yu/mesa` branch of this repo.

PR #629, opened 2024-04-10
u3: adds deletion to HAMT

joemfb:
Ported from #617 for independent review, with some test cleanup.

### DIFF
--- a/pkg/noun/hashtable.c
+++ b/pkg/noun/hashtable.c
@@ -287,7 +287,7 @@ _ch_buck_del(u3h_slot* sot_w, u3_noun key, c3_w *use_w)
     u3_noun kov = u3h_slot_to_noun(hab_u->sot_w[i_w]);
     if ( c3y == u3r_sing(key, u3h(kov)) ) {
       fin_w = i_w;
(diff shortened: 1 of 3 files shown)


pkg/noun/hashtable.c
```diff
@@ -270,6 +270,134 @@ u3h_put(u3p(u3h_root) har_p, u3_noun key, u3_noun val)
...
+    return _ch_buck_del(sot_w, key, use_w);
+  }
+  //u3h_node* han_u = (u3h_node*)han_v;
+  //if ( _ch_popcount(han_u))
+  return _ch_node_del(sot_w, key, lef_w, rem_w, use_w);
+}
+
+
+
+/** TODO: this has a bug where the freeing a hamt that this has been called with failes
```
belisarius222, line 382: @joemfb This seems like something that should be addressed before merging.

joemfb: I think it's obsolete, but need to confirm before removing it. These changes need a full review, didn't want them to sneak in.

joemfb:
obsolete as of #738

PR #635, opened 2024-04-16
add more helpful chop message

### LINKED ISSUE #634: ux: improve unhelpful error messages
This is a thoroughly unhelpful message, especially in light of previous behavior. `chop` used to truncate the event log, now `roll` needs to be run before it'll truncate. If someone is trying to `chop` and there's nothing to truncate, let's tell them to run `roll` first!

As a followup, there should be links to documentation that explain what these things do and how to use them more thoroughly than what can be provided as CLI feedback. Or `man` pages?

https://github.com/urbit/vere/blob/25a7183c3c69526789b5c5886d5849b09c3c24df/pkg/vere/disk.c#L1635

jalehman:
Resolves #634

### DIFF
--- a/pkg/vere/disk.c
+++ b/pkg/vere/disk.c
@@ -1632,7 +1632,7 @@ u3_disk_chop(u3_disk* log_u, c3_d eve_d)
   u3_disk_epoc_list(log_u, sot_d);
 
   if ( len_z <= 2 ) {
-    fprintf(stderr, "chop: nothing to do, have a great day\r\n");
+    fprintf(stderr, "chop: nothing to do, try running roll first\r\n");
     exit(0);  //  enjoy
   }
 
--- a/pkg/vere/disk.c
+++ b/pkg/vere/disk.c
@@ -1632,7 +1632,9 @@ u3_disk_chop(u3_disk* log_u, c3_d eve_d)
   u3_disk_epoc_list(log_u, sot_d);
 
   if ( len_z <= 2 ) {
-    fprintf(stderr, "chop: nothing to do, try running roll first\r\n");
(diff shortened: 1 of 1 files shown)


jalehman:
We could be even more explicit. Instead of:

`chop: nothing to do, try running roll first`

We could say something to the effect of:

`chop: nothing to do, try running ./path/to/pier/.run roll`

jalehman:
It also looks like there's no documentation for `roll` in the [runtime reference](https://docs.urbit.org/manual/running/vere#chop), and the entry for `chop` is out of date.

tinnus-napbus:
I updated the runtime reference earlier, PR just needs review/merge

jalehman:
Great, much appreciated. The time that someone needs the runtime reference is when they're running into an issue while using vere. What do you think about adding links to documentation at the point of encounter? For example:

```
printf(stderr, "chop: nothing to do, try running roll first\r\n");
printf(stderr, "info: https://docs.urbit.org/manual/running/vere#chop\r\n");
```

jalehman:
Let's not let the perfect be the enemy of the good. Given that `roll` isn't dangerous at all, recommending it seems fine. I've had this issue myself (not knowing to roll before chop) and seen several cases of others experiencing it. 

Further improvements would be great: feedback about what happened after running `roll` that could help the user diagnose whether or not the operation was still useful in the cases of 3 or 4; a `force` flag to chop to combine common operations; etc.

tinnus-napbus:
In the case of 4, if you're unsatisfied with the size reduction, you may want to start/stop the ship, roll, start/stop the ship, roll, then chop, to maximally reduce size.

tinnus-napbus (approved):

tinnus-napbus (approved):
updated with docs link

PR #638, opened 2024-04-18
Lagoon jets for IEEE 754 floats.

sigilante:
Pursuant to draft UIP at https://github.com/urbit/UIPs/blob/sigilante/lagoon-reals/UIPS.

Should be released alongside corresponding Hoon code.


sigilante:
The current root problem is that `aarch` wants a ton of SoftFloat functions that aren't linked yet.

https://github.com/urbit/vere/actions/runs/8819670490/job/24211443275?pr=638

Adding them piecemeal seems to be a fool's errand:  any way to just build the whole thing at one go for each?  I.e. include all files rather than manually listing the subset of interest?

sigilante:
There seem to be some platform-specific optimized versions linked in as well, so this may need to stay in the same vein to be effectively fast.

sigilante:
@urbit/kernel this is building now thanks to @pkova's sleuthing.

sigilante:
This is ready for review.

sigilante:
When the tests pass here on that last commit, then this is ready to merge. @joemfb @pkova @belisarius222

joemfb:
urbit/urbit#6971

PR #642, opened 2024-05-01
disk: fix broken chop error message

pkova:
#635 broke the develop branch and therefore broke CI on `urbit/urbit`. It's a real C moment that CI failed to catch this one.


joemfb (commented):
The previous version is valid in c. String literals separated by whitespace are automatically concatenated.

pkova:
Yup, not this. Reverting.

PR #646, opened 2024-05-10
ames: libnatpmp for real this time

pkova:
This is #593 that got reverted in #644 because it broke `urbit/urbit` CI. This PR fixes the issue which was unconditionally calling `uv_poll_init` on a socket that was initialized with `initnatpmp` without checking if `initnatpmp` returned an error first.

### DIFF
--- a/pkg/vere/io/ames.c
+++ b/pkg/vere/io/ames.c
@@ -3200,6 +3200,11 @@ _ames_io_exit(u3_auto* car_u)
   uv_close(&sam_u->had_u, _ames_exit_cb);
   uv_close((uv_handle_t*)&sam_u->sun_u.dns_u, 0);
   uv_close((uv_handle_t*)&sam_u->sun_u.tim_u, 0);
+  uv_close((uv_handle_t*)&sam_u->nat_u.tim_u, 0);
+
+  if (uv_is_active((uv_handle_t*)&sam_u->nat_u.pol_u)) {
+    uv_close((uv_handle_t*)&sam_u->nat_u.pol_u, 0);
+  }
 }
 
 /* _ames_io_info(): produce status info.
--- a/pkg/vere/io/ames.c
+++ b/pkg/vere/io/ames.c
@@ -3202,7 +3202,8 @@ _ames_io_exit(u3_auto* car_u)
   uv_close((uv_handle_t*)&sam_u->sun_u.tim_u, 0);
   uv_close((uv_handle_t*)&sam_u->nat_u.tim_u, 0);
 
-  if (uv_is_active((uv_handle_t*)&sam_u->nat_u.pol_u)) {
+  uv_handle_type handle = uv_handle_get_type((uv_handle_t *)&sam_u->nat_u.pol_u);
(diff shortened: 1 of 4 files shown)


pkg/vere/io/ames.c
```diff
@@ -2765,6 +2771,60 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+  u3_ames* sam_u = handle->data;
+  c3_s por_s = sam_u->pir_u->por_s;
+
+  c3_i err_i = initnatpmp(&sam_u->nat_u.req_u, 0, 0);
+
+  if (err_i != 0) {
+    return;
+  }
+
+  uv_poll_init(u3L, &sam_u->nat_u.pol_u, sam_u->nat_u.req_u.s);
```
joemfb, line 2820: It looks like `uv_poll_init()` is categorically different from other libuv init functions, and we need to check for errors here.

pkg/vere/io/ames.c
```diff
@@ -2765,6 +2771,60 @@ _ames_recv_cb(uv_udp_t*        wax_u,
   }
 }
 
+static void natpmp_init(uv_timer_t* handle);
+
+static void
+natpmp_cb(uv_poll_t* handle,
+          c3_i        status,
+          c3_i        events)
+{
```
joemfb, line 2780: We're not checking `status` for errors here. And I'm not sure if `readnatpmpresponseorretry()` can/will handle all cases for us.

joemfb (changes requested):
This PR never disposes of this new `uv_poll_t` and `uv_timer_t` handles. It should `uv_close()` them in `_ames_io_exit()`. The fact that we can't unconditionally init the `uv_poll_t` means we need some state to tell us if we should close it. It might be best to just heap allocate it so we can check if the pointer is null.

pkg/vere/io/ames.c
```diff
@@ -3126,6 +3200,11 @@ _ames_io_exit(u3_auto* car_u)
   uv_close(&sam_u->had_u, _ames_exit_cb);
   uv_close((uv_handle_t*)&sam_u->sun_u.dns_u, 0);
   uv_close((uv_handle_t*)&sam_u->sun_u.tim_u, 0);
+  uv_close((uv_handle_t*)&sam_u->nat_u.tim_u, 0);
+
+  if (uv_is_active((uv_handle_t*)&sam_u->nat_u.pol_u)) {
+    uv_close((uv_handle_t*)&sam_u->nat_u.pol_u, 0);
+  }
```
joemfb, line 3207: I'm not sure if this is quite right -- the definition of "active" for different handle types is fairly involved. What we actually want is just "has this handle every been initialized?". We can obviously track that ourselves, but I think this will also work:

```
if ( UV_UNKNOWN_HANDLE != uv_handle_get_type(&sam_u->nat_u.pol_u) ) {
  uv_close(...)
}
```

`UV_UNKNOWN_HANDLE` is `0`, and we zero-initialize the `uv_poll_t` struct by virtue of allocating the `u3_ames` struct with `c3_calloc()`.

joemfb (changes requested):

PR #649, opened 2024-05-20
play: exit if binary version mismatches latest epoch's

matthew-levan:
The invariant that replay should not occur across binary versions was previously unenforced. This enforces it by setting a flag in the disk struct to indicate version mismatch between ourselves and the latest epoch's binary version number (saved in `vere.txt`).

Then, in the synchronous replay function, we exit if the same flag is set.

Questions:
- Should we instruct the user to downgrade to a previous version before exit?
- Should we make `_disk_epoc_meta` public (i.e. `u3_disk_epoc_meta`) so we can provide the specific version to revert to in the error message?

Note: Whitespace is cleaned up here automatically thanks to my editor and I don't see why not to include them in this PR. Happy to revert with request though as I don't really care either way.

### DIFF
--- a/pkg/vere/disk.c
+++ b/pkg/vere/disk.c
@@ -1235,7 +1235,7 @@ _disk_epoc_roll(u3_disk* log_u, c3_d epo_d)
     fprintf(stderr, "disk: failed to read metadata\r\n");
     goto fail3;
   }
-  
+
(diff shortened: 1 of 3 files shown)


joemfb:
Superseded by #860

PR #653, opened 2024-05-22
Add crc32 jet

nathanlever:
Adds crc32 jet. This will be used to validate uncompressed gzip data.
https://github.com/urbit/urbit/pull/6997

### DIFF
--- a/pkg/noun/jets/e/crc32.c
+++ b/pkg/noun/jets/e/crc32.c
@@ -1,0 +1,51 @@
+/// @file
+
+#include <stdio.h>
+#include <allocate.h>
+#include "zlib.h"
+
+#include "jets/w.h"
+
+#include "noun.h"
+
+u3_noun
+u3qe_crc32(u3_noun input_octs)
+{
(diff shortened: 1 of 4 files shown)


pkg/noun/jets/e/crc32.c
```diff
@@ -0,0 +1,56 @@
...
+
+#include "noun.h"
+
+u3_noun
+u3qe_crc32(u3_noun input_octs)
+{
+  u3_atom head = u3h(input_octs);
+  u3_atom tail = u3t(input_octs);
+  c3_w  tel_w = u3r_met(3, tail);
+  c3_w  hed_w = u3r_met(3, head);
```
joemfb, line 17: `hed_w` should be the value of `head` as a `c3_w`, not its byte-width. 

```
c3_w hed_w;
if ( c3n == u3r_safe_word(head, &hed_w) ) {
  return u3m_bail(c3__fail);
}

```

joemfb (approved):
this will do

PR #654, opened 2024-05-24
boot: replay in subprocess

matthew-levan:
This PR:
1. Adds the `--serf-bin` option for specifying a serf executable
2. Sets `u3_Host.wrk_c` to `--serf-bin`'s value if specified, self otherwise
3. Performs boot's replay in a subprocess that spawns our own serf (by default) or the specified serf if provided
4. Spawns a thread in `_cw_play` that exits if it reads `EOF` from `STDIN` (thereby ensuring the child process for replaying events exits whenever the parent is killed)

These changes allow Ares (or any other runtime) to implement its own `play` subcommand and have Vere use it for the boot process's replay functionality.

Tests:
- [x] `linux-x86_64` subcommand, replay via boot, killing king via INT, KILL, TERM
- [x] `macos-aarch64` subcommand, replay via boot, killing king via INT, KILL, TERM

PR #655, opened 2024-05-24
disk: parameterize lmdb map size

matthew-levan:
Adds `--lmdb-map-size` to specify LMDB's initial memory mapping size, intended for use by users who need a size smaller than our default of 500G. Note: `case 12` to avoid collision with the other arg added in #654.


matthew-levan:
Thanks @joemfb for quick feedback; addressed.

pkova (approved):

PR #639, opened 2024-04-19
ames: resolve galaxy dns in batches; plug leak

joemfb:
This PR refactors ames' galaxy dns resolution and simplifies its STUN protocol state machine. By removing the packet queueing in the old galaxy dns handling, this PR also plugs a memory leak introduced in #305 (f9e8542f9596cfb826ee6bd434c7fcea31bb251d specifically).

PR #658, opened 2024-06-11
ames: free packet on successful galaxy dns resolution

pkova:


joemfb (approved):

joemfb:
We believe this resolves urbit/urbit#7010

joemfb:
#639 replaces this edge-triggered galaxy-dns resolution with a global batch, which should further reduce memory pressure.

PR #662, opened 2024-06-26
ames: factor out stun implementation and basic de/serialization functions

joemfb:
This PR factors out a couple small modules from ames, in anticipation of merging in the directed messaging driver. Built on top of (and targeting) #639.

### DIFF
--- a/pkg/c3/defs.c
+++ b/pkg/c3/defs.c
@@ -1,5 +1,19 @@
 #include "defs.h"
 
+c3_s
+c3_sift_short(c3_y buf_y[2]);
+c3_w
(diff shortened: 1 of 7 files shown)


pkova (approved):

pkova:
For posterity

@ripperi noticed that having a fakezod and a fake planet up on the same machine on any commit after e135995ba2aac27424b5b4c230c635af00a0cac7 would cause both ships to be at 100% cpu on the king. They were sending stun requests to eachother at full blast for some reason. It is unclear why this PR fixes the issue, but it does.

PR #663, opened 2024-06-27
vere: support .run in subprocess replay

joemfb:
This PR fixes an issue introduced in #654, whereby docked binaries could not replay.


matthew-levan (approved):
LGTM. I see how docked piers would've simply printed `invalid command` due to the extra argument with the pier path (`u3_Host.dir_c` in my original code). Your argument preparation refactor looks good, but I haven't tested it.

PR #667, opened 2024-07-02
Mesa route switching

belisarius222:
Add route switching to Directed Messaging.  Also refactor congestion control as needed.

Route switching means being able to switch between the indirect route that goes through the publisher's galaxy and the direct route to the publisher.


belisarius222:
punchlist (to be added to):
- get packets flowing after cleanup
- get basic route switching to work
- check on congestion control, does it work right?
- make congestion control not forget its state when Arvo re-sends initial packets
-

pkg/vere/io/mesa.c
```diff
@@ -1886,28 +1886,62 @@ _hear_peer(u3_mesa* sam_u, u3_peer* per_u, u3_lane lan_u, c3_o dir_o)
...
+{
+  c3_w win_w = _mesa_req_get_cwnd(req_u);
+  u3_mesa_pict* nex_u = req_u->pic_u;
+  c3_w nex_w = req_u->nex_w;
+  for ( int i = 0; i < win_w; i++ ) {
+    c3_w fra_w = nex_w + i;
+    if ( fra_w >= req_u->tot_w ) {
+      break;
+    }
+    nex_u->pac_u.pek_u.nam_u.fra_w = nex_w + i;
```
belisarius222, line 1902: I think it would be a little clearer to do `nex_u->pac_u.pek_u.nam_u.fra_w = fra_w;` instead.

PR #669, opened 2024-07-04
disk: update epoc.txt and vere.txt atomically

### LINKED ISSUE #640: Ships failing to start with "disk: unknown epoch version"
We've seen a handful of ships on the Tlon Hosting platform have some issues with the new `chop` and `roll` setup. Sometimes the newest epoch doesn't have anything set inside its `epoc.txt` file; it's just an empty file. 

We first run `roll`, and then run `chop` as part of scheduled maintenance on a ship. 

In about ~10 ships now, when we run `roll`, the `epoc.txt` file in the latest epoch directory ends up with nothing written to it. The next time we try to start the ship, it crashes with this log, because the file is empty:
```
~
urbit 3.0
boot: home is /pier/<ship-name>
disk: unknown epoch version: '', expected '1'
upgrade runtime version
unable to open event log
```

If I go into `.urb/log` inside the pier and find the latest epoch, and then put in the `1` that vere is expecting, the ship then boots successfully.

pkova:
Fixes #640

### DIFF
--- a/pkg/vere/disk.c
+++ b/pkg/vere/disk.c
@@ -1723,9 +1723,9 @@ _disk_epoc_load(u3_disk* log_u, c3_d lat_d)
       return _epoc_gone;
     }
 
-    if (  (1 != sscanf(ver_c, "%" SCNu32 "%n", &ver_w, &car_i))
-       && (0 < car_i)
-       && ('\0' == *(ver_c + car_i)) )
+    if ( !(  (1 == sscanf(ver_c, "%" SCNu32 "%n", &ver_w, &car_i))
+          && (0 < car_i)
+          && ('\0' == *(ver_c + car_i)) ) )
     {
       fprintf(stderr, "disk: failed to parse epoch version: '%s'\r\n", ver_c);
       return _epoc_fail;
--- a/pkg/vere/disk.c
+++ b/pkg/vere/disk.c
@@ -1126,23 +1126,36 @@ u3_disk_epoc_zero(u3_disk* log_u)
   c3_c epi_c[8193];
   c3_c epv_c[8193];
   snprintf(epi_c, sizeof(epv_c), "%s/epoc.tmp", epo_c);
-  FILE* epv_f = fopen(epi_c, "w");  // XX errors
-  fprintf(epv_f, "%d", U3E_VERLAT);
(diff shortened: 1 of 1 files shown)


pkg/vere/disk.c
```diff
@@ -1123,18 +1123,26 @@ u3_disk_epoc_zero(u3_disk* log_u)
...
   c3_c epv_c[8193];
-  snprintf(epv_c, sizeof(epv_c), "%s/epoc.txt", epo_c);
-  FILE* epv_f = fopen(epv_c, "w");  // XX errors
+  snprintf(epi_c, sizeof(epv_c), "%s/epoc.tmp", epo_c);
+  FILE* epv_f = fopen(epi_c, "w");  // XX errors
   fprintf(epv_f, "%d", U3E_VERLAT);
   fclose(epv_f);
+  snprintf(epv_c, sizeof(epv_c), "%s/epoc.txt", epo_c);
+  rename(epi_c, epv_c);
+  c3_unlink(epi_c);
```
joemfb, line 1134: I don't think there's any reason to write this metadata in one place and then `rename()` it. That operation is atomic, but that doesn't guard against anything that we care about -- we're just creating a new epoc here, not updating meaningful data "in place".

I think the issue is simply that these version files are not being synced to disc. After `fwrite()`, we need to call `fflush()` and then `c3_sync(fileno(...))`.

joemfb: After further discussion, there is one possible advantage to `rename()` -- any parsing errors in reading epoc.txt would then indicate tampering with the file, not a crash while writing it (or the current problem, failure to sync). The practical result would be that, in the case that vere crashes while creating a new epoc but before `rename()`-ing epoc.txt into place, that epoc would be automatically removed on startup as opposed to producing a fatal error that leaves it in place.

matthew-levan: Yes, I just tested this with a fresh fakezod. I simulated a crash before `rename()` of the `epoc.tmp` to `epoc.txt` file during a `roll`, thereby leaving only the snapshot files and an `epoc.tmp` file in the new epoch's folder. Afterwards, I restarted the ship with `./urbit zod` and it successfully detected a bogus epoch, removed it, and booted from the latest valid epoch.

PR #671, opened 2024-07-09
trace: when profiling, dump a sampling trace

Fang-:
Whenever we run under the profiling flag (`-P` `--profile`), we now collect sampling traces into `.urb/put/trace/dump.txt`.

We take care to format the output in a way that's compatible with flamegraph tooling. Specifically, we avoid writing into the first couple lines of the file, and suffix every trace with a counter (always 1), to mimic dtrace output.

(See also [brendangregg/FlameGraph](https://github.com/brendangregg/FlameGraph).)

We have not yet cleaned up old code in `u3t_samp`.

---

After running some ship with `--profile`, we can use [the flamegraph scripts](https://github.com/brendangregg/FlameGraph) to turn the dump into an svg:

```bash
./stackcollapse.pl ~/piers/some-ship/.urb/put/trace/dump.txt | ./flamegraph.pl > dump.svg
```

If the ship ran `-build-file %/sys/hoon/hoon` under the profiler, the resulting svg might look something like this:

image

### DIFF
--- a/pkg/noun/trace.c
+++ b/pkg/noun/trace.c
@@ -23,6 +23,9 @@ static c3_o _ct_lop_o;
 /// Nock PID.
 static pid_t _nock_pid_i = 0;
 
+/// dump trace file.
+static FILE* _filt_u = NULL;
+
 /// JSON trace file.
 static FILE* _file_u = NULL;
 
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
This is sick, love it.

Issue #659, opened 2024-06-12
Crash reporting from vere


jalehman:
We currently have no quantitative measure for how stable Urbit is. Having the runtime submit a report of some kind to a server when it crashes, even if light on details, would allow us to begin measuring exactly how frequently crashes occur. It would be great to know what kind of crash occurred as well, but even knowing that a crash happened would be useful.

This is the kind of feature that should be opt-in, e.g. with a runtime flag. We could cover a large portion of the running Urbit fleet by working with hosting providers to turn this on.

PR #674, opened 2024-07-11
Add stacktrace

ripperi:
Related to #659 

Adds backtrace printing to SIGSEGV signal handling. Tries to print file and line number with libbacktrace, falls back to using standard backtrace when debug info is unavailable.

Before:
Screenshot 2024-07-11 at 15 58 52

After (with dsyms):
Screenshot 2024-07-24 at 17 11 56

After (without dsyms):
Screenshot 2024-07-24 at 17 14 31


pkg/noun/manage.c
```diff
@@ -728,6 +738,156 @@ u3m_dump(void)
...
+  }
+
+  if ( data.fail == 1 ) {
+    void*  array[100];
+    c3_c** strings;
+    size_t size = backtrace(array, 100);
+
+    strings = backtrace_symbols(array, size);
+
+    if ( strings == NULL ) {
```
ripperi, line 846: `|| strings[0] == NULL`

PR #690, opened 2024-07-31
Aor jet

Quodss:
`++aor` jet. 10x speed up with this input:

```hoon
(aor 'aaaaaaaaaaaaaaaa' 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
```

### DIFF
--- a/pkg/noun/jets/c/aor.c
+++ b/pkg/noun/jets/c/aor.c
@@ -1,0 +1,83 @@
+/// @file
+
+#include "jets/q.h"
+#include "jets/w.h"
+
+#include "noun.h"
+
+
+  static u3_atom
+  _aor_tramp(u3_noun a,
+             u3_noun b)
+  {
+    if ( c3y == u3r_sing(a, b) ) {
+      return c3y;
+    }
+    else {
+      if ( c3y == u3ud(a) ) {
+        if ( c3y == u3ud(b) ) {
+          c3_w len_a, len_b, i = 0;
+          c3_y *a_bytes = u3r_bytes_all(&len_a, a);
+          c3_y *b_bytes = u3r_bytes_all(&len_b, b);
+          while ( (i < len_a) && (i < len_b) ) {
+            c3_y a_slice=a_bytes[i], b_slice=b_bytes[i];
+            if (a_slice != b_slice) {
+              return u3qa_lth(a_slice, b_slice);
+            }
+            i++;
+          }
+          return u3m_bail(c3__fail);
+        }
+        else {
+          return c3y;
+        }
+      }
+      else {
+        if ( c3y == u3ud(b) ) {
+          return c3n;
+        }
+        else {
+          if ( c3y == u3r_sing(u3h(a), u3h(b)) ) {
+            return 3;
+          }
+          else {
+            return 2;
+          }
+        }
+      }
+    }
+  }
+
+  u3_noun
+  u3qc_aor(u3_noun a,
+           u3_noun b)
+  {
+    u3_noun out = 4;
+    while ( 1 ) {
+      out = _aor_tramp(a, b);
+      if (out == 2) {
+        a = u3h(a);
+        b = u3h(b);
+      }
+      else if (out == 3) {
+        a = u3t(a);
+        b = u3t(b);
+      }
+      else {
+        return out;
+      }
+    }
+  }
+  
+  u3_noun
+  u3wc_aor(u3_noun cor)
+  {
+    u3_noun a, b;
+
+    if ( c3n == u3r_mean(cor, u3x_sam_2, &a, u3x_sam_3, &b, 0) ) {
+      return u3m_bail(c3__exit);
+    } else {
+      return u3qc_aor(a, b);
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


pkg/noun/jets/c/aor.c
```diff
@@ -6,87 +6,70 @@
...
+            }
+            c3_w len_min = c3_min(len_a_w, len_b_w);
+            for (c3_w i_w = 0; i_w < len_min; i_w++) {
+              for (c3_w j = 0; j < 4; j++) {
+                a_y = (a_words[i_w] >> (8*j)) & 0xFF;
+                b_y = (b_words[i_w] >> (8*j)) & 0xFF;
+                if ( a_y != b_y ) return __(a_y < b_y);
+              }
+            }
+            return __(len_a_w < len_b_w);
```
joemfb, line 49: if these lengths are equal, i think we have to compare the remainders numerically.

pkg/noun/jets/c/aor.c
```diff
@@ -6,87 +6,70 @@
...
           else {
-            return 2;
+            if ( c3y == u3r_sing(u3h(a), u3h(b)) ) {
+              a = u3t(a);
+              b = u3t(b);
+            }
+            else {
+              a = u3h(a);
+              b = u3h(b);
+            }
```
joemfb, line 67: i think this would be easier to follow if the non-atom cases came first, like the hoon.

joemfb (changes requested):

pkg/noun/jets/c/aor.c
joemfb: Strike that, i misread.

pkg/noun/jets/c/aor.c
Quodss: done, the code flow now follows ++aor definition in hoon

Quodss:
Please don't merge it just yet: I want to try adding little-endian optimization

joemfb (approved):
this will do

Quodss:
I included `portable.h` in the jet with an endianness check and an optimized version of the jet. It squizzes out ~17% of performance gains in the case of some ridiculous inputs (`[a=(crip (reap 1.000.000 'a')) b=(crip ['a' (reap 1.000.000 'a')])]`)

Looks like the speed up comes from [removing `shl` and `and` instructions](https://godbolt.org/z/PacKb4G54), which I expected to be optimized away in the first version of the jet.

pkg/noun/jets/c/aor.c
```diff
@@ -0,0 +1,134 @@
...
+          c3_w len_a_y, len_b_y;
+          c3_y *a_bytes, *b_bytes;
+          c3_y a_y, b_y;
+          if ( c3y == u3a_is_cat(a) ) {
+            len_a_y = 4;
+            a_bytes = (c3_y*)&a;
+          }
+          else {
+            u3a_atom* a_u = u3a_to_ptr(a);
+            len_a_y = (a_u->len_w)*4;
```
joemfb, line 40: This is wrong (as are your other length calculations in this function). The byte-length is not `4 * word-length`, it needs to precisely measure the most-significant word. You you use `u3r_met(3, a)` or calculate yourself.

joemfb (changes requested):
Those `#define`'s are vestigial, and this codebase is little-endian only. So you can just use your little-endian version and drop the other one.

pkg/noun/jets/c/aor.c
Quodss: my bad, fixed it; in my defense this didn't affect the result since in the case of different byte size but same word size the function would've returned on comparison of a non-zero byte with one of the leading zero bytes instead of length comparison after the loop

pkg/noun/jets/c/aor.c
```diff
@@ -0,0 +1,69 @@
...
+            u3a_atom* b_u = u3a_to_ptr(b);
+            b_bytes = (c3_y*)(b_u->buf_w);
+          }
+          c3_w len_min = c3_min(len_a_y, len_b_y);
+          for (c3_w i_y = 0; i_y < len_min; i_y++) {
+            a_y = a_bytes[i_y];
+            b_y = b_bytes[i_y];
+            if ( a_y != b_y ) return __(a_y < b_y);
+          }
+          return __(len_a_y < len_b_y);
```
joemfb, line 52: This can shortcircuit faster by only comparing bytes if the lengths are equal. If you get to the end of the loop just return `c3y`.

joemfb (changes requested):
Your naming conventions don't matching u3's suffix patterns. The name suffix should match the type of the variable, regardless of whether it's a pointer: `c3_y -> foo_y`, `c3_w foo_w`, &c.

pkg/noun/jets/c/aor.c
Quodss: and if the lengths are not equal, how can I skip comparing bytes?

```
> (aor 'z' 'aa')
%.n
> (aor 'a' 'zz')
%.y
```

are you sure you are not thinking about ++dor?

joemfb: My bad, you're right. Retracted.

joemfb (approved):

PR #693, opened 2024-08-02
fix linux builds

belisarius222:
First I couldn't build on Ubuntu 20.04 at all, then I couldn't build in debug mode.  These changes fix both.

This is a draft PR because I'm not sure if we need to make these changes conditional on MacOS vs. Linux.


joemfb:
Obsolete as of #719

PR #694, opened 2024-08-02
stun: fix stun

pkova:
Informal pinging has been broken since #623, luckily this was unreleased, but existed edge.


joemfb (approved):
great catch, my apologies

pkova:
For posterity, even testers on edge would not have experienced lessened connectivity because of this because if informal pinging fails we fall back to formal pinging.

PR #703, opened 2024-08-09
Fix jet mismatches and crashes

### LINKED ISSUE #700: `++cue` jet crashes the runtime when the sample is a cell
Calling `++cue` with a cell sample causes the runtime to crash non-recoverably:

```
(slum cue *^)
```

This is due to the fact that `u3s_cue_atom` is called [without checking if its argument is an atom](https://github.com/urbit/vere/blob/47d2fedefa4bd8830df15e6f19278e721ca0e3b5/pkg/noun/jets/e/cue.c#L18).

Quodss:
Resolves #700.

### DIFF
--- a/pkg/noun/jets/a/add.c
+++ b/pkg/noun/jets/a/add.c
@@ -6,7 +6,7 @@
 
 #include "noun.h"
 
-u3_noun
+u3_atom
 u3qa_add(u3_atom a,
          u3_atom b)
 {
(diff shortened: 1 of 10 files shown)


joemfb:
While you're at it, the `+swp` jet from urbit/vere#577 doesn't check the bloq size. I keep forgetting to fix that.

Quodss:
This PR will also fix comparison jet mismatches (lth, gth, lte, gte, min, max) when both inputs are equal cells

Quodss:
A couple of other mismatches I found: `++dvr` crashes the runtime unrecoverably, `++div` exits on input `[0 0 0]` even though it should return 0. Proposal: make the core "type checks" more rigorous in u3w-prefixed jet functions, and return u3_none instead of exiting. I'll commit an example with ++add

Quodss:
Closed as duplicate of #836

PR #715, opened 2024-08-21
vere64

matthew-levan:
Cribbed from #794.

This PR introduces 64-bit Vere. The principle changes involved have to do with the introduction of a new type, c3_n, which is 32 bit or 64 bit depending on if it was compiled with -DVERE64.

- [x] define final typedefs, update usages accordingly
- [x] every appearance of _tmp in pkg/noun (excluding pkg/noun/jets) (~800 locations)
  - [x] `trace.{c,h}`
  - [x] `ship.{c,h}`
  - [x] `nock.h`
  - [x] `version.h`
  - [x] `imprison.{c,h}`
  - [x] `jets.{c,h}`
  - [x] `allocate.{c,h}`
  - [x] `retrieve.{c,h}`
  - [x] `*_tests.c`
- [x] every appearance of _tmp in pkg/noun/jets
- [x] ~~look at diff of pre and post VERE64 compilation with -Wconversion flag. Investigate every instance.~~
- [x] run with llvm addr sanitizer, ubsan
  - [x] found [issues](https://gist.github.com/matthew-levan/7b7597f71e6fa920912a6b6c4d438531) with asan and ubsan as of 61fc9dee67295692b1881996d07a67cf971c8c7a (lmdb addressed in #820)
- [ ] audit every jet
- [ ] migrations: bidirectional (use MSB of version tag to mark if vere64 or not? maybe also add metadata for pointer compression)
- [x] integrate Joe's page based allocator
- [x] new rsignal_setjmp/longjmp may have volatility problems (talk to Joe. maybe use atomics?)
- [x] pkg/vere/noun-tests is completely broken
  - [x] ensure noun tests make sense/are worth having
- [x] every appearance of _tmp in any other subdirectory (~1000 locations)
- [x] c3_w_clz, ctz, pc occurrences (we should change current occurrences to _tmp as we did with u3r_word, etc and go through them 
- [x] c3_etch_word, c3_sift_word occurrences (also change to _tmp)
- [x] make sure everything works (C tests and hoon tests included) still for vere32
- [x] revendor murmur3 with a size_t len
  - [x] urbit/urbit#7186
  - [x] #842 
- [x] the event log and other lmdb state should not need migration. we should double check lmdb.c, disk.c
- [x] there's a lot of bloq size related things in the jets that could be changed to greater than 32 for the 32 bit case
- [x] `(mug .(eny 0, now 0))` equality test for fakezods booted in 32-bit and 64-bit modes


pkg/noun/jets.c
```diff
@@ -671,7 +672,7 @@ _cj_spot(u3_noun cor, u3_weak* bas)
 static u3p(u3j_fink)
 _cj_cast(u3_noun cor, u3_noun loc)
 {
-  c3_w     i_w = 0;
+  c3_w_tmp     i_w = 0;
```
matthew-levan: `u3j_fink` structures are runtime-only caches (i.e., they're not saved in snapshots), so we can safely use the architecture-dependent `c3_l` for this counter (and the `len_w` field of `u3j_fink` itself), which is used to traverse the "fine check" core chain. Using `c3_l` doesn't break existing looms, will improve performance due to architectural consistency, can handle larger chain depths on 64-bit machines if needed, and the loop counters will match the `fin_u->len_w` field.

matthew-levan (commented):
Comments on my changes to `jets.{c,h}`.

pkg/noun/nock.c
```diff
@@ -554,7 +554,7 @@ _n_arg(c3_y cod_y)
       return sizeof(c3_s);
 
     case SWIP: case SWIN:
-      return sizeof(c3_l);
+      return sizeof(c3_l_tmp);
```
matthew-levan, line 557: The bytecode format is architecture-independent and should be pinned to a static 32-bit width.

matthew-levan:
Closing in favor of #970.

PR #719, opened 2024-09-24
zig build system

ripperi:


pkova:
Some context here for future travellers:

This PR is a complete rewrite of the build system for vere, switching from Bazel to the Zig build system. We get the following benefits from doing this:

1. The ability to compile vere on any target will have exactly one dependency, Zig. We do not need `libtool`, `autoconf-archive`, Xcode cli, or even git to compile vere anymore.
2. Cross-compilation support for any target we support from any other target. To compile binaries for `aarch64-macos`, `x86_64-macos`, `aarch64-linux` and `x86_64-linux` you simply run `zig build -Dall`.
3. The zig c compiler wraps clang with some reasonable defaults. One of them is the undefined behavior sanitizer turned on by default in debug builds. This resulted in us finding and fixing many of them, see #720 and #723.
4. We upgrade clang from 15 -> 18. Note that zig build uses clang internally so now we compile our linux binaries using clang instead of gcc.
5. Get rid of Bazel.

When trawling through the build some interesting discoveries were made. One of them was that GMP was built without assembly on macos-aarch64:
https://github.com/urbit/vere/blob/ec54a66e43007e010fa3643320c38ad6db267a0d/bazel/third_party/gmp/gmp.BUILD#L23

There are also other miscellaneous quality of life improvements such as not "compiling" the ivory pill with xxd but just checking in the C file which saves around 70 seconds on my mac m2.

The new build process is installing zig 0.13 and typing `zig build`. A list of build-time options can be displayed with `zig build -h`. Building a release binary with optimizations is done with `zig build -Doptimize=ReleaseFast`. Cross-compiling all supported targets is done with `zig build -Dall`.

pkova (approved):

PR #722, opened 2024-09-27
Lick Pointer bug

mopfel-winrux:
Fixes a bug where you couldn't access any lick ports after opening the first one due to stupid pointer allocations


joemfb (approved):
good catch!

PR #733, opened 2024-10-10
Fix generated GMP assembly sources on x86_64 linux

ripperi:


joemfb:
What was different about the command use to generate these sources? Can that be captured in a README?

ripperi:
> What was different about the command use to generate these sources? Can that be captured in a README?

It seemed I had forgotten to provide at least the `--host=x86_64-linux-musl` configure option. We should definitely document the exact options we use for each target

PR #740, opened 2024-10-28
vere: adds %quic loom measurement, support it and |mass in conn

joemfb:
Resolves #


joemfb:
These changes should be ready to go. I've cleaned up crash issues with the async-request lifecycle, and refactored some of the interfaces from #675.

To work around the absence of `-W` on macos `nc`, I've used `socat -t300 -,ignoreeof UNIX-CONNECT:$SHIP/.urb/conn.sock`.

The build issues here are a little strange. I think that dependencies are not properly declared between packages in the zig build system. I had local build issues adding new motes and using them in `serf.c` -- I ended up manually removing the zig cache.

pkova (approved):

PR #760, opened 2025-02-12
mesa: improvements

pkova:
This PR contains several big improvements to the mesa driver:

* We don't allocate nouns all over the place anymore.
* To enable the previous we don't use our HAMT anymore, instead we use a [off-the-shelf hashtable](https://github.com/JacksonAllan/Verstable).
* We don't do individual `malloc` and `free` calls all over the place anymore. The allocations have been grouped into lifetimes each with their own memory arena.
* We don't allocate at all when we don't need to.

As a result we go from around 15 megabytes/second to 190 megabytes/second bandwidth for localhost peeking on my macbook M2. Corresponding speeds between two terrible 1 cpu 2gb VM:s situated on the other sides of the Atlantic go from around 2 megabytes/s to 18 megabytes/s.

There is still room to optimize the speed of directed messaging but this is certainly good enough for an initial release. The main thing that's somewhat wonky still is the congestion control.

### DIFF
--- a/pkg/vere/io/mesa.c
+++ b/pkg/vere/io/mesa.c
@@ -1297,6 +1297,9 @@ _realise_lane(u3_noun lan) {
     if ( (c3n == u3_Host.ops_u.net) ) {
       lan_u.sin_addr.s_addr = htonl(0x7f000001);
       lan_u.sin_port = htons(_ames_czar_port(lan));
+    } else {
+      lan_u.sin_addr.s_addr = htonl(u3_Host.imp_u[lan]);
+      lan_u.sin_port = htons(_ames_czar_port(lan));
     }
   } else {
     u3_noun tag, pip, por;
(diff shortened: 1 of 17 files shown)


pkg/vere/io/mesa.c
```diff
@@ -3064,7 +3007,12 @@ _mesa_io_talk(u3_auto* car_u)
   }
 
   u3_Host.wax_u.data = sam_u;
-  uv_udp_recv_start(&u3_Host.wax_u, _ames_alloc, _mesa_recv_cb);
+  uv_udp_recv_start(&u3_Host.wax_u, _mesa_alloc, _mesa_recv_cb);
+
+  c3_i rec_i = 2 * 1024 * 1024;
+  uv_recv_buffer_size((uv_handle_t*)&u3_Host.wax_u, &rec_i);
+
+  uv_send_buffer_size((uv_handle_t*)&u3_Host.wax_u, &rec_i);
```
joemfb, line 3015: It seems like we should read the buffer sizes (setting `rec_i = 0`) before setting them, to make sure we're not shrinking them.

PR #769, opened 2025-02-28
play: add `--watch-replay` flag

### LINKED ISSUE #679: Ship having issues started when passed the `-t` flag
When I tried to boot ~nattyv using the `-t` flag I get the following error

```
BOOT SHIP
Running with no STDIN
urbit -t -p 30255 --http-port 21183 --loom 32 --snap-time 60 nattyv
~
play: god save the king! committing sudoku...
Terminated
```

I have to replay without using the `-t` flag. The ship boots afterwards

matthew-levan:
Resolves #679. Read this [comment](https://github.com/urbit/vere/issues/679#issuecomment-2706929997) for a summary of the underlying issue.

This PR adds the `--watch-replay` flag to the `play` subcommand. 

When `--watch-replay` is passed to `play`, a watcher thread is spawned to ensure process orphaning does not occur. This option is only helpful when replaying events during normal pier startup, where we spawn a child process, so it's off by default. When we start a pier normally (i.e., without the `play` subcommand), we turn it on in `_cw_play_fork`. 

Keeping `--watch-replay` off by default _prevents_ operators that run `play` with no `STDIN` from experiencing issue #679, like so:
```
matt@mbp14 vere % ./urbit play zod < /dev/null
play: god save the king! committing sudoku...
```

No affordance is granted for disabling the watcher thread during normal pier startup, since we always spawn a child process to replay and want to prevent process orphaning as far as possible.


pkg/vere/main.c
```diff
@@ -2467,6 +2473,7 @@ _cw_play_fork(c3_d eve_d, c3_d sap_d, c3_o mel_o, c3_o sof_o, c3_o ful_o)
 
     argv[i_z++] = u3_Host.wrk_c;
     argv[i_z++] = "play";
+    argv[i_z++] = "--watch-replay";
```
joemfb, line 2476: does `argv` need to be declared as `[13]` now?

pkg/vere/vere.h
```diff
@@ -321,6 +321,7 @@
 
         c3_o    beb;                        //  --behn-allow-blocked
         c3_z    siz_i;                      //  --lmdb-map-size
+        c3_o    wat_o;                      //  --watch-replay
```
joemfb, line 324: This should either be removed, or integrated into `_cw_play_fork()` (controlling whether `--watch-replay` is set). For the latter, I'd make it default on and have the flag disable the watcher thread.

matthew-levan:
Closed in favor of #775.

PR #771, opened 2025-03-04
Spin Hint

mopfel-winrux:
This PR adds a `%spin` hint to vere. This hint provides a labeled compute stack in shared memory. It also includes an SSE endpoint located at `/~_~/spin` which will push the current non-empty stack to the SSE endpoint as a comma separated value string. The endpoint is set to sample the spin stack at 10kHz (0.1 ms).

To test this:
- boot a fake ship
- login to the webui
- navigate to `/~_~/spin`
- in dojo enter
```
=g |=  l=(list @)
  ~>  %spin.[(scot %ud (lent l))]
  (turn l |=(a=@ ~>(%spin.[(scot %ud a)] (met 3 (jam .)))))
```
- run `(g ~[1 2 3 4 5 6 7 8 9 10])` in dojo.
- On the website you should see:
image


pkg/noun/nock.h
```diff
@@ -4,12 +4,17 @@
...
 #include <stdio.h>
+#include <fcntl.h>
+#include <unistd.h>
 
 #include "c3/c3.h"
 #include "jets.h"
 #include "types.h"
 #include "zave.h"
 
+#define SLOW_STACK_NAME  "/spin_stack_page"
```
mopfel-winrux, line 15: This actually might get shared with every urbit running on the host. we want to have some distinguisher (pier name?) in this to help

pkg/noun/nock.c
```diff
@@ -3191,3 +3207,64 @@ u3n_nock_an(u3_noun bus, u3_noun fol)
...
+{
+  if ( c3n == u3ud(nam) ) {
+    return;
+  }
+
+  c3_w  met_w = u3r_met(3, nam);
+  
+  // Exit if full
+  if ( sizeof(stk_u->dat_y) < stk_u->off_w + met_w + sizeof(c3_w) ) {
+    return;
```
mopfel-winrux, line 3251: If this happens we need a way to make sure that we don't pop unnecessarily.

mopfel-winrux:
As per @joemfb, we want to push the current pointer onto the road system and if the inner road bails, restore it to the previous one

pkg/noun/nock.h
mopfel-winrux: This now uses the pid (for king) and ppid (for serf) to uniquely identify each pier uniquely

pkg/noun/nock.c
mopfel-winrux: This is fixed by adding an extra member in the shared page to store the overflow count

mopfel-winrux:
> As per @joemfb, we want to push the current pointer onto the road system and if the inner road bails, restore it to the previous one
> 
> As per @frodwith, the timer should only go off iff something is connected

Both of these are done. Going to mark this draft as ready for review

pkg/vere/io/http.c
```diff
@@ -1462,6 +1542,42 @@ _http_req_prepare(h2o_req_t* rec_u,
...
+                               u3i_string("no-cache")),
+                          u3nc(u3i_string("Connection"),
+                               u3i_string("keep-alive")),
+                          u3_none);
+
+    _http_start_respond(req_u, 200, hed, u3_nul, c3n);
+
+    //TODO  auth token may expire at some point. if we want to close the
+    //      slogstream when that happens, we need to store the token that
+    //      was used alongside it...
```
Fang-, line 1576: ftr the slog endpoint has the same problem. I don't think it's particularly pressing.

pkg/vere/io/http.c
```diff
@@ -1462,6 +1542,42 @@ _http_req_prepare(h2o_req_t* rec_u,
...
+    u3_hreq* req_u = _http_req_prepare(rec_u, _http_spin_new);
+    u3_noun  hed   = u3nl(u3nc(u3i_string("Content-Type"),
+                               u3i_string("text/event-stream")),
+                          u3nc(u3i_string("Cache-Control"),
+                               u3i_string("no-cache")),
+                          u3nc(u3i_string("Connection"),
+                               u3i_string("keep-alive")),
+                          u3_none);
+
+    _http_start_respond(req_u, 200, hed, u3_nul, c3n);
```
Fang-, line 1572: The slog endpoint suffers from this also, so probably out of scope for this PR, but I'll make a note here:

Probably want the initial response to contain a heartbeat, similar to and solving the same thing as urbit/urbit#7080.

pkg/vere/io/http.c
```diff
@@ -443,6 +450,24 @@ _http_req_unlink(u3_hreq* req_u)
...
+  req_u->hon_u = hon_u;
+  req_u->seq_l = hon_u->seq_l++;
+  req_u->nex_u = fig_u->siq_u;
+
+  if ( 0 != req_u->nex_u ) {
+    req_u->nex_u->pre_u = req_u;
+  } else {
+    uv_timer_start(fig_u->sin_u, _http_spin_timer_cb, SPIN_TIMER, 0);
+  }
+  fig_u->siq_u = req_u;
```
Fang-, line 468: Here we start the timer if the request list was empty. `_http_spin_timer_cb` doesn't set a new timer if the request list is empty. But `_http_spin_unlink` doesn't _unset_ the timer. So `_http_spin_link` could get called while the request list is empty _and_ there's still an outstanding timer.

Does `uv_timer_start` protect against overwriting `sin_u`? If not, we might start a second timer (if a request comes in between the unlink and the og timer firing) that'll propagate itself alongside the og timer through the callback logic.

Fang- (commented):
Some nits and a concern about potential timer doubling.

pkg/vere/io/http.c
pkova: `uv_timer_start` overwrites a pending timer for the handle if one is already pending, so it'll just be one timer no matter what.

pkg/vere/io/http.c
```diff
@@ -2915,6 +3115,23 @@ u3_http_io_init(u3_pier* pir_u)
...
+  }
+
+  htd_u->stk_u = mmap(NULL, PSIZE, 
+                      PROT_READ | PROT_WRITE, MAP_SHARED, shm_fd, 0);
+  
+  if ( MAP_FAILED == htd_u->stk_u ) {
+    perror("mmap failed");
+    u3_pier_bail(car_u->pir_u);
+  }
+
```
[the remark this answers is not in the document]
mopfel-winrux: Not exactly sure how to do this, do I want to have a function thats `u3t_spin* u3t_sstack_open()` which just returns a pointer to a brandnew open spin stack data structure?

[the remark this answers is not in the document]
joemfb: I think so, and just return a null pointer on error.

pkg/noun/nock.c
```diff
@@ -1931,6 +1936,14 @@ _n_hint_fore(u3_cell hin, u3_noun bus, u3_noun* clu)
       *clu = u3nt(u3k(tag), *clu, now);
     } break;
 
+    case c3__spin: {
+
+      //u3t_slog_cap(1, u3i_string("spin"), hin);
+      u3n_sstack_push(*clu);
+      u3z(*clu);
```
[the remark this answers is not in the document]
mopfel-winrux: done

pkg/noun/nock.c
```diff
@@ -3191,3 +3207,80 @@ u3n_nock_an(u3_noun bus, u3_noun fol)
...
+       sizeof(stk_u->dat_y) < stk_u->off_w + met_w + sizeof(c3_w) ) {
+    stk_u->fow_w++;
+    return;
+  }
+
+  u3r_bytes(0, met_w, (c3_y*)(stk_u->dat_y+stk_u->off_w), nam);
+  stk_u->off_w += met_w;
+
+  memcpy(&stk_u->dat_y[stk_u->off_w], &met_w, sizeof(c3_w));
+  stk_u->off_w += sizeof(c3_w);
```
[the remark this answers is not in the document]
mopfel-winrux: Done

pkg/vere/io/http.c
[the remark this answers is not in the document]
mopfel-winrux: Did the above.

pkg/vere/io/http.c
```diff
@@ -2695,14 +2816,76 @@ _http_stream_slog(void* vop_p, c3_w pri_w, u3_noun tan)
...
+  if ( 0 != siq_u ) {
+    c3_c* buf_c     = c3_malloc(1025);
+    u3n_spin* stk_u = htd_u->stk_u;
+    c3_w pos_w      = stk_u->off_w;
+    c3_w out_w      = 0;
+
+    while (pos_w > 0) {
+      c3_w  len_w;
+      pos_w -=4;
```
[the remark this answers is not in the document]
mopfel-winrux: Done

pkg/noun/manage.c
```diff
@@ -1005,6 +1005,10 @@ u3m_bail(u3_noun how)
     }
   }
 
+  // Reset the spin stack pointer
+  stk_u->off_w = u3R->off_w;
+  stk_u->fow_w = u3R->fow_w;
```
joemfb, line 1010: We can't reset here with the inner road spin-stack values, it needs to be done with the values from the outer road. That could be all the time in `u3m_fall()`, or only on errors (at each `_setjmp()` callsite, the various u3m_soft_* functions). That reset logic also needs to be duplicated in `_cm_signal_recover()`, where we handle errors after `u3m_signal()` (ie, `siglongjmp()`).

joemfb (changes requested):

pkg/noun/nock.c
```diff
@@ -3191,3 +3207,80 @@ u3n_nock_an(u3_noun bus, u3_noun fol)
...
+u3n_sstack_pop()
+{
+  if ( 0 < stk_u->fow_w ) {
+    stk_u->fow_w--;
+  } else {
+    c3_w len_w = (c3_w) stk_u->dat_y[stk_u->off_w - sizeof(c3_w)];
+    stk_u->off_w -= (len_w+sizeof(c3_w));
+  }
+}
+
```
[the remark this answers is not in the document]
mopfel-winrux: moved

pkg/noun/manage.c
mopfel-winrux: Note: Chatted with Joe, how this is done is fine because of how the values are set when the road gets created. I will still need to add this to `_cm_signal_recover()`

mopfel-winrux: Added this before the if statement in `_cm_signal_recover`

pkg/noun/manage.c
```diff
@@ -295,6 +295,9 @@ _cm_signal_recover(c3_l sig_l, u3_noun arg)
   tax = u3H->rod_u.bug.tax;
   u3H->rod_u.bug.tax = 0;
 
+  stk_u->off_w = u3R->off_w;
+  stk_u->fow_w = u3R->fow_w;
+
```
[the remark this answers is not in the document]
mopfel-winrux: Done

pkg/vere/io/http.c
```diff
@@ -2695,14 +2816,79 @@ _http_stream_slog(void* vop_p, c3_w pri_w, u3_noun tan)
...
+                             tan,
+                             c3_s2('\n', '\n'),
+                             u3_none);
+      u3_atom txt = u3qc_rap(3, lin);
+      u3_noun dat = u3nt(u3_nul, u3r_met(3, txt), txt);
+
+      while ( 0 != siq_u ) {
+        _http_continue_respond(siq_u, u3k(dat), c3n);
+        siq_u = siq_u->nex_u;
+      }
```
joemfb, line 2874: Need a `u3z(dat); u3z(lin);` after the loop to avoid leaking them. (You could also use or add `u3kc_rap()` and just deal with `dat`.

pkg/vere/io/http.c
```diff
@@ -459,6 +484,34 @@ _http_seq_link(u3_hcon* hon_u, u3_hreq* req_u)
...
+    if ( 0 != req_u->nex_u ) {
+      req_u->nex_u->pre_u = req_u->pre_u;
+    }
+  }
+  else {
+    fig_u->siq_u = req_u->nex_u;
+
+    if ( 0 != req_u->nex_u ) {
+      req_u->nex_u->pre_u = 0;
+    }
```
joemfb, line 505: should we have `else { uv_timer_stop(...); }` here?

pkg/vere/io/http.c
```diff
@@ -459,6 +484,34 @@ _http_seq_link(u3_hcon* hon_u, u3_hreq* req_u)
...
+  }
+
+  //  unlink from async scry request if present
+  //
+  if ( req_u->peq_u ) {
+    req_u->peq_u->req_u = 0;
+  }
```
joemfb, line 512: This is technically unnecessary, as the spin request will never cause a scry and `peq_u` will never be set. The same applies to `_http_seq_unlink()`. Not sure if it's worth keeping in either place, I don't see that changing.

pkg/vere/io/http.c
```diff
@@ -2695,14 +2816,79 @@ _http_stream_slog(void* vop_p, c3_w pri_w, u3_noun tan)
...
+      }
+
+      memcpy(&len_w, &stk_u->dat_y[pos_w], 4);
+      pos_w -= len_w;
+
+      if ( siz_w < out_w + 4 ) {
+         buf_c = c3_realloc(buf_c, siz_w*2);
+      }
+      memcpy(buf_c + out_w, "/", 1);
+      out_w +=1;
```
joemfb, line 2850: A very small nit, but I'd favor `buf_c[out_w++] = '/';`

pkg/vere/io/http.c
```diff
@@ -2695,14 +2816,79 @@ _http_stream_slog(void* vop_p, c3_w pri_w, u3_noun tan)
...
+
+      if ( siz_w < out_w + len_w ) {
+         buf_c = c3_realloc(buf_c, siz_w*2);
+      }
+
+      memcpy(buf_c + out_w, &stk_u->dat_y[pos_w], len_w);
+      out_w += len_w;
+      buf_c[out_w] = '\0';
```
joemfb, line 2858: This null terminator seems redundant, since you set one after the loop.

pkg/vere/serf.c
```diff
@@ -949,7 +949,7 @@ _serf_writ_live_exit(u3_serf* sef_u, c3_w cod_w)
       fclose(fil_u);
     }
   }
-
+
```
joemfb, line 952: The only change to this file is added whitespace.

pkg/vere/io/http.c
```diff
@@ -125,6 +132,7 @@ static void _http_start_respond(u3_hreq* req_u,
 
 static const c3_i TCP_BACKLOG = 16;
 static const c3_w HEARTBEAT_TIMEOUT = 20 * 1000;
+static const c3_w SPIN_TIMER = 100; // XX make this a command line arguement
```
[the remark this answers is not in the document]
joemfb: Seems like it'll be easy to add an argument later.

joemfb (changes requested):
A few nits:

pkg/vere/io/http.c
mopfel-winrux: Done

pkg/vere/io/http.c
mopfel-winrux: Removed both instances of this

pkg/vere/io/http.c
mopfel-winrux: done

pkg/vere/io/http.c
mopfel-winrux: This was from a debug session, removed

pkg/vere/serf.c
mopfel-winrux: cleaned up

pkg/vere/io/http.c
mopfel-winrux: added

pkg/noun/trace.c
```diff
@@ -1120,3 +1123,106 @@ u3t_etch_meme(c3_l mod_l)
...
+
+  memcpy(&stk_u->dat_y[stk_u->off_w], &met_w, sizeof(c3_w));
+  stk_u->off_w += sizeof(c3_w);
+  u3z(nam);
+}
+
+/* u3t_sstack_pop: pop a noun from the spin stack.
+*/
+void
+u3t_sstack_pop()
```
mopfel-winrux, line 1219: if stku is null return

pkg/noun/trace.c
```diff
@@ -1120,3 +1123,106 @@ u3t_etch_meme(c3_l mod_l)
...
+  munmap(stk_u, u3a_page);
+}
+
+/* u3t_sstack_push: push a noun on the spin stack.
+*/
+void
+u3t_sstack_push(u3_noun nam)
+{
+  if ( c3n == u3ud(nam) ) {
+    u3t_sstack_push(c3__cell);
```
joemfb, line 1195: Instead of recurring, this should be:

```
u3z(nam);
nam = c3__cell;
```

pkg/noun/trace.c
```diff
@@ -1120,3 +1123,106 @@ u3t_etch_meme(c3_l mod_l)
...
+u3t_sstack_exit()
+{
+  munmap(stk_u, u3a_page);
+}
+
+/* u3t_sstack_push: push a noun on the spin stack.
+*/
+void
+u3t_sstack_push(u3_noun nam)
+{
```
joemfb, line 1193: To handle cases where there's no shared memory stack (like running in the ivory pill):

```
if ( !stk_u ) {
  u3z(name);
  return;
}
```

pkg/noun/trace.c
```diff
@@ -1120,3 +1123,106 @@ u3t_etch_meme(c3_l mod_l)
...
+  memcpy(&stk_u->dat_y[stk_u->off_w], &met_w, sizeof(c3_w));
+  stk_u->off_w += sizeof(c3_w);
+  u3z(nam);
+}
+
+/* u3t_sstack_pop: pop a noun from the spin stack.
+*/
+void
+u3t_sstack_pop()
+{
```
joemfb, line 1220:
```
if (  !stk_u ) return;
```

pkg/vere/io/http.c
```diff
@@ -2915,6 +3108,14 @@ u3_http_io_init(u3_pier* pir_u)
     u3z(now);
   }
 
+  //Setup spin stack
+  htd_u->stk_u = u3t_sstack_open();
+  
+  if ( NULL == htd_u->stk_u ) {
+    u3l_log("failed to open spin stack");
+    u3_pier_bail(car_u->pir_u);
+  }
```
joemfb, line 3117: let's move this to `u3_http_io_talk()` to avoid race conditions, and make it a soft error (just print).

pkg/vere/io/http.c
```diff
@@ -1462,6 +1533,42 @@ _http_req_prepare(h2o_req_t* rec_u,
...
 }
 
+/* _http_spin_accept(): handle incoming http request on spin stack endpoint
+*/
+static c3_i
+_http_spin_accept(h2o_handler_t* han_u, h2o_req_t* rec_u)
+{
+  u3_hcon* hon_u = _http_rec_sock(rec_u);
+  c3_o     aut_o = _http_req_is_auth(&hon_u->htp_u->htd_u->fig_u, rec_u);
+
```
joemfb, line 1543: We should produce a 404 if there's no spin stack (`!htd_u->stk_u`).

pkg/noun/manage.c
```diff
@@ -2253,6 +2264,8 @@ u3m_init(size_t len_i)
       exit(1);
     }
 
+    u3t_sstack_init();
```
joemfb, line 2267: This should go somewhere else, this runs for every loom initialization, including the ivory pill. Maybe this should just go in the initialization of the serf?

joemfb (changes requested):

pkg/noun/trace.c
mopfel-winrux: Done

pkg/noun/manage.c
mopfel-winrux: Please disregard above, fixed the issue

pkg/vere/io/http.c
mopfel-winrux: Done

pkg/vere/io/http.c
mopfel-winrux: done

pkg/noun/trace.c
mopfel-winrux: Done

pkg/noun/trace.c
mopfel-winrux: Done

pkg/vere/io/http.c
```diff
@@ -2974,7 +2992,8 @@ _http_io_exit(u3_auto* car_u)
   u3h_free(htd_u->sax_p);
   u3h_free(htd_u->nax_p);
 
-  munmap(htd_u->stk_u, u3a_page);
+  if ( NULL != htd_u->stk_u )
+    munmap(htd_u->stk_u, u3a_page);
```
joemfb, line 2996: u3 style is to always uses braces (except for control flow on the same line as the if).

pkg/vere/io/http.c
```diff
@@ -2653,6 +2661,8 @@ _http_io_talk(u3_auto* car_u)
 {
   u3_httd* htd_u = (u3_httd*)car_u;
 
+
+
```
joemfb, line 2665: whitespace added.

pkg/noun/trace.c
```diff
@@ -1218,6 +1223,7 @@ u3t_sstack_push(u3_noun nam)
 void
 u3t_sstack_pop()
 {
+	if (  !stk_u ) return;
```
joemfb, line 1226: indentation is wrong here (maybe tabs instead of spaces?)

joemfb (changes requested):
These changes look good. A couple style nits:

pkg/vere/io/http.c
mopfel-winrux: done

pkg/noun/trace.c
mopfel-winrux: removed

pkg/vere/io/http.c
mopfel-winrux: done

joemfb (approved):
this will do

PR #773, opened 2025-03-06
Mars/Urth Split Restage and Deadlock Fix

matthew-levan:
Superseded by #810.

## Overview

Restages @joemfb's original work from urbit/urbit#5596. 

### Deadlock

The `libuv` deadlock bug (which was observed originally in 2022 and was the reason for the original PR's reversion) seems to have vanished. 

Why? Since I cannot reproduce the bug (after having run a live moon with this restaged work here for over 2 weeks now, and the bug was previously "easily and quickly reproducible" with live ships), I cannot determine the exact cause of its resolution (or if it's truly resolved or not!). Many things have changed since then, including the build system (`nix` to `bazel` to now `zig`), repository split, `musl` compiler version, and `libuv` version. 

Despite the uncertainty of why the deadlock bug has seemingly resolved, I propose to merge this to `edge` for further livenet testing by developers. If the bug rears its ugly head once again, further analysis and engineering can be done to fix it. If not, we can declare the bug effectively fixed.

### King/Serf vs. Mars/Urth

The Mars/Urth split is essentially a modification of Vere's dual process architecture, moving from a "king" and a "serf" to a "mars" and an "urth":

| Responsibility          | Serf | King | Mars | Urth |
|-------------------------|------|------|------|------|
| Persistence (snapshot)  | X    |      | X    |      |
| Persistence (event log) |      | X    | X    |      |
| Nock computation        | X    |      | X    |      |
| I/O                     |      | X    |      | X    |

### Disk Migrations

Since Mars is now responsible for event log persistence, the epoch system disk migrations need to be re-tested with the new architecture:

- [x] `vere-v1.x-aarch64-linux` (thank you @midden-fabler)
- [x] `vere-v2.x-aarch64-linux` (thank you @midden-fabler)
- [x] `vere-v1.x-aarch64-macos`
- [x] `vere-v2.x-aarch64-macos`
- [x] `vere-v1.x-x86_64-linux`
- [x] `vere-v2.x-x86_64-linux`
- [ ] `vere-v1.x-x86_64-macos`
- [ ] `vere-v2.x-x86_64-macos`

(If anyone else is able to test migrations on other platforms, please go ahead and report back in the comments. I will update this list accordingly.)

### To-dos
- [x] Rip out replay in subprocess
- [x] Put disk migration call in mars after replay
- [x] Get rid of all async replay stuff in `pier.c`
- [x] Square `u3_mars_init` with epoch system
- [ ] ~~Fix `kick: lost %mass on /arvo` printed at the end of `|mass` output~~ (this bug isn't caused by this PR)
- [x] Use old code for disk read stuff in Mars/Urth
- [x] Fix `_disk_acquire` callsites
- [x] Fix `_disk_release` callsites
- [x] Handle fatal errors in `_lord_on_plea_boot`
- [x] Move appropriate subprocess arguments we used to send to work over to `u3_lord_boot`
- [x] Handle pace in `mars.c`
- [x] Rift needs to deconstructed in `king.c` and event processed in `u3_mars_boot`
- [x] Ensure double boot protection is still working properly (note: double boot from keyfile protection does not work on `develop`, not just here)


joemfb (changes requested):
I've noted same basic issues above.

The big things still to review are

- boot sequence construction (moved from pier.c to mars.c)
- disk.c changes in light of the epoch system
- |mass in mars vs the serf

matthew-levan:
Closing in favor of #810.

PR #775, opened 2025-03-10
play: add `--watch-replay` flag

### LINKED ISSUE #679: Ship having issues started when passed the `-t` flag

matthew-levan:
Resolves #679. Read this [comment](https://github.com/urbit/vere/issues/679#issuecomment-2706929997) for a summary of the underlying issue.

This PR adds the `--watch-replay` flag to the `play` subcommand. 

When `--watch-replay` is passed to `play`, a watcher thread is spawned to ensure process orphaning does not occur. This option is only helpful when replaying events during normal pier startup, where we spawn a child process, so it's off by default. When we start a pier normally (i.e., without the `play` subcommand), we turn it on in `_cw_play_fork`. 

Keeping `--watch-replay` off by default _prevents_ operators that run `play` with no `STDIN` from experiencing issue #679, like so:
```
matt@mbp14 vere % ./urbit play zod < /dev/null
play: god save the king! committing sudoku...
```

No affordance is granted for disabling the watcher thread during normal pier startup, since we always spawn a child process to replay and want to prevent process orphaning as far as possible.

### DIFF
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -2442,7 +2442,7 @@ _cw_play_fork_exit(uv_process_t* req_u, c3_ds sat_d, c3_i tem_i) {
 static c3_i
 _cw_play_fork(c3_d eve_d, c3_d sap_d, c3_o mel_o, c3_o sof_o, c3_o ful_o)
 {
-  c3_c *argv[12] = {0};
+  c3_c *argv[13] = {0};
   c3_c eve_c[21] = {0};
   c3_c sap_c[21] = {0};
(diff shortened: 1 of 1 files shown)


joemfb (approved):
this will do

PR #784, opened 2025-03-21
fastboot

midden-fabler:
requires: https://github.com/urbit/urbit/pull/7124

issues:
~~|meld is required before running +pill/brass to get cache entries~~
~~userspace persistence must be enabled to get cache entries for kernel modules and userspace~~


joemfb (approved):
this will do

PR #785, opened 2025-03-23
Quodss/urwasm stateful

Quodss:
Depends on https://github.com/urbit/wasm3/pull/1

Adds stateful jet for `++run:wasm-lia`:

- Defines two sets of allocators for m3 structs and codepages;
- Updates `_reduce_monad` to enable computation suspension;
- Adds `_get_state` to load wasm3 state from persistent cache table;
- Adds `_move_state` to store wasm3 state in persistent cache table with the new key and remove an entry with the old key;
- updates `u3we_lia_run` version due to return type change in https://github.com/urbit/urbit/pull/7128;

Misc:
builtins for `c3_d`, motes

TODO:
- [x] optimizations:
    - [x] arena on a slab to prevent copying data one more time;
    - [x] smarter m3 struct realloc (currently wastes memory since wasm3 uses realloc to grow dynamic arrays);


Quodss:
incremented registration tag to reflect change in urbit/urbit side of the PR (new feature)

Quodss:
I changed the target from develop to 409 without rebasing, is it ok? this should be merged with 409

PR #794, opened 2025-04-04
64-bit Vere

xiphiness:
This PR introduces 64-bit Vere. The principle changes involved have to do with the introduction of a new type, c3_n, which is 32 bit or 64 bit depending on if it was compiled with -DVERE64.

- [ ] every appearance of _tmp in pkg/noun (~700 locations)
- [ ] run with llvm addr sanitizer, ubsan
- [ ] audit every jet
- [ ] migrations: bidirectional (use MSB of version tag to mark if vere64 or not? maybe also add metadata for pointer compression)
- [ ] integrate Joe's page based allocator
- [ ] new rsignal_setjmp/longjmp may have volatility problems (talk to Joe. maybe use atomics?)
- [ ] pkg/vere/noun-tests is completely broken
- [ ] every appearance of _tmp in any other subdirectory (~1000 locations)
- [ ] c3_w_clz, ctz, pc occurrences (we should change current occurrences to _tmp as we did with u3r_word, etc and go through them 
- [ ] c3_etch_word, c3_sift_word occurrences (also change to _tmp)
- [ ] make sure everything works (C tests and hoon tests included) still for vere32
- [ ] PMA integration (easy in vere64, vere32 requires more involved work)
- [ ] revendor murmur3 with a size_t len
- [ ] the event log and other lmdb state should not need migration. we should double check events.c, disk.c

I will make this more granular and regularly update this as involved collaborators complete things.


barter-simsum:
majority of changes in hashtable and trace. misc changes elsewhere. majority of remaining _tmp usage to convert to 32/64-bit limited to jets

barter-simsum:
recommend concurrent or after completion of _tmp conversions to run zig build with `-Wconversion` clang opt. There are likely many violations prior to VERE64, so compare the diff between output before/after VERE64 changes and inspect those specific occurences

matthew-levan:
Closing in favor of #715.

PR #800, opened 2025-04-14
term: support multiple sessions

marcusmiguel:
[arvo PR](https://github.com/urbit/urbit/pull/7144)

### To test:

-  Boot a moon with the vere/arvo changes.
-  Current shortcut for switching session is `Ctrl+O`.
-  Try opening and closing new terminal sessions with:
```
|pass [%d %shot %newterm %open %hood [<your ship> %dojo] ~]
|pass [%d %shot %newterm %shut ~]
```
- Try with `%homunculus`:
```
|install ~nidlep-sivrec %homunculus
|pass [%d %shot %homunculus %open %homunculus ~]
|pass [%d %shot %homunculus %shut ~]
```
_Note: the distributed `%homunculus` currently crashes on in the `+init-groups-state` arm of the `%tui-messenger` app, you can prevent this crash by updating the arm as follows:_
```
++  init-groups-state
  |=  bol=bowl:gall
  *(quip card ^groups)
```
### The solution:
The term.c file currently supports only a single default terminal session with features typical of console applications, such as a loading spinner. This pull request introduces a simple approach to enable multiple terminal sessions. It reuses the existing I/O handler for all sessions and adds the u3_Host.act_u variable to track the active session. All input is routed to the active session in Arvo (Urbit’s kernel), while features like the spinner are disabled for applications other than `%drum`/`%hood.`


jfranklin9000:
Coding style change in term.c:
```
void
u3_term_log_exit(void)
{
```
changed to:
```
void u3_term_log_exit(void) {

```

Fang- (changes requested):
(I haven't run this yet, but I did look at the implementation diff.)

> Current shortcut for switching session is Ctrl+O.

I don't think this is the right approach. At least, this isn't how we've intended to approach this. Primarily, I don't think the approach here makes any steps in the direction of "logging reform", and may make that harder to achieve in the future.

"Logging reform" is a catch-all term for improving everything related to vere's (and dill's) output handling. Presently, it dumps most output through stderr, and illegibly intermingles formal dill output, slogs, and runtime outputs. This is both noisy and hard to work with.

The solution we've had in mind for that has a couple of distinct pieces of work, and would start in approximately the same place that this does. Except, instead of saying "support accessing other sessions", it would say "support connecting to session _on a running vere instance, through a separate (vere) process_".

Doing so would give you a process-wise uncoupling of outputs. Sessions won't clobber each other on-switch, and vere doesn't have to take and integrate a stance on session switching/management.

Additionally, this would open up access to further "logging reform" work around controlling/restricting/redirecting the outputs of the main vere process (and eventually even disabling TUI access from that entirely, a la daemon mode), since you're no longer bound to that for your terminal access/ux.

As written, this enables multi-session support, but does so by baking it into the main vere process/flow, which will complicate (or at least doesn't help with) any output-management work we may want to do.

Not to say this isn't cool! It's certainly a much-desired line of work. And I suspect some of the changes here can be preserved/reused for the multi-process approach. But in light of future work in this area I'd be hesitant to accept this as-is.

Happy to provide more detail about the desired approach or anything else here, just let me know!

marcusmiguel:
Thanks for the review! @Fang-
I wasn't aware of the planned "logging reform". I largely agree with your comments. Could you please elaborate on the multi-process approach?

PR #806, opened 2025-04-23
http: only check cache on GET requests

will-hanlen:
fixes a bug where a POST request to /foo/bar would return the cached response for a GET at /foo/bar


joemfb (approved):
Good catch, thanks

PR #814, opened 2025-05-15
+vi registration

Quodss:
Jet registration for `+vi` door https://github.com/urbit/urbit/pull/7162


Quodss:
I didn't fill out u3j_hood[] array for new hooks, are they even necessary?

PR #827, opened 2025-05-22
nock: lose subject in autocons more eagerly

Quodss:
Currently, if Nock bytecode interpreter executes an autocons and does not need the subject to be left on top of the stack (e.g. tail position), it would delay losing it after running both the head and the tail formulas, and only lose them when consing the results. 

This PR reduces the memory usage by allowing the tail formula to consume the subject if it is not needed, losing it sooner.


joemfb (approved):
great catch

PR #831, opened 2025-05-26
nock: lose subject in autocons more eagerly, migration

dozreg-toplud:
Replaces reverted #827, which did not include loom migration logic to free cached bytecode programs, leading to segfaults on boot, unless pack/meld were applied.

I only included `u3j_reclaim` and `u3n_reclaim`, does that make sense? The other caches don't appear to be able to hold Nock bytecode.


Quodss:
also same change could probably be done in mut/kut operators?

PR #836, opened 2025-06-18
Jet fixes

Quodss:
Addresses https://gist.github.com/Quodss/022a71ebcad581109ecfb9a8359d786f.

Adds stricter assertions in arithmetic jets to get rid of jet mismatches from Nock behavior.

Rewrites comparison jets to avoid allocations by mpz.

Removes unused `drop.c` which had a bugged jet.

Adds bloq size assertion in `+swp` jet.

Adds loobean check in `+sort` jet.

Removes erroneus shortcuts from `+cut`/`+met` jets.

Adds atom check in `+cue` jet.


Quodss:
added a bugfix unrelated to jets: in `mesa.c` `_safe_sub` was evaluated twice inside of `c3_min` macro

PR #851, opened 2025-07-29
u3qb_sort use quicksort

Quodss:
Implements `+sort` jet by building an array of nouns to run quicksort algorithm on. Has much better memory usage for large lists and appears to be faster than the previous implementation: `~:(sort (gulf 1 100.000) gth)` used to run out of memory, but not anymore, and `~:(sort (gulf 1 10.000) gth)` runs about 2x faster.


Quodss:
+sort reverses the list if comparison gate always returns yes, and preserves the order if it always returns no. is this intentional?

Quodss:
rewrote the jet to use merge sort for stability. Notice the reversed comparison here: https://github.com/urbit/vere/blob/cccde8c16e780a71e13b14d2c3a54f89c3f8afdc/pkg/noun/jets/b/sort.c#L38-L48

This is done in order to mimick the behavior of +sort function in Hoon, where the first element of the array is taken as a pivot and then it is [passed to the comparator gate as a second argument](https://github.com/urbit/urbit/blob/7e46cf8cefd4e35a0417d99c7ce0bf9fd223f735/pkg/arvo/sys/hoon.hoon#L689), the first being the compared list element.

PR #865, opened 2025-08-27
Prevent stale uncounted references

### LINKED ISSUE #861: Computations with `~+` have different result
Replicated in vere 3.4 release, `develop` and `next/kelvin/409` branches.

Steps to reproduce:
1. Boot a fakezod
2. Commit the desk from https://github.com/Quodss/ska/tree/composed-pronenance (mind the branch)
3. Build and run the tests

```
> __~ -build-file /=ska=/tests/tests/hoon
> -test /=ska=/tests
```

They should run normally.

4. Uncomment line 338 in [`/desk/sur/noir.hoon`](https://github.com/Quodss/ska/blob/94ec2a38f06ffcdccf8d4627033f6b1bdd34def8/desk/sur/noir.hoon#L338)
5. Run tests again. The crash will happen in `+slot`.

Since it is a deterministic crash (i.e. "exit" i.e. bail 1), this breaks Nock.

It might take a few tries for the crash to appear.

Quodss:
Resolves #861

To check: 
- [ ] `parsers.c`
- [ ] `urwasm.c`

### DIFF
--- a/pkg/noun/jets/b/levy.c
+++ b/pkg/noun/jets/b/levy.c
@@ -5,37 +5,39 @@
 
 #include "noun.h"
 
-  static u3_noun
-  _levy_in(u3j_site* sit_u, u3_noun a)
-  {
-    if ( 0 == a ) {
-      return c3y;
-    } else {
-      u3_noun loz;
-
-      if ( c3n == u3du(a) ) {
-        return u3m_bail(c3__exit);
-      }
-      else switch ( (loz = u3j_gate_slam(sit_u, u3k(u3h(a)))) ) {
-        case c3y:  return _levy_in(sit_u, u3t(a));
-        case c3n:   return c3n;
-        default:      u3z(loz);
-                      return u3m_bail(c3__exit);
-      }
-    }
-  }
-
   u3_noun
   u3qb_levy(u3_noun a,
             u3_noun b)
   {
-    u3_noun  pro;
-    u3j_site sit_u;
-    u3j_gate_prep(&sit_u, u3k(b));
-    pro = _levy_in(&sit_u, a);
-    u3j_gate_lose(&sit_u);
+    u3_noun pro = c3y;
+    c3_o loz_o;
+
+    if ( u3_nul != a ) {
+      u3j_site sit_u;
+      u3_noun  i, t;
+      u3j_gate_prep(&sit_u, u3k(b));
+      u3k(a);
+
+      do {
+        i = u3h(a);
+
+        loz_o = u3x_loob(u3j_gate_slam(&sit_u, u3k(i)));
+        if ( c3n == loz_o ) {
+          pro = c3n;
+          break;
+        }
+        
+        t = u3k(u3t(a));
+        u3z(a), a = t;
+      } while ( u3_nul != a );
+
+      u3z(a);
+      u3j_gate_lose(&sit_u);
+    }
+
     return pro;
   }
+
   u3_noun
   u3wb_levy(u3_noun cor)
   {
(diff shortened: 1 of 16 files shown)


pkg/noun/manage.c
```diff
@@ -1867,8 +1867,6 @@ _cm_in_pretty(u3_noun som, c3_o sel_o, c3_c* str_c)
       c3_w len_w = u3r_met(3, som);
 
       if ( _(_cm_is_tas(som, len_w)) ) {
-        c3_w len_w = u3r_met(3, som);
-
```
Quodss, line 1871: bonus commit: pretty printer measured the length of the atom multiple times for some reason

pkg/noun/nock.c
```diff
@@ -3245,9 +3245,7 @@ u3n_slam_et(u3_noun gul, u3_noun gat, u3_noun sam)
 u3_noun
 u3n_nock_an(u3_noun bus, u3_noun fol)
 {
-  u3_noun gul = u3nt(u3nc(1, 0), u3nc(0, 0), 0);  //  |~(^ ~)  XX 409: just pass ~
-
-  return u3n_nock_et(gul, bus, fol);
+  return u3n_nock_et(u3_nul, bus, fol);
```
Quodss, line 3248: `u3m_soft_run` now can take `0` instead of a scry gate to mean the same thing as `_~`, right?

pkg/noun/jets/b/levy.c
```diff
@@ -5,37 +5,39 @@
...
+        if ( c3n == loz_o ) {
+          pro = c3n;
+          break;
+        }
+        
+        t = u3k(u3t(a));
+        u3z(a), a = t;
+      } while ( u3_nul != a );
+
+      u3z(a);
```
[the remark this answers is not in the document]
Quodss: technically the very first gain of the list is not necessary and we could unroll the loop, doing things as usual in the first iteration and starting to count the references after that, but IMO whatever gain we get from that is not worth the bloated code

pkg/noun/jets/d/by_jab.c
```diff
@@ -21,15 +21,21 @@ u3qdb_jab(u3_noun a,
...
-             : u3nt(u3k(n_a), u3k(l_a), u3qdb_jab(r_a, key, fun));
+      u3k(n_a);
+      if ( c3y == u3qc_gor(key, pn_a) ) {
+        u3k(r_a);
+        return u3nt(n_a, u3qdb_jab(l_a, key, fun), r_a);
+      }
+      else {
+        u3k(l_a);
+        return u3nt(n_a, l_a, u3qdb_jab(r_a, key, fun));
+      }
```
joemfb, line 38: I don't think these changes are sufficient safety. `a` is uncounted reference in these recursive calls. The equivalent of the transfer loops in the list jets is for there to be a recursive implementation of jab to transfer `a`.

joemfb (changes requested):
NB: I haven't closely reviewed the parser jet changes yet.

pkg/noun/nock.c
joemfb: It doesn't mean exactly the same thing: this will bail exit, while the prior behavior would block. But this function appears to be unused, so it's not clear that it matters.

joemfb: Good catch. The mismatch comes from urbit/urbit#7124 and the hoon is wrong -- `+mink` should crash if there is no scry handler.

pkg/noun/jets/d/by_jab.c
Quodss: you're right, I'll change the helper function to use transfer semantics

joemfb (approved):
tyfys

PR #875, opened 2025-09-19
vere: prep road layout for release

joemfb:
Removes redundant files from the road struct and adds support for lightweight bytecode migrations (see #831).

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -547,13 +547,17 @@ _pave_home(void)
   memset(u3H, 0, sizeof(u3v_home));
   u3H->ver_d = U3V_VERLAT;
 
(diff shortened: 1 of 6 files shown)


pkg/noun/manage.c
```diff
@@ -600,6 +606,13 @@ _find_home(void)
     abort();
   }
 
+  if ( ((pam_d >> 6) & 31) != U3N_VERLAT ) {
+    fprintf(stderr, "loom: discarding stale bytecode programs\r\n");
```
Quodss, line 610: should we also update Nock version in the home road here? to prevent freeing bytecode cache on every boot

joemfb: yes, good call.

PR #879, opened 2025-09-23
Forward --lmdb-map-size through _cw_play_fork

### LINKED ISSUE #878: lmdb-map-size flag is not being preserved by cw_play_fork
_cw_play_fork() manually constructs a new argv[] for the child process, but does not forward the "--lmdb-map-size" option. As a result, the child cw_play never sees the flag. 

For devices that cannot handle the default map size of 500GiB, this means that although they'll manage to, for example, mine a comet with "./urbit -c --lmdb-map-size 1073741824", once they exit the comet and try to rerun it with "./urbit mycomet --lmdb-map-size 1073741824", the process will fail with "lmdb: failed to open event log: Out of memory".

With a bunch of [strategically added print statements for debugging](https://github.com/urbit/vere/commit/d81896908cd656284da738ac48c8eea2007bc927), this is what the output looks like on a device with memory problems when you try to run a comet:

```
./urbit mycomet --lmdb-map-size 1073741824

urbit 3.5-35f7662d3d

boot: home is /home/gcec21/urbit/mycomet

[LMDB DEBUG] u3_lmdb_init: requested mapsize = 536870912000 bytes (512000.00 MiB)

[LMDB DEBUG] mdb_env_set_mapsize succeeded

[LMDB DEBUG] calling mdb_env_open(path='/home/gcec21/urbit/mycomet/.urb/log', flags=0x400000)

[LMDB DEBUG] mdb_env_open failed: code=12 (Out of memory)

lmdb: failed to open event log: Out of memory 

disk: failed to initialize lmdb

unable to open event log

play: failed: 1 signal: 0
```

In particular, note that although the lmdb-map-size flag was used, u3_lmdb_init still used the default value. Although _cw_play_fork receives the right map value, every function after it does not, leading to lmdb being handed the default map size. The output is identical when using mycomet/.run. On a device that supports the default memory size, everthing will run, but still ignoring the lmdb-map-size flag.

Gcec21:
Resolves #878
This PR fixes an issue where --lmdb-map-size was ignored when running via cw_play. The option never reached the child process because _cw_play_fork() rebuilt argv[] manually and omitted the flag. As a result, replay sessions always used the default LMDB map size regardless of the user’s input.

The following changes were made to _cw_play_fork() so that lmdb-map-size could be passed down:
- Added map_c to hold the decimal string of u3_Host.ops_u.siz_i / lmdb_map_bytes.
- Inserted --lmdb-map-size and its value into the argv[] array in _cw_play_fork().

The following change was made because it seems that once _cw_play_fork passes --lmdb-map-size forward, if cw_play doesn't know how to handle the additional option, it causes an error:
- Adds --lmdb-map-size to _cw_play

### DIFF
--- a/pkg/vere/main.c
+++ b/pkg/vere/main.c
@@ -2639,6 +2639,7 @@ _cw_play(c3_i argc, c3_c* argv[])
     { "auto-meld",         no_argument,       NULL, 7 },
     { "soft-mugs",         no_argument,       NULL, 8 },
     { "watch-replay",      no_argument,       NULL, 9 },
+    { "lmdb-map-size",     required_argument, NULL, 10 },
     { "full",              no_argument,       NULL, 'f' },
     { "replay-to",         required_argument, NULL, 'n' },
     { "snap-at",           required_argument, NULL, 's' },
@@ -2671,6 +2672,13 @@ _cw_play(c3_i argc, c3_c* argv[])
       case 9: {  //  watch-replay
         wat_o = c3y;
       } break;
+
+      case 10: {  //  lmdb-map-size
+        if ( 1 != sscanf(optarg, "%" SCNuMAX, &u3_Host.ops_u.siz_i) ) {
+          exit(1);
+        }
+        break;
+      }
 
       case 'f': {
         ful_o = c3y;


Gcec21:
You're welcome, and no worries! I'm just glad I was able to help :)

PR #892, opened 2025-10-10
conn: robustly deliver response after channel read EOF

mikolajpp:
It is presently impossible to interface with conn using the typical pattern of sending a command using one program, and receiving the result with another through Unix pipes. Conn will eagerly close the connection as soon as it hears EOF from the read end of the pipe, which does not leave any chance to deliver an future result to the recipient. The only way to reliably receive results from conn in this situation is to set long timeouts in tools such as `netcat` or `socat`.

This PR is (1) request for clarification about the protocol, whether the original behavior was intended, and (2) a first attempt to rectify this situation by:
1. Only closing the read end of the pipe on genuine errors in `newt.c`
2. Closing the channel after the result has been delivered, if the read end of the channel's pipe had reached EOF.

This PR requires an expert's look whether this change fits well with other IO drivers, given that `newt.c` is used in many other places.

With this PR applied, I have been able to successfully use `socat` with the `click` script to dramatically speedup interactions with an Urbit ship via conn.

Our backend tests runner now completes in under `20s`, while previously it took more than 4m to complete, due to a need for long timeouts at each interaction.

### DIFF
--- a/pkg/vere/io/conn.c
+++ b/pkg/vere/io/conn.c
@@ -699,7 +699,6 @@ _conn_sock_cb(uv_stream_t* sem_u, c3_i tas_i)
   u3_chan*  can_u;
   c3_i      err_i;
 
-
   can_u = c3_calloc(sizeof(u3_chan));
   can_u->mor_u.ptr_v = can_u;
   can_u->mor_u.pok_f = _conn_moor_poke;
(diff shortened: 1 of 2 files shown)


joemfb:
This change seems right for conn, but not the behavior we want for mars/urth IPC. In the long run, I don't think both of those things should use newt, but for now it'd be a bit of work to detangle them. I've pushed up a hacky workaround to preserve the old IPC behavior (see bcb965a4930f3d09d7a), can you take a look and make sure it's still suitable for your needs?

mikolajpp:
I have tested with your changes, works just as good.

PR #897, opened 2025-10-27
mesa: make _get_now_micros return micros

yosoyubik:


pkova:
I broke this for some inexplicable reason in 42be1850f0999e2472f882f555dbd78c7231dd17. In practice this meant that on `vere-v4.0` the pit entries were cleared every five and a half hours instead of every 20 seconds.

PR #899, opened 2025-10-27
mesa: don't leak lanes

dozreg-toplud:
Resolves #


joemfb:
Right, thanks. I was focused on `dat` and didn't notice that the old behavior was effectively `las = u3k(u3t(dat));`.

PR #905, opened 2025-10-30
%loop hint for nondeterministic crashes on cycles

dozreg-toplud:
Partially implements `%loop` hint proposed in [UIP-0123](https://github.com/urbit/UIPs/blob/main/UIPS/UIP-0123.md).

The crash is nondeterministic: it cannot be caught with Nock virtualizing interpreters e.g. `+mink`, and it does not require changes to the `+mink` definition.

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -665,6 +665,13 @@ _find_home(void)
     u3j_reclaim();
     u3H->pam_d = _pave_params();
   }
+
+  //  if lop_p is zero than it is an old pier pre %loop hint, initialize the
+  //  HAMT
+  //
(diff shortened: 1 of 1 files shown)


pkg/noun/allocate.h
```diff
@@ -144,8 +144,9 @@ STATIC_ASSERT( u3a_vits <= u3a_min_log,
 
         c3_w off_w;                           //  spin stack offset
         c3_w fow_w;                           //  spin stack overflow count
+        u3p(u3h_root) lop_p;                  //  %loop hint set
```
joemfb: This has to be initialized on home road for existing piers. A real migration, or just a zero-check in find-home.

PR #906, opened 2025-10-31
u3: fix alignment in u3a_pile_prep

yosoyubik:


pkg/noun/allocate.c
```diff
@@ -252,11 +252,13 @@ u3a_wealloc(void* lag_v, c3_w len_w)
...
 void
-u3a_pile_prep(u3a_pile* pil_u, c3_w len_w)
+u3a_pile_prep(u3a_pile* pil_u, c3_w len_w, c3_w lin_w)
 {
+  c3_w pad_w = -(uintptr_t)len_w & (lin_w - 1);
+
   //  frame size, in words
   //
-  c3_w wor_w = (len_w + 3) >> 2;
+  c3_w wor_w = (len_w + 3 + pad_w) >> 2;
```
joemfb: I don't think we want or need to change the frame size to accomplish alignment. The frame size (in words) is the `sizeof` (in bytes), so it already accounts for any padding necessary to align an array of these frames. What's missing is the initial alignment of the base pointer (ie, cap_p). That needs to happen, and the result stored in top_p. The pile also needs an additional member to track and restore (in `u3a_pile_done()` the original, un-aligned cap.

pkova:
Closing in favor of #1072.

PR #963, opened 2026-02-18
spin: port to windows

dozreg-toplud:
Ports spin stack stuff to Windows. Not tested yet.

### DIFF
--- a/pkg/c3/platform/windows/compat.c
+++ b/pkg/c3/platform/windows/compat.c
@@ -615,33 +615,36 @@ ssize_t pwrite(int fd, const void *buf, size_t count, off_t offset)
 static BOOL
 _shm_make_path(char *out, size_t outsz, const char *name)
 {
-    char tmp_dir[MAX_PATH];
-    DWORD len_tmp_dir = GetTempPathA((DWORD)sizeof(tmp_dir), tmp_dir);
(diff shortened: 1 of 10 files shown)


dozreg-toplud:
First approach was to create an anonymous mapping with `CreateFileMappingA(INVALID_HANDLE_VALUE, ...)`, but it didn't work because the size [had to be set explicitly at the time of mapping](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga#:~:text=An%20attempt%20to%20map%20a%20file%20with%20a%20length%20of%200%20(zero)%20fails%20with%20an%20error%20code%20of%20ERROR_FILE_INVALID.%20Applications%20should%20test%20for%20files%20with%20a%20length%20of%200%20(zero)%20and%20reject%20those%20files.).

I tried making a sized version of `shm_open` but `_open_osfhandle` would return `-1` for some reason. So I ended up making a file-backed mapping with a file in `Temp` directory

Cross-compiled to Windows and ran natively, spin stack output appears, so all seems to work

dozreg-toplud:
I did not realize that `u3t_boot`/`u3t_boff` are not related to spin hint, rolling these back for now

dozreg-toplud:
depends on #967, will open another

PR #967, opened 2026-02-19
make u3m_stop reachable

dozreg-toplud:
Previously `u3m_stop` was not called on a shutdown because the processes would exit before `uv_run` would return. This would prevent some cleanup procedures to run.

@mopfel-winrux is spin stack only initialized in urth?

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -2597,19 +2597,13 @@ u3m_init(size_t len_i)
 
 extern void u3je_secp_stop(void);
 
-/* u3m_stop_*(): graceful shutdown cleanup of urth/mars.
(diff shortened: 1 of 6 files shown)


mopfel-winrux:
No it should be initialized in mars also. I think mars actually initializes it first.

dozreg-toplud:
@mopfel-winrux thanks, I misread what was going on. The global variable is indeed initialized in mars only.

dozreg-toplud:
I initially wanted to have to separate `u3m_stop_*` routines for mars and urth, since I wanted to implement shared memory mapping between process as a file-backed mapping on windows, and only mars would have to delete the file on cleanup. Since then I refactored spin hint port on windows (see #969) which no longer creates that extra file

PR #970, opened 2026-02-23
vere64

matthew-levan:
Cribbed from #794.

This PR introduces 64-bit Vere. `c3_w` is now either 32 or 64 bit, depending on if it was compiled with `-DVERE64` (off by default, or on with `zig build -Dvere64`).

- [x] define final typedefs, update usages accordingly
- [x] every appearance of _tmp in pkg/noun (excluding pkg/noun/jets) (~800 locations)
  - [x] `trace.{c,h}`
  - [x] `ship.{c,h}`
  - [x] `nock.h`
  - [x] `version.h`
  - [x] `imprison.{c,h}`
  - [x] `jets.{c,h}`
  - [x] `allocate.{c,h}`
  - [x] `retrieve.{c,h}`
  - [x] `*_tests.c`
- [x] every appearance of _tmp in pkg/noun/jets
- [x] ~~look at diff of pre and post VERE64 compilation with -Wconversion flag. Investigate every instance.~~
- [x] run with llvm addr sanitizer, ubsan
  - [x] found [issues](https://gist.github.com/matthew-levan/7b7597f71e6fa920912a6b6c4d438531) with asan and ubsan as of 61fc9dee67295692b1881996d07a67cf971c8c7a (lmdb addressed in #820)
- [x] audit every jet
- [x] migrations: bidirectional (use MSB of version tag to mark if vere64 or not? maybe also add metadata for pointer compression)
- [x] integrate Joe's page based allocator
- [x] new rsignal_setjmp/longjmp may have volatility problems (talk to Joe. maybe use atomics?)
- [x] pkg/vere/noun-tests is completely broken
  - [x] ensure noun tests make sense/are worth having
- [x] every appearance of _tmp in any other subdirectory (~1000 locations)
- [x] c3_w_clz, ctz, pc occurrences (we should change current occurrences to _tmp as we did with u3r_word, etc and go through them 
- [x] c3_etch_word, c3_sift_word occurrences (also change to _tmp)
- [x] make sure everything works (C tests and hoon tests included) still for vere32
- [x] revendor murmur3 with a size_t len
  - [x] urbit/urbit#7186
  - [x] #842 
- [x] the event log and other lmdb state should not need migration. we should double check lmdb.c, disk.c
- [x] there's a lot of bloq size related things in the jets that could be changed to greater than 32 for the 32 bit case
- [x] `(mug .(eny 0, now 0))` equality test for fakezods booted in 32-bit and 64-bit modes
- [x] mug-verified automated migration tests in ci

Benchmarks:

- M4 Pro macOS Tahoe 26.3.1(a)

|                   | ackermann   | dec        | add       | fibonacci | bigslot  |
|-------------------|-------------|------------|-----------|-----------|----------|
| vere64 `a959ae`   | s/2.007.327 | ms/188.825 | ms/92.988 | µs/730    | ms/1.259 |
| develop `4e75c9e` | s/2.045.726 | ms/192.338 | ms/92.393 | µs/809    | ms/1.457 |

- Framework 13 Ryzen AI 9 HX 370

|                   | ackermann   | dec        | add       | fibonacci | bigslot  |
|-------------------|-------------|------------|-----------|-----------|----------|
| vere64 `a959ae`   | s/2.577.564 | ms/279.785 | ms/127.991 | µs/785   | ms/1.356 |
| develop `4e75c9e` | s/2.490.242 | ms/262.862 | ms/126.190 | µs/762   | ms/1.386 |


pkg/noun/urth.c
```diff
@@ -17,6 +17,7 @@
 #include "serial.h"
 #include "ur/ur.h"
 #include "vortex.h"
+// XX: 64 more to do
```
mopfel-winrux: can we remove this?

pkg/noun/retrieve.c
dozreg-toplud: When testing on Windows, I was getting strange results because `mpz_init_set_ui` takes `unsigned long int`, which is smaller than `c3_w` on Windows.

So you want to replace this with something more general, like:

```c
void
mpz_init_set_word(mpz_t a_mp, c3_w b_w)
{
  if ( sizeof(unsigned long int) >= sizeof(b_w) ) {
    return mpz_init_set_ui(a_mp, b_w);
  }
  mpz_init2(a_mp, 1 << u3a_word_bits_log);
  mpz_import(a_mp, 1, -1, sizeof(c3_w), 0, 0, &b_w);
}
```

dozreg-toplud: Or maybe don't define a wrapper function for words and just compare to `ULONG_MAX`. Either way you should check interactions with external API like that.

pkg/c3/platform/windows/compat.c
```diff
@@ -197,6 +197,8 @@ void* mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off)
                     (DWORD)off : (DWORD)(off & 0xFFFFFFFFL);
     const DWORD dwFileOffsetHigh = (sizeof(off_t) <= sizeof(DWORD)) ?
                     (DWORD)0 : (DWORD)((off >> 32) & 0xFFFFFFFFL);
+    const DWORD dwMaxSizeLow = (DWORD)(len & 0xFFFFFFFFL);
```
dozreg-toplud: probably should use the same `(sizeof(len) <= sizeof(DWORD)) ? ...` ternary as off_t?

pkg/c3/defs.h
```diff
@@ -48,18 +48,23 @@
 
     /* Size in words.
     */
+// (probably some of that belongs here)
+#ifndef VERE64
 #     define c3_wiseof(x)  (((sizeof (x)) + 3) >> 2)
+#else
+#     define c3_wiseof(x)  (((sizeof (x)) + 7) >> 3)
```
dozreg-toplud: probably could be just `(sizeof(x) + sizeof(c3_w) - 1) / sizeof(c3_w)`; all of this should be folded at compile time anyway so using division over shifts shouldn't matter?

pkg/noun/jets/b/find.c
```diff
@@ -5,8 +5,8 @@
 
 #include "noun.h"
 
-STATIC_ASSERT( (UINT32_MAX > u3a_cells),
-               "list index precision" );
+//STATIC_ASSERT( (UINT32_MAX > u3a_cells),
```
dozreg-toplud: you probably don't want to comment out static asserts like that, here and in lent.c it would be sufficient to replace UINT32_MAX with c3_w_max

pkg/noun/jets/c/bex.c
dozreg-toplud: does this still hold?

pkg/noun/jets/e/rs.c
```diff
@@ -24,13 +24,13 @@
...
     {
-      *(c3_w*)(&a) = SINGNAN;
+      *(c3_h*)(&a) = SINGNAN;
     }
     return a;
   }
 
   static inline void
-  _set_rounding(c3_w a)
+  _set_rounding(c3_y a)
```
dozreg-toplud: here and elsewhere: this must be a u3_atom or a type of the same width. Otherwise an indirect atom will get incorrectly interpreted

dozreg-toplud (changes requested):

pkg/noun/jets/i/lagoon.c
```diff
@@ -39,7 +39,7 @@
 
   //  $?(%n %u %d %z %a)
   static inline void
-  _set_rounding(c3_w a)
+  _set_rounding(c3_y a)
```
dozreg-toplud: you missed the rounding stuff here

PR #982, opened 2026-03-23
Revert "Async signal/bail safety"

pkova:
Reverts urbit/vere#940


dozreg-toplud:
#940 was reverted because the cold state accumulated in a child road cannot be simply discarded as the traps accumulated in the stacktrace depend on it. In practice it means having to walk the stacktrace, removing the traps, removing crash information. Not good!

A better approach would be to actually make our hashtables async-signal (i.e. they should be traversable at all times).

PR #983, opened 2026-03-23
windows: clobber nonvolatile regs

dozreg-toplud:
Per [docs](https://llvm.org/docs/ExceptionHandling.html#llvm-eh-sjlj-setjmp):
> ... this intrinsic forces register saving for the current function...

Maybe we should still call the builtin just to signal to the compiler that the registers ought to be saved?

### DIFF
--- a/pkg/c3/portable.h
+++ b/pkg/c3/portable.h
@@ -269,4 +269,9 @@
 #     define STATIC_ASSERT(e,m) \
         enum { ASSERT_CONCAT(assert_line_, __LINE__) = 1/(int)(!!(e)) }
 
+
+#ifndef u3_prep_setjmp
+#   define u3_prep_setjmp()
+#endif
+
 #endif /* ifndef C3_PORTABLE_H */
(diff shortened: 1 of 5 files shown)


dozreg-toplud:
Notably the resumption is in `windows_setjmp` now instead of its caller. Does it matter?

dozreg-toplud:
In testing both your and my branches kill the process on `!!` in Dojo...

dozreg-toplud:
Both branches were killing the process with optimizations enabled, on debug both worked. That led me to believe it's the register allocation.

I tried calling the intrinsic with the goal of clobbering all registers, which did not work for some reason. Then I added manual clobbering of all callee-saved registers (caller-saved registers are already clobbered by calling `windows_setjmp`, which I force to not be inlined), and it did work on `ReleaseFast`

dozreg-toplud:
Added register clobbering after longjmp as well as before setjmp. Tested on vere64, seems to work

dozreg-toplud:
This is rendered obsolete by #987

PR #986, opened 2026-03-30
mesa: use loobean check in conditional

yosoyubik:


pkova:
Caused some annoying `mesa: send fail_async` prints for galaxies that don't have their dns records set.

PR #991, opened 2026-04-01
Trim stracktrace on `bail:meme`

dozreg-toplud:
@midden-fabler noticed that the runtime process would occasionaly get aborted if a `bail:meme` involved large stacktrace - it would be too big to copy out in `_cm_stack_unwind`, causing us to `bail:meme` on the home road again.

This PR adds stack trace trimming on the deepmost road if we crash with `meme` moat and the stacktrace happens to be bigger than 1024 frames. It stitches up the beginning and the ending parts with `[%mean 'meme: trace']` which would typically get cut off by `+mook` anyway. Such operations on the stacktrace are safe for the nondeterministic crashes, as documented in the comment.

Note: for this to work `end` must be a subnoun of `beg`, which should be true considering the comparisons I used.

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -285,26 +285,39 @@ _cm_stack_unwind(c3_l sig_l)
   u3_noun tax;
 
   if ( c3__meme == sig_l ) {
-    u3_noun xat = u3R->bug.tax;
+    //  heuristic: if we crash with %meme then our stacktrace might be too big
+    //  to copy, so we trim it.
+    //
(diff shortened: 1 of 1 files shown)


dozreg-toplud:
And to reproduce the issue, run this in Dojo:

```
!:  |-($)
```

dozreg-toplud:
closing in favor of #993; this particular fix wouldn't work without mug clearing / `use_w` assertions since the trace is injected into Arvo later

PR #992, opened 2026-04-01
Merge next/kelvin/408 to develop

pkova:


dozreg-toplud:
FYI this silently unreverses #940

PR #993, opened 2026-04-01
manage: fix emergency buffer and north road calculation

pkova:
Resolves #


pkg/noun/manage.c
```diff
@@ -399,7 +399,7 @@ _cm_signal_deep(void)
   //
   if ( 0 == u3H->rod_u.bug.mer ) {
     u3H->rod_u.bug.mer = u3i_string(
```
dozreg-toplud: you forgot to change this to tape

pkg/noun/allocate.h
```diff
@@ -285,7 +285,7 @@ STATIC_ASSERT( u3a_vits <= u3a_min_log,
 
     /* u3a_is_north(): yes if road [r] is north road.
     */
-#     define  u3a_is_north(r)  __((r)->cap_p > (r)->hat_p)
+#     define  u3a_is_north(r)  __((r)->mat_p > (r)->rut_p)
```
joemfb, line 289: A road is north if the heap is low and the stack is high; south is opposite. This check has always been implemented using the innermost bounds of the stack ("cap") and heap ("hat").

The heap growing over the stack is guarded by explicit checks in the allocator, but the stack growing over the heap is not (the "cap" is constantly moving, so such checks have a measurable cost). Instead, a `PROT_NONE` guard page floats between the heap and the stack. Stores into it are caught by a global `SIGSEGV` handler and the page is repositioned if possible. If not the fault is handled with `u3m_signal(c3__meme)`. In that state, the old implementation of `u3a_is_north()` will give the wrong answer, which wreaks havoc on refcounting and de/allocation logic.

This change makes `u3a_is_north()` robust against overflow states, by comparing the immutable outermost bounds of the heap ("rut") and stack ("mat").

pkg/noun/manage.c
```diff
@@ -1209,8 +1209,8 @@ u3m_hate(c3_w pad_w)
   u3R->ear_p = u3R->cap_p;
   u3m_leap(pad_w);
 
-  u3R->bug.mer = u3i_string(
-    "emergency buffer with sufficient space to cons the trace and bail"
+  u3R->bug.mer = u3i_tape(
+    "emergency buffer"
```
joemfb: This is overkill, since at-most three cells are required for re-consing the bail mote and trace in `u3m_bail()`. At the same time, since this is a fresh road, we'll be allocating a full page of cells and putting the remainder in the cell pool. So I'm not sure if the difference matters.

joemfb (changes requested):
`u3_melt_all()` / `u3_meld_all()` should be changed to lose and reallocate the emergency buffer (or simply not canonicalize it). Its purpose is defeated if it ends up structurally shared with any other part of the state.

joemfb (approved):

PR #997, opened 2026-04-08
http: support multi-cert %set-config gifts

Fang-:
Eyre is going to start passing `%set-config` gifts that contain a list of per-domain SSL certificates in [the foreseeable future](https://github.com/urbit/UIPs/pull/85). Here, we update `http.c` to handle these and pick the right cert based on the incoming request.

The way we do this is by registering a "bootstrap" TLS context on the HTTPS port's server, and giving that an SNI (server name indication) callback. At that point in the connection we get told the target hostname by the client, and can go see if there's a matching cert for that hostname. If there is, we swap out the TLS context for one with the matching cert. Otherwise, we continue serving the request with the bootstrap/fallback context.

Eyre is expected to provide the domain-cert list in priority order, where at the very least "most generic" certs (commonly, wildcard certs) are at the end of the list. The last cert in the list will be used for bootstrap and fallback.

Draft because there is no corresponding urbit/urbit PR for the eyre changes yet. The relevant lull/eyre changes haven't even been pushed to the working branch! Will link those here as soon as they are, but the changes to `http.c` here should be reviewable enough standalone, so wanted to get this up to kick that process off.

### DIFF
--- a/pkg/vere/io/http.c
+++ b/pkg/vere/io/http.c
@@ -82,14 +82,21 @@ typedef struct _u3_h2o_serv {
     struct _u3_httd* htd_u;             //  device backpointer
   } u3_http;
 
+  typedef struct _u3_cert {
+    u3_noun          tuf;               //  domain (as turf, %$ == *)
+    uv_buf_t         key_u;             //  PEM RSA private key
+    uv_buf_t         cer_u;             //  PEM certificate chain
+    SSL_CTX*         ctx_u;             //  domain-specific SSL context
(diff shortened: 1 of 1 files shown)


Fang-:
With the realization that a single cert may be used for multiple distinct domains, and that a naive approach has arvo send a distinct map entry for each domain, this will result in a new `uv_buf_t` for every domain even if they may all use the same cert. Probably not desirable, some (yet undesigned) work to be done here to resolve that.

PR #1006, opened 2026-04-29
meld: make dropping memo/ford persistent tables optional

dozreg-toplud:
Adds arguments to `u3_meld_all` to control whether we free the persistent memoization buckets.

Adds `memo-drop` and `ford-drop` CLI arguments to free the buckets (they are retained by default).

Added argument parsing to `%meld` task: if an argument is a pair of loobeans then they are passed, otherwise fall back to not freeing the buckets.

NB: On automeld we free the persistent memo bucket but not the ford caches. Thoughts?

### DIFF
--- a/pkg/vere/io/conn.c
+++ b/pkg/vere/io/conn.c
@@ -669,7 +669,7 @@ _conn_moor_poke(void* ptr_v, c3_d len_d, c3_y* byt_y)
         } break;
         case c3__meld: {
           _conn_send_noun(can_u, u3nc(u3k(rid), c3y));
-          u3_pier_meld(con_u->car_u.pir_u);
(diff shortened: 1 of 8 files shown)


matthew-levan:
For blob storage, it'd be useful for automeld to drop all the caches so
that chop actually deletes blobs that have zero references left (assuming
the user runs |rm, |tomb on the blob first). Without dropping the ford
cache, a dangling reference from clay will remain, and blobs won't ever be
deleted.

dozreg-toplud:
@matthew-levan ford caches would still be freed in all other occasions, like memory pressure. `auto-meld` is an opt-in binary flag. If we free ford caches on automeld we'd have to rebuild the marks after each automeld.

dozreg-toplud:
also adds `[%urth %meld *]` case handling for conn

PR #1007, opened 2026-05-04
http: don't copy bytes from octs

dozreg-toplud:
Prevents an extra copy of bytes in urth process by taking an ownership of the atom from which the bytes are taken. Still does a copy if the atom is too small so it needs to be extended, or if len_w is precisely equal to met(3, atom), so the null terminator would have to be added.


dozreg-toplud:
TODO test to see if performance gains are meaningful

PR #1011, opened 2026-05-13
ford: drop caches more granularly

dozreg-toplud:
Adds a global variable that tracks the road depth at which the ford cache was freed.

In top-level wrapper set it to zero (might be set to non-zero value by ^C-ing during kernel upgrade).

On `%drop` set it to current road depth and free current Ford table.

On leaving the road check if the caches are old by comparing the variable against current road depth, freeing the caches if they are old. Note the comment on the operation order.

### DIFF
--- a/pkg/noun/nock.c
+++ b/pkg/noun/nock.c
@@ -1922,7 +1922,7 @@ _n_hilt_fore(u3_noun hin, u3_noun bus, u3_noun* out)
     } break;
 
     case c3__drop: {
-      u3m_Ford_fresh_road_depth_w = u3m_road_depth();
+      u3m_Ford_fresh_road_depth_w = c3_max(u3m_road_depth(), u3m_Ford_fresh_road_depth_w);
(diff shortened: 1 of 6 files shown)


joemfb (approved):
discussed in the blitz, LGTM.

pkg/noun/nock.c
```diff
@@ -1922,7 +1922,9 @@ _n_hilt_fore(u3_noun hin, u3_noun bus, u3_noun* out)
     } break;
 
     case c3__drop: {
-      u3C.wag_w |= u3o_free_ford;
+      u3m_Ford_fresh_road_depth_w = c3_max(u3m_road_depth(), u3m_Ford_fresh_road_depth_w);
```
dozreg-toplud, line 1925: @joemfb I don't think taking maximum is necessary actually: `u3m_Ford_fresh_road_depth_w` is always less then or equal to the current road depth, so taking the maximum of it and the current road depth is equivalent to just taking the current road depth.

PR #1015, opened 2026-05-17
nock: earlier shortcircuit for nonequality of a direct atom with a noun

dozreg-toplud:
When testing SKA transpiler output efficiency I noticed that some `u3r_sing` overhead could be removed by shortcurcuiting earlier if at least one of the arguments is a direct atom. This made hoon.hoon compilation 3-5% faster.

### DIFF
--- a/pkg/noun/allocate.h
+++ b/pkg/noun/allocate.h
@@ -643,7 +643,10 @@ u3a_post_info(u3_post);
         */
           u3_weak
           u3a_gain(u3_weak som);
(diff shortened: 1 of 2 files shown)


dozreg-toplud:
Having to enter `u3a_gain/lose` to check if a noun is direct was causing nontrivial (>10%) overhead in SKA bytecode interpreter, adding a guard makes hoon.hoon compilation a bit faster even in the current bytecode interpeter.

PR #487, opened 2023-07-03
compile with flto

barter-simsum:
Resolves #486

I haven't spoken with @fighet-parnet in a minute so I'm forking his original
change and including a basic change to install.md requested. Currently, testing
to make sure gdb isn't borked as previously experienced and that there are no
performance regressions.

Will close !469 if we merge this. I've reviewed the bazel changes there and
everything LGTM.

PR #1017, opened 2026-05-18
build: enable lto everywhere

dozreg-toplud:
Seems like this gives ~7% performance boost. Takes forever to link though.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -111,6 +111,7 @@ const BuildCfg = struct {
     tracy_callstack: bool = false,
     tracy_no_exit: bool = false,
     gen_cdb: bool = false,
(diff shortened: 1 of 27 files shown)


dozreg-toplud:
Added `-Dno-lto` flag to supress LTO in case if you need to build optimized build quickly. Also disabled LTO for macos targets, it doesn't work on them for some reason.

joemfb:
Previously: #487; for macos, thinlto was used in #469.

dozreg-toplud:
I am getting `error: error: using LLD to link macho files is unsupported` when trying to build with LLD on, and it wouldn't build without LLD either

PR #1021, opened 2026-05-31
noun: migrate lagoon jet to stateless SoftBLAS

sigilante:
## Summary

Migrates the lagoon jet to the current SoftBLAS API.

- Bump `ext/softblas` to [`40ff4fa`](https://github.com/urbit/SoftBLAS) — the allocation-free / stateless rewrite that dropped the global `softblas_roundingMode` for a per-call `rndMode` argument (plus B1–B5 correctness fixes, NaN canonicalization). Drop the now-deleted `src/softblas_state.c` from the wrapper's source list.
- Update `pkg/noun/jets/i/lagoon.c` for the new API: thread the active rounding mode through all 32 BLAS calls via a file-static `_la_rnd` (replacing the removed global), and rename the local `_set_rounding` → `_set_rounding_la` so it doesn't collide with `softblas.h`'s new inline `_set_rounding`.

The jet source tracks numerics canonical (`lagoon/vere`); see urbit/numerics#37.

## Testing
Built with zig 0.15.2; `add`/`dot`/`mmul`/`mod` verified correct on a fake `~hex`, and the full lagoon test suite passes (run against the matching 64-bit `ml/64` build).

### DIFF
--- a/ext/softblas/build.zig
+++ b/ext/softblas/build.zig
@@ -24,7 +24,6 @@ pub fn build(b: *std.Build) void {
     lib.addCSourceFiles(.{
         .root = dep_c.path(""),
         .files = &.{
-            "src/softblas_state.c",
             "src/blas/level1/sasum.c",
             "src/blas/level1/dasum.c",
             "src/blas/level1/hasum.c",
--- a/ext/softblas/build.zig.zon
+++ b/ext/softblas/build.zig.zon
@@ -7,8 +7,8 @@
             .path = "../softfloat",
         },
         .softblas = .{
-            .url = "https://github.com/urbit/SoftBLAS/archive/cbffb33f19ea02f9ffbd184d445123c57929ec53.tar.gz",
-            .hash = "N-V-__8AAJD5MgBhfBHYae8jFlcaQw9R-TRw4tcUFB3rO9-q",
+            .url = "https://github.com/urbit/SoftBLAS/archive/40ff4fa0a3fba0a566ac5fcb8ab93d280fddf7af.tar.gz",
+            .hash = "N-V-__8AAObNNABARc0F3nVusDf1nQ4yxZH78h5A6HkWWAA9",
         },
     },
     .paths = .{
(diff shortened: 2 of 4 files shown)


sigilante:
Coordination note: #1057 rewrites this same file's validation layer and several kernels on develop (refcounts, `_check`, NaN comparisons, mod/range/trace/transpose results). Whichever of these lands first, the other needs a port across it — the fix set and the stateless-rounding plumbing touch the same lines throughout. A reference for combining them exists: the numerics repo's `lagoon/vere64/noun/jets/lagoon.c` (urbit/numerics#75) is exactly this PR's stateless lineage with the #1057 fixes applied, ship-verified — with ONE deliberate difference to keep out of the upstream version: its %mod implements the numerics desk's C-fmod semantics, whereas upstream %mod must keep the +toi door-mode/crash semantics as in #1057.

sigilante:
Rebased onto #1057 (base retargeted to `sigilante/lagoon-check-fixes`), now a single stateless-SoftBLAS commit. Bumps the vendored SoftBLAS to the stateless build (rounding passed per call, no softblas_roundingMode global; src/softblas_state.c dropped) and threads the door mode through a file-static `_la_rnd` set by the renamed `_set_rounding_la` — renamed because the stateless SoftBLAS header now defines its own `_set_rounding`, while keeping #1057's punt-on-unknown-mode c3_o return. `_la_rnd` is the trailing arg of every ?axpy/?gemm/?dot/?scal call. Also fixes one @rq jet in jets/e/rq.c still reading a 128-bit quad as 4 machine words on the 64-bit loom.

Post-vere64 the single source tree builds both looms, so this collapses with #1022 (the old 64-bit split) into one change — #1022 can be closed once this lands.

Verification: builds clean at both loom widths (-Werror); refcount linter (#1059) reports 0 findings on lagoon.c (strict-weak). The full numerics lagoon suite (13 files) runs green jetted on a fresh fakezod for both binaries (206 OK arms 32-bit, 219 64-bit, 0 failures) — bit-identical to the stateful path, as expected for a plumbing change. Merge #1057 first.

PR #1022, opened 2026-05-31
noun: migrate lagoon jet to stateless SoftBLAS (64-bit)

sigilante:
## Summary

Migrates the 64-bit lagoon jet to the current SoftBLAS API, on top of `ml/64`.

- Bump `ext/softblas` to [`40ff4fa`](https://github.com/urbit/SoftBLAS) — allocation-free / stateless rewrite (per-call `rndMode` replacing the global `softblas_roundingMode`; B1–B5 correctness fixes). Drop the deleted `src/softblas_state.c` from the wrapper.
- Update `pkg/noun/jets/i/lagoon.c` for the new API: thread the active rounding mode through all 32 BLAS calls via a file-static `_la_rnd`, and rename the local `_set_rounding` → `_set_rounding_la` (avoids colliding with `softblas.h`'s new inline). **Preserves the 64-bit calling convention** — the braced `u3r_mean {sample, &ptr}` pairs and `syz_x+1` byte handling.

The jet also picks up the recent numerics correctness fixes the 64-bit copy was missing (mod round-the-quotient, `_get_dims` leak, range/bloq guards, argmin/argmax index). Source tracks numerics `lagoon/vere64`; see urbit/numerics#37.

## Testing
Built on `ml/64` with zig 0.15.2. Booted a fake `~tex`, verified `add`/`dot`/`mmul`/`mod` produce correct values, and the **full lagoon test suite passes** — all 9 suites, `ok=%.y`.


matthew-levan:
LGTM, will wait for #1021 before merging this one.

PR #1026, opened 2026-06-03
jets: simpler definition of constants

dozreg-toplud:
@matthew-levan, when trying to deduplicate now_w definition we might've forgotten to update `now_z` as well.


matthew-levan:
Looks like the compiler doesn't like what you're doing here. It says you can suppress the warning by wrapping the `sizeof(c3_w)` call in parentheses.

matthew-levan:
Merged. TYFYS.

PR #1027, opened 2026-06-04
jets: fix buffer overrun

dozreg-toplud:
Wasn't that a buffer overrun? `buf_w[8]` is outside of the atom buffer.

### DIFF
--- a/pkg/noun/jets/e/secp.c
+++ b/pkg/noun/jets/e/secp.c
@@ -63,9 +63,9 @@ _cqes_in_order(u3_atom a)
       c3_y i_y;
       c3_w *buf_w = a_u->buf_w;
       // loop from most to least significant words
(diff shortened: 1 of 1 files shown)


matthew-levan (approved):

joemfb (approved):
We discussed this in ccrs. This fix is correct, the old code was wrong, but almost always (`1 / (2^32 - 1)`) effectively correct anyway (since random garbage will likely not have all bits set, and a refcount will certainly not have all bits set).

dozreg-toplud:
I think this reasoning applies for well-formed input, i.e. input for which this function should return one. I guess if a jet was called in such a way that this function got an eight word-long atom which was greater than `now_w` buffer we would get a false positive and a jet mismatch: the function would compute some result instead of crashing. 

If in practice that function was called with only well-formed inputs then there's nothing to worry about.

PR #1034, opened 2026-06-15
retrieve: `u3r_met` returns `c3_d`

matthew-levan:
promotes `u3r_met` to return `c3_d` so it no longer truncates the bit/byte length of big atoms.

sweeps every callsite: comparisons are left as-is, values feeding 64-bit consumers (`u3i_slab_init`, `u3r_chop_words` widths) are widened to `c3_d`, and values that need 32-bit width get a recoverable `bail:fail` guard before an explicit `(c3_w)` narrow, preserving the behavior u3r_met previously raised on overflow. `rap`/`rep` keep their 32-bit `tot_w` overflow guards. `met`/`xeb` now return via `u3i_chub` so counts past `2^31` produce a correct atom instead of a truncated one.

### DIFF
--- a/ext/natpmp/build.zig.zon
+++ b/ext/natpmp/build.zig.zon
@@ -4,8 +4,7 @@
     .version = "0.0.1",
     .dependencies = .{
         .natpmp = .{
-            .url = "https://mirrors.wikimedia.org/ubuntu/ubuntu/pool/main/libn/libnatpmp/libnatpmp_20230423.orig.tar.gz",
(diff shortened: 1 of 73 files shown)


joemfb:
The assertion pattern used here replaces the recoverable `bail: fail` exceptions that would have been generated in `u3r_met()` with unrecoverable `bail: oops` exceptions that will take down the process.

matthew-levan:
@joemfb thank you, fixed.

PR #1037, opened 2026-06-16
u3: adds bump allocation on inner roads

joemfb:
This PR implements a bump allocator (`u3a_flag_sand`) and enables it by default on all inner roads (this can be disabled before road entry with `u3o_no_sand`), targeting #1036. Included microbenchmarks show rougly 2x performance increase in cons and refcounting relative to the status quo.

Bump-allocating is not viable for some workloads -- compiling hoon.hoon used roughly 6GB on my machine. Enabling it by default would require explicit opt-out in several scenarios, it's likely safer to make it explicitly opt-in. Either scenario will require "user-space" control via a hint or somesuch.


joemfb:
This PR will need to be updated to reflect #1015 once it's merged.

PR #1039, opened 2026-06-18
trace: key spin hint with ship name in addition to process id

pkova:
Tlon hosting hosts several ships in multiple kubernetes containers inside a single pod. I had the misfortune of learning that /dev/shm is shared between the containers and this fact in conjunction with their deterministic process ids causes these ships to thrash eachothers spin pages with explosive consequences. I have personally verified this using a local kubernetes cluster and the posix shared memory tests at https://github.com/abcdesktopio/podshmtest.

Here I key the shared memory by ship name in addition to the process id, if anybody is running the same ship inside multiple containers at the same time they're gonna have bigger problems than the binary crashing.

The thing I don't understand is how vere-v4.3 is relatively stable even though this bug already exists. It must be the case that the limited offsets caused by the bug I fixed in a38528f were somehow guarding us from segfaults...

### DIFF
--- a/pkg/noun/trace.c
+++ b/pkg/noun/trace.c
@@ -1131,10 +1131,17 @@ u3t_etch_meme(c3_l mod_l)
 /* u3t_sstack_init: initalize a root node on the spin stack 
 */
 void
-u3t_sstack_init()
+u3t_sstack_init(c3_d* who_d)
 {
   c3_c shm_name[256];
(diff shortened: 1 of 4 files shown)


pkova:
@joemfb are there any lifecycle concerns with `who_d` in either urth or mars here?

joemfb (approved):
No lifecycle problems, `who_d` is available in these places.

PR #1043, opened 2026-06-24
trace: uses short `SLOW_STACK_NAME` on macOS

matthew-levan:


pkova:
Macos only allows extremely short file names in shm_open, `PSHMNAMELENGTH` is 31. The intent of my previous fix #1039 was to fix the behavior on x86-64 linux in kubernetes, here we revert to the old behavior on macos.

PR #1044, opened 2026-06-26
noun: math.hoon transcendental jets (@rd/@rs/@rh/@rq)

sigilante:
Registers the `math.hoon` transcendental library jets (from the numerics
project's `/lib/math`) in the userspace `%non` chapter, so the four IEEE
precision doors run native SoftFloat instead of interpreted Hoon.

## What
- **15 transcendentals × 4 doors** — `exp log sin cos tan atan atan2 asin acos
  sqt cbt pow pow-n log-2 log-10` for `@rd` (f64), `@rs` (f32), `@rh` (f16,
  native), `@rq` (f128) — registered as `non → math → <door> → arm` in the
  **135, 136 and 137** trees.
- Jet bodies in `jets/i/math.c` (chub-based I/O, word-size-agnostic), the
  `u3qi_*`/`u3wi_*` decls in `q.h`/`w.h`, and the `build.zig` source entry.

## Verification
On a fresh hoon-135 fakezod: all four doors fire (`exp:rd/rs/rh:math`
~1 µs/call, `exp:rq` ~1.8 µs, vs ~250 µs interpreted) and `-test %/tests/lib`
is **247/247 green**, bit-exact to the Hoon door (same reduction, coefficients,
Horner order). Jet ≡ interpreted Hoon confirmed arm-by-arm.

### DIFF
--- a/pkg/noun/build.zig
+++ b/pkg/noun/build.zig
@@ -389,6 +389,7 @@ const c_source_files = [_][]const u8{
     "jets/f/ut_rest.c",
     "jets/g/plot.c",
     "jets/i/lagoon.c",
+    "jets/i/math.c",
     "jets/tree.c",
     "jets/137/tree.c",
     "jets/136/tree.c",
--- a/pkg/noun/build.zig
+++ b/pkg/noun/build.zig
@@ -391,6 +391,7 @@ const c_source_files = [_][]const u8{
     "jets/f/ut_rest.c",
     "jets/g/plot.c",
     "jets/i/lagoon.c",
+    "jets/i/math.c",
     "jets/tree.c",
     "jets/137/tree.c",
     "jets/136/tree.c",
--- a/pkg/noun/jets/135/tree.c
+++ b/pkg/noun/jets/135/tree.c
@@ -944,8 +944,156 @@ static u3j_core _135_non__la_core_d[] =
     {}
   };
 
+//  numerics math.hoon transcendentals: non -> math -> rd -> <fn>
+static u3j_harm _135_non__math_rd_exp_a[] = {{".2", u3wi_rd_exp}, {}};
+static u3j_harm _135_non__math_rd_log_a[] = {{".2", u3wi_rd_log}, {}};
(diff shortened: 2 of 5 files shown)


sigilante:
CI failure: near is a reserved identifier/macro on the CI toolchain (it compiles on macOS clang but not Linux gcc — near is a legacy reserved word). That hits the @rh inverse-trig code in math.c and affects both vere PRs.

sigilante:
**Dependency note:** this PR is based directly on `develop` and is **independent of #1021** (the 32-bit lagoon→SoftBLAS migration) — no shared files, no dependency, so the two can merge in either order. Its 64-bit counterpart #1045 *is* stacked on #1022.

sigilante:
Rebased onto post-vere64 develop (single-source tree now builds both loom widths). Since the whole math.c is new on develop, only the shared jet-registration files needed merging; `math.c` is unchanged in substance from the prior head plus two rebase adaptations, folded into the last commit:

- The four single-argument `@r` jet helpers type their `u3r_at(u3x_sam, cor)` sample `u3_weak` before the `u3_none`/`u3ud` guard (refcount linter, #1059).
- The five `u3r_mean` call sites move from the old flat, 0-terminated form to develop's `{axe, &out}` braced-pair form (`u3r_vmean`), which `-Werror`/`-Wmissing-braces` now requires.

Verification: builds clean at both widths (`zig build`, `zig build -Dvere64`, `-Werror`); refcount linter reports 0 findings on `math.c` (strict-weak). The numerics libmath math test suite (12 files: exp/log/sqrt/tan/trig/atan/ainv/rh/rq/derived/rounding/constants) runs green jetted on a fresh fakezod for both binaries (272 OK arms 32-bit, 411 64-bit, 0 failures).

Note: numerics#74 (the `_math_rnd` file-static leaving `u3qi_*` composites call-history-dependent) is a latent API-hygiene issue — no in-tree caller reaches a composite without its `u3wi_*` wrapper setting the mode first — and the clean fix is a rounding-mode-parameter refactor across ~12 entry points. Left as a separate follow-up rather than bundled into this rebase.

This makes #1045 (the separate 64-bit math port stacked on #1022) redundant; it can be closed once this lands.

PR #1045, opened 2026-06-26
noun: math.hoon transcendental jets (@rd/@rs/@rh/@rq) [ml64]

sigilante:
64-bit (ml64) counterpart of #1044. Registers the math.hoon transcendental
library (numerics /lib/math) in the userspace %non chapter so the four IEEE
precision doors run native SoftFloat instead of interpreted Hoon.

Stacked on `neal/lagoon-softblas-64`; merge after the 64-bit lagoon work.

## What
- 15 transcendentals × 4 doors (`exp log sin cos tan atan atan2 asin acos sqt
  cbt pow pow-n log-2 log-10`) for `@rd`/`@rs`/`@rh`/`@rq` as
  `non → math → <door> → arm` in the **135** tree.
- Jet bodies in `jets/i/math.c` (chub I/O; byte-identical to the 32-bit copy
  bar the two two-arg `u3r_mean` call sites), `q.h`/`w.h` decls, build.zig entry.

## Verification
On a fresh hoon-136 fakezod (with 136 scaffolding applied locally): all four
doors fire (`exp:rd/rs/rh` ~1 µs/call, `exp:rq` ~1.8 µs, vs ~250 µs
interpreted) and `-test %/tests/lib` is **247/247 green**, bit-exact to the
Hoon door and identical to the 32-bit runtime.


sigilante:
CI failure: near is a reserved identifier/macro on the CI toolchain (it compiles on macOS clang but not Linux gcc — near is a legacy reserved word). That hits the @rh inverse-trig code in math.c and affects both vere PRs.

sigilante:
**Dependency note:** this PR is **stacked on #1022** (`neal/lagoon-softblas-64`, the 64-bit lagoon→SoftBLAS migration) — its base branch already contains #1022's commits, so **#1022 should merge first**, and this PR will retarget to `ml/64` once #1022 lands. (Its 32-bit counterpart #1044 is independent of #1021.)

PR #1047, opened 2026-06-29
noun: /lib/twoc two's-complement jets (scalar + Lagoon %int2 arrays)

sigilante:
Bit-exact C jets for the numerics `/lib/twoc` two's-complement integer library, under the `non` chapter alongside `math`/`lagoon`/`unum`. Hoon side: urbit/numerics#66.

**No vendored library** — the arithmetic is plain integer C plus **GNU MP** (already linked into `pkg/noun`).

## Changes
- `jets/i/twoc.c` — one `%twid` core (the width-keyed door where `/lib/twoc` keeps its logic; the bloq-keyed facade re-exports its gates, so this fires for both call paths). Reads `wid` from the door sample (gate axis 30) and dispatches: native `c3_d` for `wid<=64`, `__int128` for `<=128`, GMP (`u3r_mp`/`u3i_mp`) for `>128`. Arms: `add sub neg mul div rem pow abs gth lth lte gte`. Declines to Hoon on out-of-width operands and zero divisors.
- `jets/i/twoc.h` — shared two's-complement kernels (used by `twoc.c` and `lagoon.c`).
- `jets/i/lagoon.c` — the element-wise array jets (`add/sub/mul/div/mod-rays`) gain a `%int2` case beside `%i754`, looping native two's-complement per lane over the ray byte buffer (one C pass vs the super-linear Hoon bit-packing loop).
- `jets/w.h`, `q.h` — `u3wi_twid_*` / `u3qi_twid_*` declarations.
- `build.zig` — add `jets/i/twoc.c` (no new `linkLibrary`; GMP already linked).
- `jets/{135,136,137}/tree.c` — register `twid` under `non` (sibling of `lagoon`).

## Validation
- Standalone vs an independent reference: 8280 native + 5692 GMP checks pass.
- Fires bit-exact on a hoon-135 fakezod. Scalar proven two ways (a planted sentinel confirmed the C path executes via both `twid`-direct and the `twoc` facade; timing toggle ~3.3×). `%int2` array ops verified jet == Hoon for all five; ~310,000× over interpreted at 200k lanes.

Built and fired on the 32-bit runtime.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1048, opened 2026-06-29
noun: Lagoon %int2 comparison + reduction array jets

sigilante:
Stacked on #1047. Hoon/master side: urbit/numerics#67.

Extends the Lagoon array jets with `%int2` cases beside `%i754` for the comparison and reduction ops (no GEMM):

- `gth/gte/lth/lte` — per-lane signed (two's-complement) compare → 1/0 ray.
- `cumsum` — full wrapping sum → scalar ray.
- `dot` — sum of products (wrapping) → scalar ray.
- `min`/`max` — extremum by two's-complement order → scalar ray.
- `argmin`/`argmax` — forward ravel-index of the first extremum (bare index).

Native per lane (8…128-bit, no GMP); signed compares reinterpret the lane as the matching signed C type. Scalar reductions box via a `_la_scalar_box` helper = `+scalar-to-ray`'s shape (all-1s of the input rank).

**Also fixes the i754 reduction-jet result shape** (carries urbit/numerics#68 into the runtime): `cumsum`/`min`/`max`/`dot` shared the same `_la_scalar_box` helper, replacing the prior hardcoded `~[1 1]`/`~[len 1]`/`~[1]` that only matched one rank — a silent jet/Hoon mismatch on the boxed meta. `trace` (always rank-2) and `diag` (returns a vector) are correct as-is.

Verified jet == numerics Hoon on a hoon-135 fakezod across native widths incl. 128-bit, ties, and negatives — for both `%int2` and (now) `%i754` reductions.

### DIFF
--- a/pkg/noun/jets/i/lagoon.c
+++ b/pkg/noun/jets/i/lagoon.c
@@ -2377,7 +2377,7 @@
 ** carries its leading-0x1 sentinel (== +spac on a 1-element ray).
 */
   static u3_noun
-  _la_int2_box(u3_noun x_shape, u3_noun x_bloq, u3_noun x_kind,
+  _la_scalar_box(u3_noun x_shape, u3_noun x_bloq, u3_noun x_kind,
                u3_noun x_tail, u3_noun r_data)
   {
     c3_d rank = 0;  u3_noun s = x_shape;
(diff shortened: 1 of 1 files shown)


sigilante:
Rebased onto #1047 (base retargeted to `sigilante/twoc-jets`), so this PR is now just the %int2 comparison/reduction commit. The %int2 cases (gth/gte/lth/lte, cumsum/dot/min/max/argmin/argmax) slot into #1057's rewritten comparison and reduction wrappers; the +dot case was added by hand to #1057's restructured (unifying-equality-safe) dot wrapper, which the auto-merge could not place. All `_la_int2_*` helpers and int2 wrapper results are typed u3_weak; `_la_int2_box` (which transfers r_data into the boxed ray) carries `@Refcount: transfers `r_data`` so the reduction wrappers do not read as leaking it (refcount linter, #1059).

#1048's separate i754 reduction-shape fix is dropped — #1057 already boxes those with the all-ones-of-rank +scalar-to-ray shape (`_ones_shape`), which the %int2 box matches.

Verification: builds clean at both loom widths (-Werror); linter reports 0 findings on lagoon.c (strict-weak). lagoon-int2 (5 arms: modular wrap, signed div/rem, abs, comparison convention, cumsum) and twoc (13 arms) run green jetted for both binaries. Merge order: #1057, then #1047, then this.

PR #1049, opened 2026-06-29
noun: fix Lagoon argmin/argmax jets to return the forward (ravel) index

sigilante:
The Lagoon `argmin`/`argmax` jets compute the result index as `len_x - i - 1` (**reversed**), so the jetted index silently disagrees with the Hoon — which returns the forward ravel index of the *first* extremum — and with the unjetted interpreter, whenever the extremum isn't at index 0.

This was already fixed in the numerics master copy (urbit/numerics#29, `sigilante/fix-argmin-argmax-index`) but never carried into the vere runtime, so `develop` still ships the reversed index (8 occurrences across argmin + argmax).

Fix: use the forward index `i`.

**Verified** on a hoon-135 fakezod — for a 3-vector with its min/max at logical index 2, the jet now returns `2` (was `0`), matching the unjetted Hoon:
```
[jet=2 hoon=2 jet-max=2 hoon-max=2]
```

### DIFF
--- a/pkg/noun/jets/i/lagoon.c
+++ b/pkg/noun/jets/i/lagoon.c
@@ -589,7 +589,7 @@
         for (c3_d i = 0; i < len_x; i++) {
            if(f16_lt(((float16_t*)x_bytes)[i], min_val16)) {
              min_val16 = ((float16_t*)x_bytes)[i];
-             min_idx = (len_x - i - 1);
+             min_idx = i;
            }
         }
         break;}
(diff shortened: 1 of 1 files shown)


sigilante:
Superseded by #1057, which contains this exact index fix plus the surrounding work the same audit turned up in these kernels (NaN-correct comparison scans — SoftBLAS's `f_gt` macros are true against NaN — and freeing the scan buffers). Closing in favor of #1057.

PR #1009, opened 2026-05-05
jets: manual noun splitting instead of u3r_mean

dozreg-toplud:
@joemfb I tried replacing u3r_mean calls with explicit core splitting. Some observations:

1. Compiler did eliminate redundant deconstructions;
2. `+ride` call on Hoon source seems ~5% _slower_ with explicit splitting.

The way I implemented it was:

1. Define macros that take `u3_noun` or `u3_weak` and return `u3_weak`:

```
#define u3r_head_weak(SOM) ({                   \
  u3_weak _som = SOM;                           \
  ( u3_none == _som ) ? u3_none                 \
    : (_(u3ud(_som))) ? u3_none                 \
    : ( ((u3a_cell *)u3a_to_ptr(_som))->hed );  \
})

#define u3r_tail_weak(SOM) ({                   \
  u3_weak _som = SOM;                           \
  ( u3_none == _som ) ? u3_none                 \
    : (_(u3ud(_som))) ? u3_none                 \
    : ( ((u3a_cell *)u3a_to_ptr(_som))->tel );  \
})

#define u3r_head(SOM) ({                      \
  u3_noun _som = SOM;                         \
  ( _(u3ud(_som)) ) ? u3_none                 \
  : ( ((u3a_cell *)u3a_to_ptr(_som))->hed );  \
})

#define u3r_tail(SOM) ({                      \
  u3_noun _som = SOM;                         \
  ( _(u3ud(_som)) ) ? u3_none                 \
  : ( ((u3a_cell *)u3a_to_ptr(_som))->tel );  \
})
```

2. Apply a [generated python script](https://gist.github.com/dozreg-toplud/f392be7cac99ca8435717b4817fb196a) to all files in `/jets` directory, turning `u3r_mean` into composition of the function-like macros defined above

Am I missing some way to make it more efficient?

PR #1057, opened 2026-07-07
noun: fix validation, refcounts, and results in lagoon jets

sigilante:
Prompted by ~dozreg-toplud's refcount linter, which flagged leaked nouns in the lagoon jets' `_check`, plus a hand audit of the surrounding code against the Hoon library the jets serve. The audit turned up considerably more than the linter findings: broken validation, jet/Nock semantic divergences, wrong-result kernel bugs, and a family of leaks. Everything lands in `pkg/noun/jets/i/lagoon.c`.

## `_check` helper

The old `_check` was wrong in four independent ways:

- `u3qa_dec`'s result was never freed (the linter's finding), and `u3r_met`'s raw `c3_w` return was type-punned through `u3_atom`.
- The shape product was computed in `c3_d`, truncated into a `u3_atom`, and compared word-for-word against a noun — so any well-typed ray with ≥ 2³¹ elements spuriously failed `_check` (jet crashes where the Nock succeeds), and a crafted shape whose product wraps could *pass* it.
- The product could silently overflow 64 bits.
- The bloq went into `u3r_met` unvalidated; `u3r_met` requires `a_y < 37` and only enforces it with `c3_dessert`, which release builds compile out — undefined behavior on a hostile bloq.

The rewrite takes `meta`/`data` as two borrowed nouns (no cell is consed, nothing is allocated), computes the product with overflow detection (an overflowed product cannot equal the block count of any real atom, which mirrors the Hoon), handles bloqs ≥ 37 the way `+met` does, and bails `%exit` only where `+check` itself crashes (cell dims, improper shape list, zero data underflowing `+dec`).

Every call site previously passed `_check(u3nc(x_meta, x_data))`. `u3nc` transfers ownership, but the metas/datas are borrowed from the sample, and the cell was never freed — one leaked cell per jet invocation, with the ownership violation masked *by* the leak (naively freeing the cell would have corrupted the sample's refcounts).

## Wrapper ↔ Hoon divergences

- **transpose/diag passed the whole core to `_check`** (`_check(cor)`), which then read the gate's battery as ray metadata — these two jets unconditionally bailed `%exit` whenever they fired. This had gone unnoticed because there are no transpose tests in the desks that exercise the jets.
- The elementwise (`+add`/`+sub`/`+mul`/`+div`/`+mod`), comparison (`+gth`/`+gte`/`+lth`/`+lte`), scalar (`+add-scalar` &c), `+abs`, and `+trace` wrappers **never called `_check` at all**, though the Hoon asserts it — the jet computed a result where the Nock deterministically crashes.
- `+ravel` has no `+check` and never reads the kind, so its wrapper now **punts** (not bails) on an inconsistent ray.
- `+dot` only asserts equal *shapes* outside the `%i754` path; the full meta-equality and consistency assertions moved inside the `%i754` case so other kinds punt to the Nock instead of bailing.
- `+mmul` never compared the two rays' bloq/kind (the Hoon computes elementwise on each ray's own meta; the kernel assumes one width), and its gemm switch had no default — an out-of-range bloq **silently returned a zero matrix**. Both cases now punt.
- Reduction results (`+cumsum`/`+min`/`+max`/`+dot`) hard-coded shapes `~[1]`, `~[1 1]`, or `~[n 1]`; `+scalar-to-ray` actually returns an all-ones shape of the input's rank. The dot wrapper's `~[n 1]` produced a shape/data-inconsistent ray outright.
- `_set_rounding` now reports an unrecognized rounding mode so callers punt, instead of `u3m_bail(c3__fail)`.

## Kernel bugs (wrong results)

All previously unobservable because the wrapper bailed before reaching them, or untested:

- **trace** computed `(dot d d)` — the sum of *squares* of the diagonal — instead of `(cumsum (diag a))`. Identity-matrix tests mask this (1² = 1).
- **transpose** read and wrote with rows/columns swapped (correct only for square matrices), and the wrapper returned the *unswapped* shape.
- **argmin/argmax** returned `len - i - 1` instead of the ravel index `i`.
- **diag** and **abs** return elements in natural order, in lockstep with urbit/urbit#7388, which removes erroneous flops in `+diag` and `+el-wise-op` (the "compensate for LSB" comment there is mistaken — `+rep` puts the first list element lowest, so the flop *reverses* the elements). Skew note: the old abs jet already returned natural order, so jetted and unjetted ships disagree about `+abs` today — the Hoon fix converges with deployed jet behavior; the old diag jet never returned at all. The two PRs should land together.
- **trace** honors the door rounding mode (its sum previously used whatever `softfloat_roundingMode` was left over); **linspace** fences an indirect count atom before using it as a raw integer in allocation sizes.
- **mod** (ray-ray and scalar) rounded the quotient with a hardcoded round-toward-zero, but `+toi` rounds in the door mode — `7 mod 2` is `-1` under `%n`, not `1`. It also *returned values* where the Hoon crashes: a non-finite quotient (zero or NaN divisor) is `(need ~)`. And **mod-scalar/div-scalar** multiplied by a rounded `1/n` instead of dividing — wrong even for exact quotients (`21 mod 7` gave `7`; `21/7` gave `2.9999998`). All four kernels now divide directly, round via `softfloat_roundingMode`, and bail `%exit` on a non-finite quotient.
- **gth/gte** used SoftBLAS's `f_gt`/`f_ge` macros, which are `(!le)`/`(!lt)` — **true** whenever either operand is NaN, where IEEE gt/ge are false. The same disease reaches **min/max/argmax** through `f_min`/`f_max`. All now use `lt`/`le` with swapped operands, which also reproduces the Hoon reel-fold exactly (NaN at the head is sticky, interior NaN is skipped, first-of-ties wins).
- **range** counted elements with a one-shot `ceil((b-a)/d)`, but the Hoon iterates `x+d` in the door mode testing each sum against `b` — for `d = .0.1` the two *shapes* disagreed outright (10 vs 11 elements). The kernel now replicates the iteration (the values are the accumulated sums), the wrapper takes the count from the result's block count, and it punts on non-finite bounds/step, a zero step, or a stalled accumulator (`x+d == x` short of `b`), where the Nock loops forever.

Also fixed leaks of `_get_dims` arrays in the diag/dot wrappers and trace kernel (with raw `c3_d` dimensions used directly as nouns — invalid for dims ≥ 2³¹), the argmin/argmax scan buffers, and over-kept result shape cells in linspace/range.

## Verification

Fresh fake ship (`-F bus`, `urbit-408k` pill) on this branch, against the base `lib/lagoon.hoon` with the urbit/urbit#7388 flop fixes applied (its `%add-rays` &c hints match the tree registration), differentially compared with a de-jetted copy of the same library (parent `~%  %non` hint renamed so nothing registers):

- 36/36 elementwise-equality checks pass, including: transpose/mmul of non-square matrices, diag/trace, reduction shapes at rank 1, argmin/argmax with extrema away from index 0 and with NaN at and after the head, mod under `%n`/`%u`, mod-scalar and div-scalar on exact quotients, gth/gte/lth against NaN, abs on negative entries, ravel, linspace, range with `d = .0.1` and with a negative step, and add/cumsum/mod-scalar/range under non-default rounding modes via `+lake`.
- Zero-divisor `mod` crashes on both doors (jet bails `%exit` where the Nock's `(need ~)` crashes).
- Jets confirmed firing, not punting: 40×40 `%i754` mmul+trace runs in ≤3 s jetted vs ~25 s interpreted, with bit-identical results.

Notes for reviewers: `u3r_met` returns `c3_w`, so `+check` on atoms of ≥ 2³² blocks is still constrained upstream of this file; and the SoftBLAS gemm reads `B` with `ldc` instead of `ldb` (harmless here since the jet passes `ldb == ldc`, but worth fixing in SoftBLAS). SoftBLAS's `f_gt`/`f_ge`/`f_min`/`f_max` macros are NaN-incorrect for the same `!le` reason and deserve an upstream fix there too; this PR routes around them locally.


pkg/noun/jets/i/lagoon.c
dozreg-toplud, line 101: An indirect atom in the shape would blow this up.

dozreg-toplud, line 2251: Same

pkg/noun/jets/i/lagoon.c
```diff
@@ -3011,32 +3201,44 @@
...
          )
       {
         return u3m_bail(c3__exit);
       } else {
         switch (x_kind) {
           case c3__i754:
-            _set_rounding(rnd);
+            //  ... and the %i754 path, (cumsum (mul a b)), asserts
+            //  equal metas and consistency in +bin-op.
+            if ( c3n == u3r_sing(x_meta, y_meta) ||
```
dozreg-toplud, line 3230: This comparison invalidates uncounted references into subnouns of the compared nouns, in particular x_shape, due to UE.

dozreg-toplud (changes requested):

pkg/noun/jets/i/lagoon.c
dozreg-toplud, line 2225: what if Na and Nb are equal but indirect so not pointer-equivalent? Na and Nb need to be checked for directness too

dozreg-toplud (changes requested):

pkg/noun/jets/i/lagoon.c
dozreg-toplud: Also, what if this overflows?

sigilante:
## Rebase note (2026-09-11)

Rebased onto develop after the vere64 merge (#970): one source tree now builds both loom widths (`-Dvere64`), so this PR targets both. develop's lagoon.c had meanwhile replaced `u3r_mean` with manual noun splitting (#1009), renamed the half-word types (`c3_h`, `u3i_half`), and changed `_set_rounding` to take a `u3_atom`; the fix set is re-applied on top of all of that.

The branch is two commits:

1. **`noun: fix validation, refcounts, and results in lagoon jets`** -- the original fix set, rebased (unchanged in substance; see the description below).
2. **`noun: lagoon refcount-linter fixes, u3_weak retyping, %fail on non-terminating +range`** -- the review findings from ~dozreg-toplud's refcount linter (#1059) and follow-up notes:
   - `_get_length` and the `mmul` dimension reads used `u3x_atom` products as C integers, so an indirect atom in a shape would blow up; they now go through `u3r_cat` (which also makes the `Na != Nb` check compare values, not noun words).
   - `u3wi_la_dot` kept borrowed references into `x_meta` across `u3r_sing(x_meta, y_meta)`. Unifying equality rewrites equal subnouns in place, so the borrowed `x_shape` could be freed before its use in the result. The wrapper now reads every field it returns only after the last comparison that could have rewritten it.
   - `u3qi_la_*` / `u3wi_la_*` are typed `u3_weak` (definitions and q.h/w.h), since they return `u3_none` to punt.
   - `+range`: where the Hoon can never terminate -- a NaN upper bound, or a stalled accumulator (zero/underflowing step, NaN, infinite start) -- the jet now bails `%fail` instead of punting. An infinite loop is a non-deterministic failure, not a deterministic `%exit`, and punting it to the Nock hangs the ship. The absurd-count cap is still a punt.
   - The dot and trace wrappers use manual noun splitting like the rest of the file on develop.

Verification on this rebase:
- refcount linter (#1059 at 3fa686ff8b, libclang 21) on the lagoon translation unit: 0 findings, also under `--strict-weak`. For contrast, on develop's current lagoon.c it reports the `_check` leak, seven `u3nc(x_meta, ...)` ownership errors, and four indirect-atom-as-integer sites -- the bugs this PR fixes.
- Builds clean at both widths (`zig build`, `zig build -Dvere64`) with `-Werror`.
- urbit/urbit#7388's `tests/lib/lagoon` regression suite (22 arms) run on a fresh fakezod from the 408k pill with each binary: all green on both the 32-bit and 64-bit binaries (fresh fakezod, 408k pill); the urbit/urbit#7388 parity suite (22 arms) also passes on both widths. A sub-jet operand-order bug (see the extra commit) was caught by lagoon-arithmetic test-sub-asym.

Issue #626, opened 2024-03-28
Ship unrecoverably crashed after hitting loom limit via HTTP file upload & eyre cache insertion


tinnus-napbus:
I have a files app that takes a file via an HTTP POST request, saves it in state and inserts it in the Eyre cache for quick access via the browser. I repeatedly uploaded large files until it presumably hit the loom limit and crashed with bail:meme. Upon attempting ship restart it just crashed again, and will no longer start. Crashing is perhaps understandable but it presumably shouldn't permanently brick the ship.

initial crash:

```
bail: meme
home: bailing out
pier: EOF
~zod:dojo> pier: work exit: status 0, signal 6
```

on reboot:

```
❯ ./piers/zod/.run
~
urbit 3.0
boot: home is /home/mike/piers/zod
disk: loaded epoch 0i109
loom: mapped 2048MB
boot: protected loom
live: mapped: MB/760.184.832
live: loaded: KB/16.384
boot: installed 967 jets
---------------- playback starting ----------------
play: events 152470-152509
allocate: reclaim: half of 3 entries
allocate: reclaim: half of 1 entries
allocate: reclaim: memo cache: empty

bail: meme
home: bailing out
zsh: IOT instruction (core dumped)  ./piers/zod/.run
```

this was after uploading a 100mb file repeatedly

joemfb:
Can you meld or run with a larger loom?

PR #1061, opened 2026-07-14
noun: drop lop_p before pack to avoid SIGBUS on a poisoned loop-hint cache

tomholford:
`pack`/`meld` can `Bus error` every time if a prior `bail: meme` left a dangling noun in the loop-hint cache (`lop_p`). This drops `lop_p` before compacting so the rewriter never chases the corrupt pointer. Recovered a real stuck pier; same class as #626. Opus 4.8 was used to troubleshoot this issue I experienced with my ship and to implement this resolution. WOMM.

<details>
<summary>Details</summary>

## Summary

A `bail: meme` raised mid cache-insertion can leave a partially-built noun in the loop-hint set (`u3R->lop_p`). Nothing clears `lop_p` before `u3m_pack` — `u3m_reclaim` clears `har_p` and `|meld` clears `per_p`, but `lop_p` is untouched — so `u3a_rewrite_compact` → `u3h_relocate(lop_p)` later chases the dangling child pointer and faults. On a saved snapshot this is deterministic: every `pack`/`meld` dies with `Bus error` (`SIGBUS` / `BUS_ADRERR`) and the bad entry survives reboots.

## Changes

- `pkg/noun/manage.c` (`u3m_pack`): after `u3m_reclaim` and before `u3a_pack_seek`, re-point `u3R->lop_p` to a fresh table.
  - Intentionally does **not** `u3h_free` the old table — freeing walks the same dangling pointer. Re-pointing the root makes the stale entry unreachable, so the compactor never traverses it.
  - The orphaned table is unreferenced and reclaimed by the normal mark-sweep (`u3a_sweep`, via `u3m_grab` / `|mass`) — it doesn't persist. `_free_words` reclaims at the allocator level and never dereferences the bad child pointer.
  - `lop_p` is a rebuildable loop-detection hint cache.

## Validation

Reproduced and fixed on a live v4.6 planet whose `|bump` had bailed `%meme`:

- Stock v4.6: `pack` and `meld` both `Bus error` in `_pack_relocate_mark`, via `u3h_relocate` of `lop_p` (`allocate.c:1805`); gdb showed `si_addr` (a ~15 GB loom offset) landing in the LMDB `data.mdb` mmap → `BUS_ADRERR`.
- Patched v4.6: `pack` → rc 0; `meld` → rc 0, reclaimed ~391 MiB. The pier then booted on **stock** v4.6 and `|bump` completed; the next mark-sweep reclaimed the orphaned `lop_p` (~74 MB, printed as `palloc: sweep: leaked`) with no crash.

</details>

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -2777,6 +2777,17 @@ u3m_pack(void)
   //
   u3m_reclaim();
 
+  //  drop the loop-hint set before compacting.  a bail:meme raised mid
+  //  cache-insertion (see urbit/vere#626) can leave a dangling noun in lop_p;
+  //  u3a_rewrite_compact() then chases the corrupt pointer and faults (observed
+  //  as SIGBUS in _pack_relocate_mark, via u3h_relocate() of lop_p).  we
+  //  deliberately do NOT u3h_free() the old table -- freeing chases the same
+  //  bad pointer -- and instead re-point the root so the stale entry becomes
+  //  unreachable and the compactor skips it.  lop_p is a pure loop-detection
+  //  hint cache and is rebuilt on demand.
+  //
+  u3R->lop_p = u3h_new();
+
   //  sweep the heap, finding and saving new locations
   //
   u3a_pack_seek(u3R);


dozreg-toplud:
`bail: meme` on a home road aborts the process, `bail: meme` on a child road discards that road together with its `lop_p`. So `bail: meme` couldn't have placed `lop_p` in a bad state.

If you want to look at the treatment of other hashtables by analogy: `har_p` needs to be cleared because the values contain raw pointers and we do not bother with rewriting those. `per_p` is no longer cleared by default with `|meld`, you have to supply additional flags, so this is just a cleanup, it's not about the safety of the operation.

`lop_p` is closer to `per_p` in that regard: it's a hashtable that maps `[subject formula]` to `~`, in other words a hashset of nouns. It should be safe to move around.

As a reminder, `lop_p` is populated with a subject and a formula on entry into a `%loop`-hinted formula, and they are removed on exit. There are very few %loop hints in the code right now ~~(none in %base AFAIK, probably none in %groups I would guess)~~ _(I remembered that we have one in ford lightning to detect global build cycles)_, and in between event handling the hashtable should be empty (we can't put something in it and leave it there by bailing, as discussed above). An empty hashtable is 512 bytes big.

The fact that `lop_p` was huge (74 MB!) and malformed points at some other issue. **Was the ship's binary ever downgraded?** `lop_p`'s space in the road struct was carved out from the future-proof zero-initialized buffer, and moving back and forth between versions before the fix in #975 could cause subtle corruptions like that.

tomholford:
@dozreg-toplud Thanks for the informative response.

> The fact that `lop_p` was huge (74 MB!) and malformed points at some other issue. **Was the ship's binary ever downgraded?** `lop_p`'s space in the road struct was carved out from the future-proof zero-initialized buffer, and moving back and forth between versions before the fix in #975 could cause subtle corruptions like that.

This makes a lot more sense. Have been running this pier for some time now, it's been through the wringer for sure. I have probably done something dumb like this in the past. Closed as this is likely #975

PR #1063, opened 2026-07-20
palloc: crash on weak if u3o_leak_crash

dozreg-toplud:
Resolves #

### DIFF
--- a/pkg/noun/palloc.c
+++ b/pkg/noun/palloc.c
@@ -1780,16 +1780,16 @@ _sweep_counts(void)
     }
   }
   if ( weq_w ) {
-    if ( u3C.wag_w & u3o_verbose ) {
-      u3a_print_memory(stderr, "palloc: sweep: weaked", weq_w);
-    }
+    u3a_print_memory(stderr, "palloc: sweep: weaked", weq_w);
     if ( u3C.wag_w & u3o_leak_crash ) {
       u3_assert(!"weak");
     }
   }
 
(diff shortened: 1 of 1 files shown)


pkg/noun/palloc.c
```diff
@@ -1783,13 +1783,13 @@ _sweep_counts(void)
...
+      u3_assert(!"weak");
+    }
   }
 
-  if ( u3C.wag_w & u3o_verbose ) {
-    u3a_print_memory(stderr, "palloc: off-heap: used", u3a_Mark.len_w);
-    u3a_print_memory(stderr, "palloc: off-heap: total", u3a_Mark.siz_w);
-  }
+  u3a_print_memory(stderr, "palloc: off-heap: used", u3a_Mark.len_w);
+  u3a_print_memory(stderr, "palloc: off-heap: total", u3a_Mark.siz_w);
```
joemfb: The printf just shows how much space was used for metadata during the traversal. it should stay gated by `-v`, i's the one at line 1784 that should not be so gated (since will crash if there are any "weaks").

dozreg-toplud:
@joemfb maybe by the time we run tests urwasm batteries are already on the home road, so refcounting noops anyway?

dozreg-toplud:
confirmed with printouts

joemfb:
@dozreg-toplud I see, that makes sense.

joemfb (approved):

PR #1071, opened 2026-07-29
http: actually unlink first item; return early

dozreg-toplud:
If the first element of the linked list `htp_u->htd_u->htp_u` was the element we wanted to unlink, then the function would noop. 

Also, if an element got unlinked, we would keep iterating over the linked list instead of returning early. I am assuming that the linked list does not hold duplicate entries.

There is potential for more optimizations: if the linked list is guaranteed to hold `htp_u` then the while loop could be made unconditional, removing extra checks. Though I do not know if this actually holds.

### DIFF
--- a/pkg/vere/io/http.c
+++ b/pkg/vere/io/http.c
@@ -1908,6 +1908,7 @@ _http_serv_unlink(u3_http* htp_u)
     while ( pre_u ) {
       if ( pre_u->nex_u == htp_u ) {
         pre_u->nex_u = htp_u->nex_u;
+        return;
       }
       else pre_u = pre_u->nex_u;
     }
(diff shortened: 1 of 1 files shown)


dozreg-toplud:
Huh? Can't reproduce the crash with `--gc-abort` at [ff3942d](https://github.com/urbit/vere/pull/1071/commits/ff3942d63354144bd74c3a4e6d621288beb3eb3f)

dozreg-toplud:
What is going on? eddcfc9 and ff3942d should be the same, yet the latter failed in CI with a leak.

dozreg-toplud:
Let's get #1072 merged first, then I'll rebase to run the CI with UB fixes,

joemfb (approved):

PR #1072, opened 2026-07-30
ubsan fixes, alignment in u3a_pile

dozreg-toplud:
Strange CI behavior in #1071 prompted me to run CI tests locally with UBSan, and fix all places where it complained.

The most substantial fix is introduction of alignment in `u3a_pile`. I have a faint memory that something like this was supposed to be added to Vere but was never merged, so let's discuss it again. In `u3a_pile_prep` we align the stack and store both stack offsets before and after the alignment. In `u3a_pile_done` we compare `cap_p` with the post-alignment value, and if the stack is empty we restore it to the pre-alignment value. Alignment is assumed to be a power of two, and the input alignment is in bytes, just like the length.

In unix.c ubsan complained that we increment a null pointer, which is supposed to be illegal? I refactored it slightly so that we formally increment it only if it is not null.

In mars.c the function pointer type and the actual function were technically incompatible.

The rest is `1 << X` type fixes.

### DIFF
--- a/pkg/noun/manage.c
+++ b/pkg/noun/manage.c
@@ -1022,7 +1022,7 @@ u3m_leap(c3_w pad_w)
   {
     u3a_pile pil_u;
     c3_p     ptr_p;
-    u3a_pile_prep(&pil_u, sizeof(u3a_road) + 15); // XX refactor to wiseof
+    u3a_pile_prep(&pil_u, sizeof(u3a_road) + 15, alignof(u3a_road)); // XX refactor to wiseof
     ptr_p = (c3_p)u3a_push(&pil_u);
 
     //  XX add push_once, push_once_aligned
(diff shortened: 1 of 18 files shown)


dozreg-toplud:
The faint memory was about #906, cc @yosoyubik

pkg/noun/manage.c
```diff
@@ -1022,7 +1022,7 @@ u3m_leap(c3_w pad_w)
   {
     u3a_pile pil_u;
     c3_p     ptr_p;
-    u3a_pile_prep(&pil_u, sizeof(u3a_road) + 15); // XX refactor to wiseof
+    u3a_pile_prep(&pil_u, sizeof(u3a_road) + 15, alignof(u3a_road)); // XX refactor to wiseof
```
joemfb: We can get rid of the `+15` and subsequent manual alignment here, and just pass `16` instead of `alignof(u3a_road)`.

joemfb (changes requested):

PR #1078, opened 2026-08-04
build: fixes `-Dgenerate-commands`

matthew-levan:
## Summary

`zig build -Dgenerate-commands` produced a `compile_commands.json` that `clangd`
couldn't use, making editor diagnostics unreliable across the repo. Two
independent defects in `CdbGenStep`:

1. **Invalid JSON** — fragments were joined without stripping zig's trailing
   comma, so every entry was separated by `},,{`. The file never parsed, and
   clangd silently fell back to inferred flags rather than reporting an error.

2. **Incomplete databases** — the step rebuilt the database from only the
   fragments a given build happened to produce, then deleted them. Incremental
   builds truncated it to a handful of entries, fully cached builds wrote
   nothing at all, and `*_tests.c` never appeared. Fragments now accumulate,
   with each flag set getting its own directory so entries can't mix.

Details are in the commit messages.

## Verification

Verified on `develop` with both commits applied:

- full generation → valid JSON, 255 entries
- fully cached rerun → database preserved, not truncated
- `zig build -Dgenerate-commands nock-test`, then rerun → `nock_tests.c` collected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ MODULE.bazel.lock
 # clangd
 **/.cache
 compile_commands.json
-/cdb
+/cdb-*
 
 # Swap files.
 *.swo
--- a/build.zig
+++ b/build.zig
@@ -28,7 +28,16 @@ const CdbGenStep = struct {
         var cwd = std.fs.cwd();
 
         // Open fragments directory (created by zig via -gen-cdb-fragment-path).
+        //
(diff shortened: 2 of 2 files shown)


dozreg-toplud (approved):

matthew-levan:
@pkova can we merge this tomorrow?

PR #1079, opened 2026-08-07
build: change distribution links

dozreg-toplud:
Mirror links keep 404'ing


dozreg-toplud:
@pkova maybe we should just vendor those tarballs?

PR #1081, opened 2026-08-14
windows: implements demand paging and sparse loom

matthew-levan:
Demand paging and a sparse loom on Windows, plus the migration path that depended on both.

Measured on a 2GB Windows VM, second boot against a 786MB snapshot (serf process):

| | `develop` | this branch |
|---|---:|---:|
| working set | 781,788 K | **29,824 K** |
| I/O read bytes at boot | 785,941,999 | **34,288** |
| page faults | 196,314 | **8,322** |

Running the same build with `--no-demand` reproduces the `develop` numbers to within 0.03%, which is the control: demand paging accounts for the whole difference, and the fallback path is unchanged.

## Why this needed a subsystem

The loom leans on three POSIX behaviours Windows lacks.

**`mmap(MAP_FIXED)` atomically replaces part of an existing mapping.** `MapViewOfFileEx` fails if the address is occupied, and `UnmapViewOfFile` only takes a whole view's base, so you cannot punch a hole in the loom to drop the image into. Placeholder reservations (Win10 1803+) do provide it: reserve a range, split it, replace each piece with a mapping.

**Overcommit.** A pagefile-backed section is charged against RAM + pagefile at *creation*, so a 16GB loom would not boot on a 2GB box at all — this failed before demand paging could be exercised. `SEC_RESERVE` makes the section's pages reserved rather than committed.

**Uniform `mprotect`.** `VirtualProtect` cannot span mappings, cannot touch reserved pages, and rejects `PAGE_READWRITE` on a copy-on-write view. This is a series of small adjustments at existing call sites.

## Layout

`pkg/noun/platform/windows/wloom.{c,h}` holds all the Win32. `events.c`, `manage.c` and `disk.c` call into it and never see a `HANDLE`.

```
[0, map_i)      copy-on-write view of image.bin, protected read-only
[map_i, len_i)  view of a SEC_RESERVE pagefile section, committed on touch
```

Two decisions carry the weight.

**The volatile half is a section view, not private memory.** Moving the image boundary means unmapping and remapping it, and Windows cannot resize a private region from below without discarding it — the heap and stack would go too. Section contents outlive their views. The view sits at section offset == loom offset, so addresses are stable across boundary moves.

**The image is mapped `PAGE_WRITECOPY`, then protected down to `PAGE_READONLY`.** Stores still fault into `u3e_fault` rather than being silently privatized. The view must be *created* write-copy; a read-only view cannot be upgraded later.

`VirtualAlloc2` and `MapViewOfFile3` are linked from `api-ms-win-core-memory-l1-1-6`, which sets the floor at Win10 1803. Imports resolve eagerly, so an older system refuses to start the binary rather than degrading — deliberate, and documented in `wloom.c`.

## Sparseness

Pages commit on first touch. `u3m_fault` resolves that before `u3m_water`, which dereferences the road — the road lives in the loom and may itself be untouched, so a first touch there would fault inside the handler. The vectored handler now routes read faults as well as writes, since a reserved page faults on read too; that matches libsigsegv, which always delivered both.

Ranges the kernel writes into (`pread` into the loom) are committed up front, because a handler cannot rescue a kernel-mode access. The guard page is committed before being protected. `u3e_yolo` walks region by region, skipping reserved pages rather than committing the whole loom.

Sparseness survives a save because `_ce_patch_save_page` already skips pages the allocator marks `u3a_free_pg`, so untouched memory is never read back.

`u3m_init` takes the placeholder path regardless of `u3o_no_demand`; that flag now only chooses whether the image is mapped or blitted.

## Migration

`disk.c` mapped the stale loom anonymously and then mapped the old image over it with `MAP_FIXED`. Both halves fail on Windows. The stale loom is now **read, not mapped**: the stale image and the migrated snapshot are the same file, and Windows keeps a file's section attached to the handle that created it, so any mapping holds that file against every later resize. No arrangement of mappings is correct while source and destination are one file.

`u3e_live` also no longer loads an image the caller is about to discard. A migration calls it and then `u3m_pave(c3y)` immediately, so every page loaded was thrown away — and under demand paging each subsequent write took a copy-on-write fault and a page-in against a snapshot nobody would read. That was why `--no-demand` was *faster* for a migration, which is otherwise backwards.

The stale loom bases are fixed by pointer arithmetic but were claimed lazily, long after the loader and CRT had placed theirs; a 64→32 migration hit `ERROR_INVALID_ADDRESS` at `u3_Loom_d`, and the collision moved on rebuild. They are now staked at startup, while the address space is still empty.

Four ordering bugs fell out, all latent on POSIX:

- `_disk_migrate_loom` released the stale loom *after* `u3m_save`; `_disk_migrate_h` had been fixed this way already, its sibling had not.
- `_ce_loom_unmapf` dropped the live image view but left its descriptor open, and the section stays attached to that descriptor.
- `u3m_init` runs again after a migration and was not idempotent; `MAP_FIXED` had been covering for it.
- a zeroed `c3_o` reads as `c3y`, so every region slot looked occupied.

## Stack overflow

A C stack overflow reported `dig: over` on Linux and killed the serf on Windows. Two causes, both needing fixes:

POSIX runs the handler on a dedicated alternate stack (`Sigstk`, via libsigsegv). Windows has no equivalent — a vectored handler runs on the stack that just overflowed — so `SetThreadStackGuarantee` reserves the headroom.

And Windows raises `EXCEPTION_STACK_OVERFLOW` once: catching it consumes the thread's guard page and nothing puts it back, so the second overflow has nothing to trip. `_resetstkoflw` restores it once the stack has unwound.

## Not supported

`--swap` errors out rather than failing obscurely; the ephemeral file is remapped a page at a time, which the 64KB allocation granularity precludes against a 16KB loom page. Snapshot validation still forces `u3o_no_demand` on Windows, since it reads the loom in the window where the image is unmapped.

## Testing

`wloom_probe.c` (standalone, not built) checks the Win32 properties the design rests on and that the docs do not state outright — above all that a section's contents *and* commitment survive unmapping and remapping its view, which if false would corrupt a snapshot silently rather than failing. It also measured the toss mechanisms: `DiscardVirtualMemory` frees pages and leaves them reading as zero, matching `MADV_DONTNEED`.

`wloom-test` (Windows targets only) covers address staking end to end, including the collision it prevents. Both pass 9/9 on hardware, along with the existing unit tests and both migration directions.

## Follow-ups, not in this PR

A Windows CI runner (build, unit tests, loom probe, migration roundtrip) was written and set aside; it needs a fix to `_disk_release`, which unlinks `.vere.lock` while its descriptor is still open — harmless on POSIX, but Windows refuses, so the lockfile outlives the pier. That is an independent bug, not caused by this work.


matthew-levan:
# Boot and run tests between optimized 32-bit builds (`-Doptimize=ReleaseFast`) of `ml/64-mingw-sparse` @ abb5d17d07e4f4f18ba4b25bd5776c5060fd08f9 vs. `ml/64` @ dbe52c39dac031ff6ff7ca5d77fd1d919559f8f2

---

### Post fresh-boot (no pier restart) with `--urth-loom 29 --loom 31`:

`ml/64-mingw-sparse`

sparse-zod-boot

`ml/64`

64-zod-boot

### After restarting pier (`|exit` then `urbit zod --urth-loom 29 --loom 31`):

`ml/64-mingw-sparse`

sparse-zod-work

`ml/64`

64-zod-work

### After restarting pier with demand paging OFF (`|exit` then `urbit zod --urth-loom 29 --loom 31 --no-demand`

`ml/64-mingw-sparse`

sparse-zod-work-no-demand

`ml/64`

64-zod-work-no-demand

### After restarting the pier results table (mars)

Snapshot is ~786 MB. `--urth-loom 29 --loom 31`.

| Metric | `ml/64-mingw-sparse` | `ml/64` | sparse `--no-demand` | `ml/64` `--no-demand` |
| --- | ---: | ---: | ---: | ---: |
| Working set | **29,824 K** | 781,788 K | 782,020 K | 782,020 K |
| Peak working set | **29,824 K** | 781,788 K | 782,028 K | 782,020 K |
| Memory (private working set) | 15,232 K | 2,256 K | 2,268 K | 2,292 K |
| Memory (shared working set) | 14,592 K | 779,532 K | 779,752 K | 779,728 K |
| Commit size | 190,460 K | 170,868 K | 170,816 K | 170,904 K |
| Page faults | **8,322** | 196,314 | 196,382 | 196,349 |
| I/O read bytes | **34,288** | 785,941,999 | 785,942,000 | 785,942,000 |

**26× less resident memory, ~23,000× less file I/O at boot, 24× fewer page faults.**

`ml/64` reads the entire snapshot into the loom at startup — I/O read bytes is almost exactly the size of `image.bin`. The sparse branch maps it instead, so it reads 34 KB and only faults in the ~29 MB the ship actually touches.

The three right-hand columns agree to within 0.03% on every metric, which is the control: demand paging accounts for the whole difference, and `--no-demand` on the sparse branch reproduces `ml/64` exactly — so the fallback path (also taken on Windows older than 10 1803) is unchanged.

Commit size is *higher* on sparse by ~20 MB, as expected: the image is mapped `PAGE_WRITECOPY`, so pages the ship dirties become private copies, which this column counts. `ml/64` keeps the loom in one pagefile section, where the cost sits in shared working set instead.

The urth process is unaffected in all four runs (~35 MB working set, ~24k page faults, ~6 KB read).

The fresh-boot pair is confounded: a fake zod saves during boot, and the two samples are three minutes apart against a snapshot timer, so some of that 488 vs 910 MB gap is likely one process having saved and remapped while the other hadn't. The restart test is the controlled one.

These numbers don't measure sparseness: every column is per-process, and `SEC_RESERVE` only shows in system commit charge.

### Boot time

`ml/64-mingw-sparse`

```
C:\Users\Administrator\Desktop\vere>powershell -Command "Measure-Command { .\zig-out\x86_64-windows-gnu\urbit.exe -F nec --urth-loom 29 --loom 31 -x -t }"
disk: loaded epoch 0i0
loom: mapped 2048MB
boot: protected loom
live: logical boot
boot: installed 2179 jets
boot: parsing %brass pill
disk: loaded epoch 0i0
loom: mapped 2048MB
boot: protected loom
live: mapped: MB/785.940.480
boot: installed 2179 jets


Days              : 0
Hours             : 0
Minutes           : 2
Seconds           : 6
Milliseconds      : 910
Ticks             : 1269105048
TotalDays         : 0.00146887158333333
TotalHours        : 0.035252918
TotalMinutes      : 2.11517508
TotalSeconds      : 126.9105048
TotalMilliseconds : 126910.5048
```

`ml/64`

```
C:\Users\Administrator\Desktop\vere-ref>powershell -Command "Measure-Command { .\zig-out\x86_64-windows-gnu\urbit.exe -F nec --urth-loom 29 --loom 31 -x -t }"
disk: loaded epoch 0i0
loom: mapped 2048MB
boot: protected loom
live: logical boot
boot: installed 2179 jets
boot: parsing %brass pill
disk: loaded epoch 0i0
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/785.940.480
boot: installed 2179 jets


Days              : 0
Hours             : 0
Minutes           : 2
Seconds           : 8
Milliseconds      : 368
Ticks             : 1283686536
TotalDays         : 0.00148574830555556
TotalHours        : 0.0356579593333333
TotalMinutes      : 2.13947756
TotalSeconds      : 128.3686536
TotalMilliseconds : 128368.6536
```

matthew-levan:
# The Windows loom: a lifecycle walkthrough

How vere initializes, runs, and releases the snapshot with demand paging enabled on Windows. Written for someone who knows `events.c` and POSIX memory management, and not much Win32.

Paths are relative to the repository root.

## 0. Five Windows concepts

POSIX collapses these into `mmap`; Windows splits them, and nearly every design decision in this subsystem falls out of the split.

| Concept | What it is | POSIX analogue |
|---|---|---|
| **Section object** | A kernel object representing *backing store*, created by `CreateFileMapping`. Backed by a file, or by the pagefile if you pass `INVALID_HANDLE_VALUE`. | the `fd` argument to `mmap` (or `MAP_ANON`) |
| **View** | An actual mapping of a section into your address space, via `MapViewOfFile*`. | the return value of `mmap` |
| **Reserve vs commit** | `MEM_RESERVE` takes address space only. `MEM_COMMIT` promises backing store, charged against RAM + pagefile. | no equivalent — Linux overcommits, so every mapping is effectively "reserved" |
| **Placeholder** | `MEM_RESERVE_PLACEHOLDER` — a reservation you are permitted to later *split* and *replace* with a view. Windows 10 1803+. | `MAP_FIXED` |
| **VEH** | Vectored exception handler — a callback for access violations. | `SIGSEGV` handler |

The one that trips people up: **a section can itself be reserved rather than committed** (`SEC_RESERVE`). That is what makes the loom sparse.

## Minimum Windows

`VirtualAlloc2` and `MapViewOfFile3` arrived in Windows 10 1803, and are exported from the `api-ms-win-core-memory-l1-1-6` API set rather than from `kernel32`. Linking them (see the `linkSystemLibrary` calls in [build.zig](build.zig) and [pkg/noun/build.zig](pkg/noun/build.zig)) therefore sets vere's floor at that release: Windows resolves imports eagerly at load, so an older system refuses to start the binary at all rather than degrading. That floor is deliberate.

## 1. Boot, part one: reserving the loom

[`u3m_init`](pkg/noun/manage.c#L2532) → [`u3_wnd_loom_init`](pkg/noun/platform/windows/wloom.c#L424) → [`_wnd_reserve`](pkg/noun/platform/windows/wloom.c#L346).

`_wnd_reserve` makes three Win32 calls in order:

1. `CreateFileMappingW(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE | SEC_RESERVE, …)` — creates a **pagefile-backed section** the size of the whole loom. This is the loom's anonymous memory. `SEC_RESERVE` is the crucial flag: its pages are reserved, not committed, so creating it costs no commit charge. Without it, Windows charges the entire loom against RAM plus paging file immediately — that was the original `ERROR_COMMITMENT_LIMIT` failure on a small box.

2. `VirtualAlloc2(NULL, bas_v, len_i, MEM_RESERVE | MEM_RESERVE_PLACEHOLDER, PAGE_NOACCESS, …)` — reserves the loom's address range as a **placeholder**. This is the piece that makes everything else possible: a plain reservation cannot be replaced by a mapping, but a placeholder can.

3. `MapViewOfFile3(sec_h, NULL, bas_v, 0, len_i, MEM_REPLACE_PLACEHOLDER, PAGE_READWRITE, …)` — maps the section over the entire placeholder.

The function writes into the slot *only on success* ([wloom.c:391](pkg/noun/platform/windows/wloom.c#L391)), so a failed reservation leaves it free rather than half-claimed.

After this the loom is one view of a reserved section, with nothing committed. Functionally this is `mmap(MAP_ANON|MAP_PRIVATE)` on Linux-with-overcommit.

## 2. Boot, part two: mapping the image

[`u3e_live`](pkg/noun/events.c#L1523) opens `image.bin`, applies any leftover patch, calls `u3e_foul()` to mark every page dirty, then reaches the demand-paging branch and calls the Windows [`_ce_loom_mapf`](pkg/noun/events.c#L1098).

That function is short and mostly arithmetic:

```c
gan_w = granularity_bytes >> (u3a_page + u3a_word_bytes_shift);  // 64KB/16KB = 4
map_w = pgs_w & ~(gan_w - 1);   // granule-floored page count
tal_w = pgs_w - map_w;          // ragged tail, 0..3 pages
```

Placeholders can only be split at the 64KB allocation granularity, but a loom page is 16KB. So the image is mapped up to the last whole 64KB granule, and the leftover ≤3 pages are `pread` into memory by [`_ce_loom_blit_pages`](pkg/noun/events.c#L1017) — which commits them first, because a kernel-mode write (`ReadFile` into your buffer) cannot be rescued by a fault handler.

The real work is [`_wnd_remap`](pkg/noun/platform/windows/wloom.c#L264), the heart of the subsystem:

1. [`_wnd_image`](pkg/noun/platform/windows/wloom.c#L238) — `CreateFileMappingW(fil_h, NULL, PAGE_WRITECOPY, …)`, a copy-on-write section over `image.bin`. Deliberately done **before** any teardown, so a failure here is not destructive.

2. [`_wnd_hold`](pkg/noun/platform/windows/wloom.c#L207) — collapse whatever is mapped back into one placeholder. `UnmapViewOfFileEx(bas, MEM_PRESERVE_PLACEHOLDER)` unmaps the view *but keeps the address range reserved as a placeholder* — that flag is the difference between "give the address space back" and "hold it for me". If an image view already existed, the high view is unmapped too and `VirtualFree(…, MEM_RELEASE | MEM_COALESCE_PLACEHOLDERS)` merges the two placeholders back into one.

3. `VirtualFree(bas, byt_i, MEM_RELEASE | MEM_PRESERVE_PLACEHOLDER)` — **splits** the placeholder at the image boundary. Same call, different flag, opposite meaning from step 2's coalesce.

4. `MapViewOfFile3(img_h, …, MEM_REPLACE_PLACEHOLDER, PAGE_WRITECOPY, …)` — image into the low placeholder.

5. `VirtualProtect(bas, byt_i, PAGE_READONLY, &old_u)` — immediately downgrade to read-only. **This two-step is not optional.** The view must be *created* write-copy so it is permitted to be dirtied privately; but it must *run* read-only so stores trap into `u3e_fault` for dirty tracking. A view created `PAGE_READONLY` can never be upgraded later.

6. `MapViewOfFile3(sec_h, …, bas + byt_i, offset byt_i, len - byt_i, …)` — the pagefile section over the remainder.

Two things worth internalising about step 6:

- **The offset equals the loom offset.** Loom address *X* always maps to section offset *X*. That is what keeps addresses stable when the boundary moves.
- **The volatile half is a section view, not private memory.** Moving the boundary means unmapping and remapping this half, and Windows cannot resize a private region from below without discarding it — the heap and stack would go with it. Section contents outlive their views.

## 3. Running: three kinds of fault

Everything enters through [`_windows_exception_filter`](pkg/noun/platform/windows/veh_handler.c#L10). `ExceptionInformation[0]` is 0 for a read, 1 for a write; `[1]` is the faulting address. Returning `EXCEPTION_CONTINUE_EXECUTION` re-runs the faulting instruction, which is how "fix it and retry" works.

It calls [`u3m_fault`](pkg/noun/manage.c#L2322), which handles three cases in a deliberate order.

**(a) First touch of a reserved page** — the sparse case. Resolved *first*, at [manage.c:2331](pkg/noun/manage.c#L2331), via [`u3_wnd_loom_fault`](pkg/noun/platform/windows/wloom.c#L486): `VirtualQuery` the address, and if `State == MEM_RESERVE`, `VirtualAlloc(…, MEM_COMMIT, PAGE_READWRITE)`.

The ordering is load-bearing for two reasons. `u3m_water` runs right after and dereferences `u3R` — which *lives in the loom* and may itself be untouched, so a first touch there would fault inside the fault handler. And the loom bounds check comes after too, because a migration holds a stale loom outside the live loom's range.

**(b) Write to a clean image page** — the demand-paging case. Falls through to [`u3e_fault`](pkg/noun/events.c#L246), which sets the dirty bit and calls [`_ce_flaw_mprotect`](pkg/noun/events.c#L170). That is plain `mprotect(PROT_READ|PROT_WRITE)` — but the shim in [compat.c](pkg/c3/platform/windows/compat.c#L314) `VirtualQuery`s the range and substitutes `PAGE_WRITECOPY` when `AllocationProtect` is write-copy, because `VirtualProtect` *rejects* `PAGE_READWRITE` on a CoW view. It keys off `AllocationProtect`, not `Protect`, since an already-copied page reports `PAGE_READWRITE` while its view is still write-copy.

**(c) The guard page** — unchanged from POSIX, except that `gar_w == 0` now means *unposted* rather than page zero ([events.c:250](pkg/noun/events.c#L250)).

## 4. Saving

[`u3e_save`](pkg/noun/events.c#L1404), in order:

1. `_ce_patch_compose` — walk the dirty bitmap, hash and write dirty pages to `memory.bin` / `control.bin`. Allocator-free pages are skipped, which is why sparseness survives a save.
2. `_ce_patch_sync`, `_ce_patch_verify`.
3. **[`_ce_loom_unmapf`](pkg/noun/events.c#L1060)** — the new hook, a no-op on POSIX. Two jobs: `u3_wnd_loom_unmapf` drops the image view (`_wnd_remap` with `byt_i = 0`, so the pagefile section covers the whole loom again), *and* it closes and reopens `u3P.img_u.fid_i`. Windows keeps a file's section attached to the handle that created it, so unmapping alone does not release the file — the descriptor has to be cycled.
4. `_ce_patch_apply` → [`_ce_image_resize`](pkg/noun/events.c#L810) → [`u3_wnd_truncate`](pkg/noun/platform/windows/wloom.c#L168) (`SetFilePointerEx` + `SetEndOfFile`, used instead of `ftruncate` so failures report a real Win32 error), then `pwrite`s the patch pages into `image.bin`.
5. `_ce_loom_mapf` again — remap at the new boundary.

Step 3 exists entirely because Windows refuses to truncate a mapped file. On POSIX steps 3–5 are just `mmap(MAP_FIXED)` doing the right thing.

Step 5 is the boundary-move case, and the one still untested by `wloom_probe.c`.

## 5. Release

[`u3e_stop`](pkg/noun/events.c#L1666) closes the fd and frees the bitmap; the loom itself goes away at process exit. If `u3m_init` runs a second time in the same process — which happens after a migration — [`u3_wnd_loom_init`](pkg/noun/platform/windows/wloom.c#L424) calls [`_wnd_release`](pkg/noun/platform/windows/wloom.c#L599) first, making the reservation idempotent the way `MAP_FIXED` always was.

`_wnd_release` unmaps both views and releases **each half separately**, since a split reservation is two allocations and `VirtualFree(bas, 0, MEM_RELEASE)` would only free the first.

## Tracing it live

The two highest-value breakpoints:

- [`_wnd_remap`](pkg/noun/platform/windows/wloom.c#L264) — every mapping change goes through it.
- [`u3_wnd_loom_fault`](pkg/noun/platform/windows/wloom.c#L486) — every first touch.

The two places where Windows semantics are genuinely unlike anything in POSIX are the placeholder split/coalesce dance in `_wnd_remap` steps 2–3, and the copy-on-write protection two-step in steps 4–5.

matthew-levan:
Now with `u3_wnd_loom_toss`, we "toss" unused memory back to the Windows kernel via `DiscardVirtualMemory` so it can reclaim it from the process's working set. You can see here in this screenshot that, after fresh booting a fakezod (and without restarting it) and idling in `dojo` for a few moments, the mars process only needs a fraction (~20MB) of the memory which was used for booting (~2GB).

sparse-zod-toss

PR #1096, opened 2026-09-03
ml/64 mig debugging

matthew-levan:
Resolves @mopfel-winrux's `bail:foul` and similar 32 -> 64-bit migration crashes. See below stacktrace for reference:

```
root@nativeplanet:/home/nativeplanet# docker logs dacrex-sanneb -f
BOOT SHIP
Running with no STDIN
urbit -t -p 30032 --http-port 22947 --loom 33 --snap-time 60 dacrex-sanneb
Probing pier with urbit -Lx...
~
urbit 4.6-44cf010
boot: home is /urbit/dacrex-sanneb
loom: mapped 16384MB
lite: arvo formula 4ce68411
lite: core 641296f
lite: final state 641296f
disk: loaded epoch 0i1026150615
loom: mapped 8192MB
loom: 0x2200000000 fid_i 17 len 3898572800
loom: mapped 8192MB
boot: protected loom
live: mapped: GB/3.898.572.800
loom: 32->64 migration running...

bail: foul
home: bailing out

Stacktrace:
0   u3m_stacktrace                      pkg/noun/manage.c:901
1   u3m_bail                            pkg/noun/manage.c:0
2   u3a_gain                            pkg/noun/allocate.c:0
3   _copy_32_noun                       pkg/noun/copy_migrate.h:294
4   _copy_32_hamt                       pkg/noun/copy_migrate.h:312
5   _ch_walk_node_h                     pkg/noun/hashtable.c:930
6   _ch_walk_node_h                     pkg/noun/hashtable.c:938
7   u3h_walk_with_h                     pkg/noun/hashtable.c:962
8   u3_migrate_d                        pkg/noun/migrate.c:93
9   _disk_migrate_loom                  pkg/vere/disk.c:1819
10  _disk_epoc_load                     pkg/vere/disk.c:2044
11  u3_disk_load                        pkg/vere/disk.c:2321
12  u3_mars_load                        pkg/vere/mars.c:1466
13  _cw_work                            pkg/vere/main.c:3112
14  _cw_utils                           pkg/vere/main.c:3207
15  main                                pkg/vere/main.c:3233
16  libc_start_main_stage2              /home/runner/work/_temp/abc83e4f-167e-415d-9e4e-5beaec47c0c4/zig-x86_64-linux-0.15.2/lib/libc/musl/src/env/__libc_start_main.c:
```

### DIFF
--- a/pkg/noun/copy_migrate.h
+++ b/pkg/noun/copy_migrate.h
@@ -1,7 +1,12 @@
 /* copy_migrate.h: template for full-noun-copy migrations.
 **
 **   Each copying migration in pkg/past/ shares the same skeleton:
-**     - verstable dedup map keyed by old noun
+**     - verstable dedup map keyed by old noun; the map RETAINS every new
+**       noun it records, and releases them all in _done().  Without that,
+**       any free during the walk (a cache table evicting past max_w, or a
+**       put replacing an entry whose key compares equal) would leave the
+**       map pointing at dead nouns, and the next dedup hit would gain a
+**       freed box (bail: foul) or worse.
 **     - iterative DFS descent over the source noun with a frame stack
 **     - HAMT walker that copies key/val pairs into the target loom
 **
@@ -256,7 +261,7 @@ U3C_SYM(_next)(U3C_SYM(_ctx) *cop_u, U3C_OLD_NOUN old)
 
     if ( c3n == U3C_OLD_IS_CELL(old) ) {
       u3_atom new = U3C_SYM(_atom)(old);
-      vit_u = vt_insert(&(cop_u->map_u), old, new);
+      vit_u = vt_insert(&(cop_u->map_u), old, U3C_NEW_A_GAIN(new));
       u3_assert( !vt_is_end(vit_u) );
       return new;
     }
@@ -295,7 +300,7 @@ U3C_SYM(_noun)(U3C_SYM(_ctx) *cop_u, U3C_OLD_NOUN old)
     }
     else {
       new = U3C_NEW_I_CELL(top_u->hed, new);
-      vit_u = vt_insert(&(cop_u->map_u), top_u->cel, new);
+      vit_u = vt_insert(&(cop_u->map_u), top_u->cel, U3C_NEW_A_GAIN(new));
       u3_assert( !vt_is_end(vit_u) );
       cop_u->len_w--;
     }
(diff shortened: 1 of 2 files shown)


matthew-levan:
The bug is a use-after-free between two data structures that disagree about who owns the copied nouns.

The setup. A copying migration walks the old loom and builds each noun in the new loom. To preserve sharing, it keeps a dedup map from old reference to new reference. When the walk reaches an old noun it has already copied, it looks the new one up, gains a reference, and returns it instead of copying again. That map holds plain, unretained references (no refcount bumping). The only owner of a fresh copy is whatever structure the copy was just placed into: its parent cell, or the target hash table once the entry is put.

The free. The three tables being rebuilt are copied with u3h_put, and u3h_put can drop a pair it already holds. Two ways. If the table is a cache created with a max_w, the put after it fills evicts the pair at the clock arm. If the incoming key compares structurally equal to an existing key, the put replaces that pair in place. Both paths call u3z on the old pair, which recursively frees its key and value, and every sub-noun that nothing else references.

The dangling map. Those freed nouns are still recorded in the dedup map, because nothing ever removes them. The next time the walk meets an old noun that maps to one of them, it gains a freed box. In the loose case the box has a zero refcount and u3a_gain bails with foul, which is the first ship. In the worse case the box has already been reused for something else. The gain increments a stranger’s header, the new cell points into unrelated memory, and later refcount and equality traffic follows garbage references. Since a 64-bit reference decodes to any address above the loom base, that garbage can reach the privately mapped 32-bit source and corrupt it in place. That is the second ship, faulting on a source atom whose header had been overwritten.

Why it hid. Fresh piers never displace anything. The persistent cache starts far below its cap, keys are unique, so every put is a pure insert and the map stays valid. Long-lived ships fill the cache and can accumulate exactly the conditions that make a put drop something. The migration test fixtures are fresh piers, so CI never saw it.

The fix. Make the dedup map an owner. Each new noun is gained as it is recorded, so the map holds a reference alongside the parent or table. A put may still displace a pair, and the pair’s sub-nouns may still lose their structural owner, but the map’s reference keeps them alive and valid for any later dedup hit. When the walk is over and no lookup can happen again, _done iterates the map and loses each value. Nouns that are still reachable drop back to their proper count. Nouns that were displaced hit zero and are freed then, which is where they should have been freed all along.

What it costs. One refcount increment per copied noun and one decrement at the end, plus a pass over the map. It needs no knowledge of which target tables are caches, no cache-size juggling, and no special handling for either displacement path, which is why it beats both the deferred-free list and unbounding the cache.

Issue #490, opened 2023-07-10
Using recv with small buffer can crash Urbit


guaraqe:
It is possible to crash a Urbit by taiking to its socket (such as in `click`), but reading too few bytes. To test, you can run a command with `click`, but change the script so the size of the buffer in `nc` is something very small, such as 2.

This has affected hosting, when reading the vats of ships with too many apps installed. This is under control already, but it would be nice to solve.

@lukebuehler

matthew-levan:
I'm having trouble reproducing this on my Ubuntu machine with the instructions you provided here; would you mind posting a more detailed reproduction? I will continue tinkering in the meantime.

ashelkovnykov:
@matthew-levan Issue should probably be moved to [here](https://github.com/urbit/tools), since it's not an issue with `khan.hoon` or `conn.c`, but with the click script

guaraqe:
@ashelkovnykov We observed it in production, without using `click` at all, so this is not related to the script itself. I am writing a reproducer.

guaraqe:
I added a reproducer to the https://github.com/guaraqe/urbit-benchmark repo, in this commit: https://github.com/guaraqe/urbit-benchmark/commit/499ef778710ba716aa948c830addce10548d380d

I couldn't reproduce it with `netcat`, so I wrote a simple clone with Python, and got the correct result. If you run `./code-click-python ../salsyp-samzod` you should get the expected result, but if you change `SIZE` in the Python script from `1024` to `1`, the Urbit running at `../salsyp-samzod` should break with many messages like:

```
~salsyp_samzod:dojo> newt: write failed broken pipe
conn: moor bail -32 broken pipe
~salsyp_samzod:dojo> newt: write failed broken pipe
conn: moor bail -32 broken pipe
~salsyp_samzod:dojo> newt: write failed broken pipe
```

ashelkovnykov:
> We observed it in production, without using `click` at all, so this is not related to the script itself. I am writing a reproducer.

@guaraqe  Right, my apologies - misread

joemfb:
@guaraqe the expected behavior on early disconnection is that those errors are printed and the socket is closed, but the ship should stay up and not crash. Is that what you're seeing?

The newt wire-framing includes a 5 byte header: one tag byte (`0`) and 4 bytes of little-endian length. The right pattern for consuming it is to read 5 bytes, then read the length specified, then read 5 bytes, &c.

matthew-levan:
[...]

PR #1101, opened 2026-09-05
conn: make chan teardown idempotent and stop re-entering bail on a dead pipe

### LINKED ISSUE #1100: conn.c: King segfaults when a client disconnects with replies in flight (broken pipe -> loom fault)
### Description

The King (Urth) process can be **segfaulted by a client of `conn.c`** (the `.urb/conn.sock` control socket) that disconnects while the ship still owes it a reply. This is a remote-ish denial of service against any tooling that drives a ship over `conn.sock`: a single misbehaving or interrupted client takes the whole ship down, losing the running process (though not the event log).

This looks like the same underlying fault as the still-unresolved #490 ("Using recv with small buffer can crash Urbit"), but reproduces without a small `recv` buffer and without `click` — simply by **closing the socket with replies still in flight**.

### Environment

- `vere 4.6` (Linux x86_64, musl/zig build), `%zuse 408` base
- Reproduced on two ships: a fake `~feb` and a real production ship. Both crashed identically.

### Symptom

The console prints, then the process dies:

```
~sampel:dojo> newt: write failed broken pipe
conn: moor bail -32 broken pipe
~sampel:dojo> newt: write failed broken pipe
loom: external fault: 0x10 (0x200000000 : 0x280000000)

Stacktrace:
0   u3m_stacktrace                      pkg/noun/manage.c:870
1   u3m_fault                           pkg/noun/manage.c:2315
2   sigsegv_handler                     handler-unix.c:269
3   -                                   restore.s
4   u3_king_commence                    pkg/vere/king.c:1093
5   main                                pkg/vere/main.c:3389
6   libc_start_main_stage2              __libc_start_main.c
Segmentation fault (core dumped)
```

Sometimes it survives the first few broken pipes and crashes on a later one, which suggests a use-after-free / freed-noun access rather than a deterministic null deref.

### Reproduction

Speak the `conn.c` newt protocol directly (5-byte header: `0x00` + 4-byte LE length, then a jammed `[request-id command args]` noun), **pipeline many requests, and close the socket without reading the replies**. Minimal Python (no `click`, no `nc`):

```python
import socket
# jam/newt helpers omitted for brevity; encode [rid [%peek %| %once view desk spur]] nouns
s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.connect('/path/to/pier/.urb/conn.sock')
for i in range(400):
    s.sendall(newt(( i, peek_once('e', 'bindings') )))       # any non-trivial scry
    s.sendall(newt(( 1000+i, peek_once('bx', '', 'debug/timers') )))
s.close()   # <-- vanish while ~400 replies are still owed
# ship prints "newt: write failed broken pipe" repeatedly, then segfaults
```

The larger and more numerous the pending replies, the more reliably it crashes. A single large reply (e.g. scrying `%e` `channel-state` on a busy ship) abandoned mid-write is enough on its own.

### Expected behavior

As @joemfb noted on #490: on early client disconnection the expected behavior is that the broken-pipe errors are printed, the channel/socket is closed, and **the ship stays up**. Instead the write failure on a dead socket propagates into a fault in the King.

### Impact

Any long-lived process that reads ship state over `conn.sock` (monitoring/dashboards, `vats` readers, custom agents' tooling) can crash the ship if it is killed, times out, or is restarted while a reply is in flight. Because pending peek replies for a busy ship can be large, the window is not small. Defensive clients can mitigate by keeping one persistent connection and draining all owed replies before closing, but the ship should not be crashable by a client that simply goes away.

### Notes / possible cause

The path runs through `_conn_moor_bail` -> channel teardown while outstanding peek/fyrd continuations (`u3_cran`) exist. The continuations' back-pointer to the channel is nulled and the callbacks null-check it, but the crash landing in `u3_king_commence` via a loom fault points at a noun being read after it (or its road) is gone, rather than a plain null deref. Happy to provide core dumps or a fuller reproducer.

nisfeb:
### Description

Resolves #1100.

A `conn.sock` client that disconnects while the king still owes it replies can take the ship down. Channel teardown was not idempotent, and the error path re-entered itself:

1. A reply write fails with `EPIPE` → `_newt_write_cb` → `_conn_moor_bail`.
2. `_conn_moor_bail` tries to *send* a `[0 %bail ...]` noun to the client that just went away. That write goes to the same broken pipe. `u3_newt_send` calls `bal_f` directly when `uv_write` fails synchronously, so `_conn_moor_bail` re-enters — and it re-enters before `liv_o` is cleared, so it sends again.
3. Separately, the read side seeing EOF and each queued write failing all call `_conn_moor_bail` for the same chan. Every call ran `_conn_close_chan`, which called `u3_newt_moat_stop` → `uv_close` on handles that were already closing and queued another `%done` to khan, while the chan is freed from the first close callback.

Changes:

- `_conn_close_chan`: return early if the chan's pipe is already closing, making teardown idempotent.
- `_conn_moor_bail`: return early if already closing; clear `liv_o` *before* the courtesy `%bail` write; and skip that write entirely on `EPIPE`/`ECONNRESET`, where the peer is provably gone and the write can only fail and re-enter.
- `u3_newt_send`: drop the write (freeing the buffer) if the stream is already closing rather than submitting it and reporting the inevitable failure through `bal_f` into a teardown already in progress. `_newt_write_cb` already treats writes on a closing handle as canceled; this makes the submit side agree.

### Testing

Both binaries built from this tree with zig 0.15.2 (`zig build -Doptimize=ReleaseFast -Dtarget=x86_64-linux-musl`), run against the *same* fake ship pier with the same script: pipeline N peeks over one conn.sock connection, then close the socket without reading any replies.

The crash needs enough pipelined replies to exceed the socket send buffer, so writes are still queued when the client goes away. Below that threshold the ship never sees EPIPE at all. On unpatched `dbc562a`:

| requests | `conn: moor bail` lines | result |
|---|---|---|
| 2, 10, 40, 100, 200 | 0 | survives |
| 300 | 465,775 | dies, SIGSEGV (exit 139) |
| 400 | 465,757 | dies, SIGSEGV (exit 139) |

A client that sends a single command and closes — the case c0a35c6f63 addresses — is unaffected on develop. Several hundred pipelined requests whose replies are never read is not.

With this branch, 400 requests three times in a row: 3 `conn: moor bail` lines total, one per run, ship still answering. There is no `loom: external fault` stack trace on develop; the process simply takes SIGSEGV after the bail storm.

A self-contained reproducer (python3 only, no helpers) is in the PR comments.

### Related

#490 — same symptom, could not be reproduced at the time. The reproducer in #1100 hits it reliably.

### DIFF
--- a/pkg/vere/io/conn.c
+++ b/pkg/vere/io/conn.c
@@ -288,6 +288,12 @@ _conn_close_chan(u3_shan* san_u, u3_chan* can_u)
   u3_chan*  inn_u;
   u3_cran*  ran_u;
 
+  //  already closing; teardown is idempotent.
+  //
+  if ( uv_is_closing((uv_handle_t*)&can_u->mor_u.pyp_u) ) {
+    return;
+  }
+
   //  unset chan on all pending requests.
   //
   for ( ran_u = can_u->ran_u; ran_u; ran_u = ran_u->nex_u ) {
@@ -335,12 +341,24 @@ _conn_moor_bail(void* ptr_v, ssize_t err_i, const c3_c* err_c)
   u3_chan*  can_u = (u3_chan*)ptr_v;
   u3_shan*  san_u = can_u->san_u;
 
+  //  already closing; freed from the close callback.
+  //
+  if ( uv_is_closing((uv_handle_t*)&can_u->mor_u.pyp_u) ) {
+    return;
+  }
+
   if ( err_i != UV_EOF ) {
     u3l_log("conn: moor bail %zd %s", err_i, err_c);
+
+    //  notify the client, unless it's already gone: writing to a dead
+    //  pipe fails again and re-enters here.
+    //
     if ( _(can_u->liv_o) ) {
-      _conn_send_noun(can_u, u3nq(0, c3__bail, u3i_half(err_i),
-                      u3i_string(err_c)));
       can_u->liv_o = c3n;
+      if ( (UV_EPIPE != err_i) && (UV_ECONNRESET != err_i) ) {
+        _conn_send_noun(can_u, u3nq(0, c3__bail, u3i_half(err_i),
+                        u3i_string(err_c)));
+      }
     }
   }
 
--- a/pkg/vere/newt.c
+++ b/pkg/vere/newt.c
@@ -387,7 +387,16 @@ u3_newt_mojo_stop(u3_mojo* moj_u, u3_moor_bail bal_f)
 void
 u3_newt_send(u3_mojo* moj_u, c3_d len_d, c3_y* byt_y)
 {
-  n_req* req_u = c3_malloc(sizeof(*req_u));
+  n_req* req_u;
+
+  //  a closing stream will never deliver this.
+  //
+  if ( uv_is_closing((uv_handle_t*)&moj_u->pyp_u) ) {
+    c3_free(byt_y);
+    return;
+  }
+
+  req_u = c3_malloc(sizeof(*req_u));
   req_u->moj_u = moj_u;
   req_u->buf_y = byt_y;
 


pkova:
I'm unable to reproduce the crash on develop. Are you sure the clanker built the binary from develop and not vere-v4.6? Master doesn't yet have c0a35c6f63.

nisfeb:
I told it to test against the latest but had it do it again and improve the example case below:

-----

Built from `develop`, not 4.6.

**Build provenance.** The tree is a clone of urbit/vere built at `dbc562a496` (`next` fixups #1099). `git merge-base --is-ancestor c0a35c6f63 dbc562a496` passes, so `c0a35c6f63` is in it. I rebuilt the control just now from that commit checked out detached in a clean tree, confirming `grep -c uv_is_closing pkg/vere/io/conn.c` is 0 before building.

The printed rev can't be used to tell my two binaries apart. `build.zig` takes `buf[41..48]` of `.git/logs/HEAD`, which is the *first* reflog line so both my patched and unpatched builds print `5.0-dbc562a` regardless of what is checked out. That may be worth a separate look.

**Why it probably didn't reproduce.** It needs enough pipelined replies to exceed the socket send buffer, so writes are still queued when the client goes away. Below that threshold the ship never even sees EPIPE. Same pier, same script, only the request count varying, on `dbc562a496`:

| requests | `conn: moor bail` lines | result |
|---|---|---|
| 2, 10, 40, 100 | 0 | survives |
| 200 | 0 | survives |
| 300 | 465,775 | dies, exit 139 |
| 400 | 465,757 | dies, exit 139 |

So a client that sends one command and closes (the scenario in c0a35c6f63) is fine on develop. Several hundred pipelined requests whose replies are never read is not.

**Correction to the issue.** The `loom: external fault` stack trace I quoted in #1100 came from ships running 4.6. On `dbc562a496` there is no stack trace: the process emits ~465k `newt: write failed broken pipe` / `conn: moor bail -32 broken pipe` pairs and then dies with SIGSEGV (wrapper reports `status=139`).

With this branch, the same 400-request run three times in a row: 3 bail lines total, one per run, ship still answering.

**Reproducer**, self-contained, nothing but python3 (the version in the issue omitted the jam/newt helpers, which I suspect is the actual gap). `python3 repro_1100.py /path/to/fake/pier 200`:

```python
import os, socket, sys, time

def jam(n):
    out = pos = 0
    def put(v, w):
        nonlocal out, pos
        out |= (v & ((1 << w) - 1)) << pos; pos += w
    def mat(a):
        if a == 0:
            put(1, 1); return
        b = a.bit_length(); c = b.bit_length()
        put(1 << c, c + 1)
        put(b & ((1 << (c - 1)) - 1), c - 1)
        put(a, b)
    def enc(x):
        if isinstance(x, tuple):
            put(1, 2); enc(x[0]); enc(x[1])   # cell
        else:
            put(0, 1); mat(x)                 # atom
    enc(n)
    return out.to_bytes(max(1, (pos + 7) // 8), 'little')

def newt(n):
    b = jam(n); return b'\x00' + len(b).to_bytes(4, 'little') + b

def tas(s): return int.from_bytes(s.encode(), 'little')

def path(*segs):
    n = 0
    for s in reversed(segs): n = (tas(s), n)
    return n

def peek(view, desk, *spur):
    sp = path(*spur) if spur else (0, 0)
    return (tas('peek'), (1, (tas('once'), (tas(view), (tas(desk) if desk else 0, sp)))))

pier, count = sys.argv[1], int(sys.argv[2]) if len(sys.argv) > 2 else 200
sock = os.path.join(pier, '.urb', 'conn.sock')

s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.connect(sock)
for i in range(count):
    s.sendall(newt((i, peek('e', 'bindings'))))
    s.sendall(newt((10000 + i, peek('bx', '', 'debug', 'timers'))))
s.close()                      # leave every reply owed
print('sent %d requests, closed without reading' % (count * 2))
time.sleep(10)

c = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
c.settimeout(10)
try:
    c.connect(sock); c.sendall(newt((1, (tas('peel'), path('live'))))); c.recv(64)
    print('ship still answers -> PASS')
except Exception as e:
    print('ship does not answer: %r -> FAIL' % (e,)); sys.exit(1)
```

Note the AF_UNIX 108-char path limit if your pier lives somewhere deep. Ship was a fake ship booted from a brass pill; `%e /bindings` and `/bx/debug/timers` are just convenient non-empty replies, anything of similar size should do.

PR #1102, opened 2026-09-08
Hashcons keys and values of persistent HAMTs

dozreg-toplud:
1. Adds `dup_p` HAMT in the road struct that contains deduplicated nouns. The key and the value in each pair is the same `u3_noun` value.
2. `u3z_save_m` (which is the function that `nock.c` uses for memoization) deduplicates both the keys and the values
3. The hashconsing set is pruned in `u3a_reclaim`, which currently runs every 1024 events.
4. The pruning is done in place with a new pruning iterator `u3h_prune_with`. At first I wrote it to run u3h_del on the HAMT as we walk it which is probably very wrong, then I allocated a new HAMT to copy the live values in. The current approach is probably the best as it does not need to allocate and the does not have to copy live values, which would probably mean almost all values.
5. The pruning condition is: the value is only referred to by the dedup HAMT, which means one refcount on the key-value pair (more than that could happen due to meld I imagine) + exactly two counts on the value in the key-value pair. We assert that the key and the value are the same u3_noun value.

Typical torn-writes-related reservations apply I guess, although as long as `u3h_prune_with` is not used on HAMTs that get walked in the signal handler it should be fine?..

Some numbers: `..zuse` contains ~422K unique non-direct nouns, with the maximum depth of the noun being 1000. So `dup_p` would take around 10-20 MB on vere32/64 respectively, and it would take ~1M events to fully flush out a key of that size, which is around one day at 10 events/second.

### DIFF
--- a/pkg/noun/equality_tests.c
+++ b/pkg/noun/equality_tests.c
@@ -19,7 +19,11 @@ _test_unify_home(void)
   u3_noun  b = u3nt(0, 0, 0);
   c3_w kep_w;
 
-  u3_assert( u3t(a) < u3t(b) );
+  if ( u3t(a) > u3t(b) ) {
+    u3_noun t = a;
+    a = b;
+    b = t;
+  }
   kep_w = u3t(a);
 
   (void)u3r_sing(a, b);
--- a/pkg/noun/allocate.c
+++ b/pkg/noun/allocate.c
@@ -1660,7 +1660,7 @@ u3a_print_quac(FILE* fil_u, c3_h den_h, u3m_quac* mas_u)
 u3m_quac*
 u3a_mark_road()
 {
-  u3m_quac** qua_u = c3_malloc(sizeof(*qua_u) * 16);
+  u3m_quac** qua_u = c3_malloc(sizeof(*qua_u) * 17);
 
   qua_u[0] = c3_calloc(sizeof(*qua_u[0]));
   qua_u[0]->nam_c = strdup("namespace");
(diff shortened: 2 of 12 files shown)


dozreg-toplud:
I also removed cax.for_p null check as it was moved from the future-proof buffer to `cax` substruct at some point. But then I could not remember any migrations for that. @midden-fabler does the null check still make sense?

dozreg-toplud:
Also restores comments for structs in struct-defining macros that got removed in vere64 branch.

dozreg-toplud:
Tested some more, some observations:
  1. Number of entries is a lot more than I anticipated, ~50K for urth, ~5M for mars
  2. I tried booting two moons, one on develop, another on this branch. Both boot from the latest brass pill, get OTA from the planet and update a couple of desks.
  3. After they were done total sweep for develop was 475 MB, 479 MB for HC branch
  4. However, `dup_p` got marked with 8 MB of keys. Considering that `dup_p` gets marked almost the last, it means that those keys are not reachable from anywhere else.
  5. I emulated pruning by packing the ship like 20 times, which corresponds to 20 rounds of pruning of `dup_p`. The sweeps got down to 468 MB on HC branch vs 474 MB on develop

A few minutes ago @midden-fabler melded his ship and gained few gigs of memory by deduplicating the memoization tables.

So I guess hashconsing is marginally benificial even for quiet ships and should help a lot ships with more pathological cases of memoization.

pkg/noun/equality_tests.c
```diff
@@ -19,7 +19,11 @@ _test_unify_home(void)
   u3_noun  b = u3nt(0, 0, 0);
   c3_w kep_w;
 
-  u3_assert( u3t(a) < u3t(b) );
```
dozreg-toplud, line 22: @joemfb is that a valid thing to do? The test was crashing, and my guess is that some cells were freed during deduplication of jet state or bytecode or whatever, and then the cells from the free lists got reused for `a` and `b` above, violating the ordering? So instead of asserting I just swap the variables if the order is different.

dozreg-toplud:
By the way, %drop hint is still necessary. Pruning is used on the deduplication set only, and the nouns need to be freed everywhere else for them to get removed from the set, in the absence of weak refcounting for cache tables.

Also deduplication was extend to jet state and bytecode, although I did not notice much of an improvement, but probably because the testing moons were short-lived?

pkg/noun/equality_tests.c
joemfb: yeah, this is a good change.

dozreg-toplud:
TODO:
  - [x] Why e38517a slowed things down even though ostensibly it did the same thing when it comes to structural sharing

dozreg-toplud:
Duh, I forgot to use `u3z_save_m_dedup` instead of `u3z_save_m`... The former was added because the latter has slightly awkward refcount semantic: it claims in the comment that it retains the arguments but it's actually an identity function of `val` when it comes to refcounting, so it can safely be called with an owned `val` and the product can then be freed. At least urwasm jet does this. In `u3z_save_m_dedup` the product and `val` are explicitly transferred as they are not the same.

PR #1104, opened 2026-09-08
Downgrade detection for ford HAMT

dozreg-toplud:
#953 introduced `cax.for_p` at the end of the road struct, which happened to not need a full migration at the time because the loom is zero-initialized, the HAMT post was put at the end of the struct etc. A pseudo-migration was added which checked `cax.for_p` for being equal to zero, initializing it if that was the case.

Similarly to the issue addressed in #975, a downgrade would cause the allocations owned by that HAMT to get orphaned, so they would get e.g. freed by `|mass`. During the upgrade back to the version of Vere with `cax.for_p` the null check for cax.for_p would fail, leaving the possibly bogus value in place.

If I understand correctly, the layout of the loom is:

```
[ver_d][pam_d][arv_u][lom_u]...................V##################V
       (page_size - sizeof u3v_home) ^         ^ rut_p            ^ mat_p
```

and nothing should normally be put in between of the `u3v_home` struct and the bottom of the home road heap. So we will check if those bytes are actually equal to zero.

### DIFF
--- a/pkg/noun/allocate.h
+++ b/pkg/noun/allocate.h
@@ -315,6 +315,8 @@ STATIC_ASSERT( u3a_vits <= u3a_min_log,
     U3_W(S) for_p;                                    \
   } cax;
 
+  //  If the last field of the road struct (including nested sub-struct helpers)
+  //  changes, this value should also change.
   #define U3A_ROAD_LAST_FIELD cax.for_p
 
       U3_DEFINE_PAIR(u3a_road, U3A_ROAD_BODY);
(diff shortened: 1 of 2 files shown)


dozreg-toplud:
Instead of using sizeof(u3v_home) which is not guaranteed to change when things are added to the road struct the offset is calculated with `offsetof` built-in and ~~the macro that represents the last field of the road struct, including nested sub-structs.~~

... and a terminating `c3_y end_y[0]` zero-sized value for precise measurements of the actual size, i.e. sizeof(u3v_home) minus final padding.

PR #1110, opened 2026-09-18
noun: keeps `jmp_buf` out of the road layout

matthew-levan:
On 64-bit Windows, libnoun and libvere disagree about the layout of `u3a_road` and therefore of `u3v_home`. This predates the blob work; it surfaced there because a new test called `u3_meld_all` from a `pkg/vere` test binary, whose `u3_melt_all` asserts `&(u3H->rod_u) == u3R` and failed even though libnoun's own checks of the same condition pass.

**Cause.** `u3a_road` embedded a `jmp_buf` in `u3a_road_esc_{h,d}`, and on Windows the two libraries see different `jmp_buf`s. libnoun is compiled with `pkg/noun/platform/windows` on its include path, so `<setjmp.h>` resolves to the 8-aligned struct in `platform/windows/setjmp.h`; libvere and the test executables only get the headers noun installs, so they see mingw's `jmp_buf`, which on x64 is `SETJMP_FLOAT128[16]` with 16-byte alignment. The union is 2048 bytes either way, but its alignment differs, so `u3a_road_d` is 8-aligned in libnoun and 16-aligned in libvere: `offsetof(u3v_home_d, rod_u)` is 40 in one and 48 in the other, and every road field after `esc` (`how`, `all`, `hep`, `cel`, `jed`, `byc`, `ski`, `bug`, `pro`, `cax`, ...) is read 8 bytes off by all of libvere on 64-bit Windows. The same split affects `u3a_road_h` inside a 64-bit binary, the layout a bitness migration reads a 32-bit snapshot through. 32-bit Windows is unaffected because both `jmp_buf`s there are 4-aligned and the road already needs 8 for its `c3_d` fields. The failure signature in CI makes the split visible directly: the assertion in melt.c (libvere) fails, then `u3m_bail` in libnoun prints `home: bailing out`, its own evaluation of the same comparison being true.

**Fix.** A loom layout must not depend on a libc type. The escape slot is now a plain word array (`c3_h[256]` / `c3_d[256]`, the same sizes as before); manage.c, the only setjmp user, casts into it and keeps the 64-bit bail reason in the last word. Static assertions pin the escape buffer sizes, the road alignments, and the `rod_u` offsets (40 and 32), so every translation unit must agree at compile time. On every platform that already agreed, the layout is unchanged, so existing snapshots are unaffected; on 64-bit Windows the on-disk layout was libnoun's, which is the one kept.

**Test.** `vere-noun-test` gains `_test_home_road`, compiled with libvere's include view: it checks the road address against `u3R`, then writes `bug.mer`, a field past the escape buffer, and has libnoun mark and sweep with `u3o_leak_crash` armed, so a layout disagreement aborts. It fails on 64-bit Windows before this change and passes after.

Verified locally in both loom configurations (vere-noun, nock, retrieve, serial, and palloc tests) and by cross-compiling both Windows targets; the same two commits pass all four CI jobs, Windows 64-bit included, on `ml/bob-fds-win` (#1109).


dozreg-toplud:
my question would only be why not `__attribute__((aligned(16)))` (pre-C11 clang/gcc) or `alignas(16)` (C11 keyword)

PR #1111, opened 2026-09-18
mars: expires blob leases by committed events

matthew-levan:
Replaces the wall-clock blob lease TTL with expiry counted in committed events, and removes lease renewal.

A king lease only has to outlive the events queued ahead of the poke that carries the blob, and that queue is counted in events, not seconds. The wall clock measured something unrelated, which is why the design needed a 5-minute renewal timer: a long event, `|pack`, `|meld`, or a sleeping laptop advanced the clock while the `%blas` that would refresh the lease sat in the same FIFO behind the stall.

## mars

- `_mars_lease_take` issues one lease at install, expiring once `dun_d + _mars_lease_wid_d` (1 << 16) commits. Renewal mode (`new_o`) is gone.
- The sweeper at the top of `_mars_work` compares the PQ root against `dun_d` instead of `gettimeofday`. It runs before every writ, so nothing expires late by more than one writ.
- `_mars_play_leases` prunes rows whose expiry event is already committed and re-arms the rest at their persisted event number. It already ran after replay, so `dun_d` is final when it compares.
- The `%blas` handler is removed.

## king and lord

- The 5-minute renewal timer and its bank walk are deleted from `king.c`. The king only registers the delete hook, which still sends `%blrl` when its last reference to a blob dies.
- `u3_lord_blob_lease` and the `%blas` writ are removed from the protocol.

## LEASES table

The row format is unchanged; `exp_d` is now an event number, with 0 meaning never. No migration. The table stays durable so leases can outlive a power failure for future non-king mars clients; the missing piece for that world is a client id per row, which is additive.

## Verification

- `zig build` and `zig build blob-test` pass on 32- and 64-bit (48 tests each).
- `zig build -Dvere64 -Dtarget=x86_64-windows-gnu` builds.
- The three LMDB lease tests pass unchanged; they treat the expiry column as an opaque number.

Not covered by a test: the sweeper and the boot pruning predicate are static inside `mars.c`, so no unit test drives expiry by committing events.


matthew-levan:
Closing in favor of #1118.

